PR #148, opened 2025-01-02
Add a script to run the raw TTIR from models excel file with TTRT

mmanzoorTT:
We can run the models with `EXECUTE_OP_BY_OP` in CI; so we don't need TTRT script to run TTIR graphs. 
This script is added for reference.


mmanzoorTT:
No more required. closing it.

PR #163, opened 2025-01-06
Add workflow for ttnn md files generation

ddilbazTT:
Add workflow for ttnn md files generation to nightly tests. Currently, created files are not pushed to docs/ops/ttnn. As next iteration, these files could be published on Pages. Removed "Runs on hardware" from created files.


ddilbazTT:
For reference, [here](https://github.com/tenstorrent/tt-torch/actions/runs/12630217184) is a successful nightly test with the new workflow

vmilosevic (approved):

PR #170, opened 2025-01-07
Fix halting N300 tests.

LPanosTT:
Open the subprocesses which compile and run ops in the op-by-op flow by spawning rather than forking the main process. 

If an error is thrown during the compilation, catch it, terminate the process, then raise it again.


tests/torch/test_basic.py
```diff
@@ -9,6 +9,11 @@
 from tt_torch.tools.verify import verify_module
 from tt_torch.tools.utils import CompilerConfig
 
+import os
+
+print("LISTING THE CONTENTS OF /dev/hugepages-1G")
+print(os.listdir("/dev/hugepages-1G"))
```
LPanosTT, line 15: Don't mind this. It will be removed.

.github/workflows/run-model-tests.yml
```diff
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         build: [
-          {runs-on: n150, name: "run1", test_names: "stable_diffusion, Qwen, MobileNetV2, clip, flan_t5, mlpmixer, resnet, vilt, albert, codegen, glpn_kitti, mnist, resnet50, RMBG, unet_carvana, mgp-str-base, musicgen_small, segformer, torchvision, yolos"},
-          {runs-on: n150, name: "run2", test_names: "t5, whisper, autoencoder_conv, deit, gpt2, mobilenet_ssd, roberta, timm, xglm, autoencoder_linear, detr, beit, distilbert, hand_landmark, openpose, segment_anything, unet, yolov3, bert, dpr, hardnet, opt, speecht5_tts, unet_brain, yolov5, bloom, falcon, llama, perceiver_io, squeeze_bert, gpt_neo"},
+          {runs-on: n150, n300, name: "run1", test_names: "stable_diffusion, Qwen, MobileNetV2, clip, flan_t5, mlpmixer, resnet, vilt, albert, codegen, glpn_kitti, mnist, resnet50, RMBG, unet_carvana, mgp-str-base, musicgen_small, segformer, torchvision, yolos"},
```
AleksKnezevic, line 17: Should we skip the nightlies until we figure out what's causing the hangs?

PR #181, opened 2025-01-09
Compare PCC/ATOL in ModelTester

LPanosTT:
- Removed unused code from `ModelTester`
 - Add PCC/ATOL verification of model outputs
 - Minor configuration specificities needed to allow some models to pass

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -2,24 +2,54 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import torch
+import onnx
+from onnx.tools import update_model_dims
+import onnxruntime
 import numpy as np
 import collections
 import re
 from typing import List, Dict, Tuple
 from tt_torch.dynamo.backend import backend
+from tt_torch.onnx_compile import compile_onnx
 from tt_torch.tools.utils import CompilerConfig, CompileDepth
 import json
+from onnx import version_converter
 from pathlib import Path
+from tt_torch.tools.verify import verify_against_golden
 
 
 class ModelTester:
-    def __init__(self, model_name, mode, compiler_config=None):
+    def __init__(
+        self,
+        model_name,
+        mode,
+        pcc=0.99,
+        required_atol=None,
+        relative_atol=None,
+        compiler_config=None,
+    ):
         if mode not in ["train", "eval"]:
             raise ValueError(f"Current mode is not supported: {mode}")
         self.model_name = model_name
         self.mode = mode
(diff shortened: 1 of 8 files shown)


tests/utils.py
```diff
@@ -84,79 +121,21 @@ def run_model(self, model, inputs):
...
-        outputs = self.run_model(model, inputs)
-        loss = self.append_fake_loss_function(outputs)
-        loss.backward()
-        # Again, use the gradient of the input (`test_input.grad`) as the golden result for the training process.
-        results = self.get_results_train(model, inputs, outputs)
-        return results
-
     @torch.no_grad()
     def test_model_eval(self, on_device=True):
-        model = self.set_model_eval(self.model)
```
AleksKnezevic, line 153: perhaps simply `self.model = self.model.eval()`

tests/utils.py
```diff
@@ -166,433 +145,85 @@ def test_model(self, on_device=True):
...
-                )
-                break
-        return input_vals
-
-    def _adjust_where_self(self, input_vals):
-        for input_val in input_vals:
-            if input_val["name"] == "condition":
-                input_val["val"] = torch.randint(0, 2, input_val["val"].shape).type(
-                    torch.bool
+        assert (atol is not None) != (
```
AleksKnezevic, line 156: can you move this stuff to verify.py?

AleksKnezevic (changes requested):

tests/utils.py
LPanosTT: We don't want to use `self.model` throughout that function because if we set `self.model = self.compile_model(model, self.compiler_config)` (a few lines below) then we'll lose the PyTorch model.

tests/utils.py
LPanosTT: Sure

tests/utils.py
AleksKnezevic: I see. Makes sense. Is it better to just hold onto a `framework_model` and `compiled_mode` reference? And drop `self.model`?

LPanosTT: Probably. On it.

LPanosTT: I've gotten rid of `set_model_eval` since it seems to do nothing. I would like to make this a superclass in the future and create an `OnnxModelTester` and `PyTorchModelTester` in the future when I do this for ONNX models. So I'd rather get rid some things that seem unnecessary now.

PR #184, opened 2025-01-10
fix checking unused getitems in op by op flow

brataTT:
This resolves failure in model tests
where ops with single output that are consumed by multiple other ops result in an exception

### DIFF
--- a/tests/torch/test_softmax.py
+++ b/tests/torch/test_softmax.py
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+from torch import nn
+import pytest
+
+import tt_torch
+from tt_torch.tools.verify import verify_module
+
+
+@pytest.mark.xfail(reason="softmax_kernel_impl not implemented for Bool")
+def test_safe_softmax_bool():
+    class Basic(nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            return torch._safe_softmax(x, 1)
+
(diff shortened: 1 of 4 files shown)


AleksKnezevic (approved):

tests/models/beit/test_beit_image_classification.py
```diff
@@ -46,6 +46,7 @@ def _extract_outputs(self, output_object):
     "model_name",
     ["microsoft/beit-base-patch16-224", "microsoft/beit-large-patch16-224"],
 )
+@pytest.mark.xfail(reason="Fails due to passing bool to softmax")
```
AleksKnezevic, line 49: if this test is intended to pass, please remove xfail.

tests/torch/test_softmax.py
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: Apache-2.0
+import torch
+from torch import nn
+import pytest
+
+import tt_torch
+from tt_torch.tools.verify import verify_module
+
+
+@pytest.mark.xfail(reason="softmax_kernel_impl not implemented for Bool")
```
AleksKnezevic, line 12: This will need to be a test that uses the current transformers library (in which safe_softmax passes). You can add this in a subsequent pr.

AleksKnezevic (changes requested):

tests/torch/test_softmax.py
brataTT: I just tested this with older transofmers, and it still fails
For some reason, model tests pass
Since it does still fail, I kept it in

AleksKnezevic: perfect. maybe the update was just to use safe_softmax for this model. Actually, that makes sense since it's a transformers, not a torch update.

AleksKnezevic (approved):

PR #187, opened 2025-01-10
Mark test_linear_with_bias_no_embedded_constants as xfail to unblock tt-mlir uplift

uazizTT:


AleksKnezevic (commented):
Is this mergable into main? Won't it xpass until the uplift?

uazizTT:
> Is this mergable into main? Won't it xpass until the uplift?

Oh yes you are right, it will only work if it was skipping the test instead. Let me modify the uplift PR.

PR #193, opened 2025-01-13
Add comparison of PCC/ATOL for model tests

LPanosTT:
- Removed unused code from `ModelTester`
 - Add PCC/ATOL verification of model outputs
 - Minor configuration specificities needed to allow some models to pass

### DIFF
--- a/tests/models/Qwen/test_qwen2_token_classification.py
+++ b/tests/models/Qwen/test_qwen2_token_classification.py
@@ -45,7 +45,7 @@ def test_qwen2_token_classification(record_property, model_name, mode, nightly):
     if nightly:
         cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
 
-    tester = ThisTester(model_name, mode, compiler_config=cc)
+    tester = ThisTester(model_name, mode, relative_atol=0.01, compiler_config=cc)
     with torch.no_grad():
         results = tester.test_model()
 
@@ -53,7 +53,8 @@ def test_qwen2_token_classification(record_property, model_name, mode, nightly):
         logits = results.logits
         predicted_token_class_ids = logits.argmax(-1)
         predicted_tokens_classes = [
-            tester.model.config.id2label[t.item()] for t in predicted_token_class_ids[0]
+            tester.framework_model.config.id2label[t.item()]
+            for t in predicted_token_class_ids[0]
         ]
         input_ids = tester.inputs["input_ids"]
         tokens = tester.tokenizer.convert_ids_to_tokens(input_ids[0])
--- a/tt_torch/tools/verify.py
+++ b/tt_torch/tools/verify.py
@@ -11,6 +11,101 @@ from tt_torch.dynamo.backend import backend
 from tt_torch.tools.utils import calculate_atol, calculate_pcc
 
 
+def verify_against_golden(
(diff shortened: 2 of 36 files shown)


tests/models/Qwen/test_qwen2_token_classification.py
```diff
@@ -11,12 +11,12 @@
 class ThisTester(ModelTester):
     def _load_model(self):
         return Qwen2ForTokenClassification.from_pretrained(
-            self.model_name, torch_dtype=torch.bfloat16
+            self.model_name, torch_dtype=torch.float32
```
AleksKnezevic, line 14: We should be consistent. Either load every model and all activations in bf16 or default to fp32 (I don't think you need to add `torch_dtype=torch.float32` to the test. It's the default.

tt_torch/tools/verify.py
```diff
@@ -11,6 +11,102 @@
 from tt_torch.tools.utils import calculate_atol, calculate_pcc
 
 
+def verify_against_golden(
+    golden_tensors,
+    calculated_tensors,
+    pcc=0.99,
```
AleksKnezevic, line 17: nit, can we call these required_pcc, required_atol, etc?

tt_torch/tools/verify.py
```diff
@@ -11,6 +11,102 @@
...
+    ), f"Expecting the calculated tensors to be a tuple of tensors after _extract_outputs. Got type: {type(calculated_tensors)}"
+    assert len(golden_tensors) == len(
+        calculated_tensors
+    ), "Expecting the number of golden and calculated tensors to be the same."
+
+    pccs, pcc_passeds = [], []
+    atols, atol_thresholds, atols_passeds = [], [], []
+
+    # Distinct value to put in the `pccs` list so we can append the correct log
+    SKIPPED_PCC_CALCULATION_FOR_SINGLE_VALUE = "skipped for single value"
```
AleksKnezevic, line 38: maybe use `None`?

AleksKnezevic (changes requested):

tt_torch/tools/verify.py
LPanosTT: I thought about that initially, but settled on this incase we end up having multiple possible reasons to skip pcc/atol, we'd want to be able to distinguish them. However, that's a small change and `None` is cleaner for now. I'll change it.

tt_torch/tools/verify.py
```diff
@@ -11,6 +11,106 @@
...
+
+    passed = True
+    err_msg = ""
+    msg = ""
+    for i, ((pcc_passed, pcc_), (atol_passed, atol_), atol_threshold) in enumerate(
+        zip(zip(pcc_passeds, pccs), zip(atols_passeds, atols), atol_thresholds)
+    ):
+        msg = msg + f"Results for output {i}:\n"
+        if pcc_passed:
+            if pcc_ == SKIPPED_PCC_CALCULATION_FOR_SINGLE_VALUE:
```
AleksKnezevic, line 76: I don't think this is necessary. If it's a rank 0 tensor, PCC is undefined.

AleksKnezevic (approved):
Tiny nit change requested, once you fix feel free to merge.

mmanzoorTT (approved):

PR #213, opened 2025-01-20
Add location information propagation from FX module

LPanosTT:
Added a custom `ContextCache` class so I could override how `FxImporter` determines locations to get some rudimentary location information in the graph.

Some bindings for runtime API

Ensure that location information is dumped into the module string if
TT_RUNTIME_DEBUG=ON

### DIFF
--- a/tt_torch/csrc/tt-mlir-interface.cpp
+++ b/tt_torch/csrc/tt-mlir-interface.cpp
@@ -33,7 +33,6 @@
 #include "stablehlo/dialect/StablehloOps.h"  // from @stablehlo
 #include "stablehlo/transforms/Passes.h"     // from @stablehlo
 
-#define TTMLIR_ENABLE_STABLEHLO
 #include "ttmlir/Dialect/TTIR/Pipelines/TTIRPipelines.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h"
@@ -92,7 +91,7 @@ std::string compileStableHLOToTTIR(std::string_view code) {
   }
   std::string buffer;
   llvm::raw_string_ostream os(buffer);
-  mlir_module.get()->print(os);
+  mlir_module.get()->print(os, mlir::OpPrintingFlags().enableDebugInfo());
   os.flush();
 
   return buffer;
@@ -150,7 +149,7 @@ compileTTIRToTTNN(std::string_view code) {
 
   std::string buffer;
   llvm::raw_string_ostream os(buffer);
-  mlir_module->print(os);
+  mlir_module->print(os, mlir::OpPrintingFlags().enableDebugInfo());
   os.flush();
 
   return std::make_tuple(binary, buffer);
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -19,8 +19,10 @@ from tt_torch.tools.utils import (
 )
 
 import tt_mlir
-from torch_mlir.ir import Context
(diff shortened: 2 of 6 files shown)


tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -33,7 +33,6 @@
 #include "stablehlo/dialect/StablehloOps.h"  // from @stablehlo
 #include "stablehlo/transforms/Passes.h"     // from @stablehlo
 
-#define TTMLIR_ENABLE_STABLEHLO
```
AleksKnezevic, line 36: Thank you!

tt_torch/dynamo/backend.py
```diff
@@ -39,10 +40,18 @@
 import tempfile
 
 
+class TTContextCache(ContextCache):
+    def get_node_location(self, node: torch.fx.Node) -> Optional[Location]:
+        return Location.name(node.name, context=self._c)
```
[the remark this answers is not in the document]
AleksKnezevic: I mean, I know what it does, but how does it work?

tt_torch/dynamo/backend.py
```diff
@@ -492,12 +501,27 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
     if compiler_config.compile_depth == CompileDepth.STABLEHLO:
         return executor
 
-    ttir = tt_mlir.compile_stable_hlo_to_ttir(module.operation.get_asm())
+    # Need to set enable_debug_info=True to get the location information for the ops in the asm string
+    ttir = tt_mlir.compile_stable_hlo_to_ttir(
+        module.operation.get_asm(enable_debug_info=True)
```
AleksKnezevic, line 506: Is this significantly slower? Can you try a largeish graph? If so, CI should have it disabled by default.

tt_torch/dynamo/backend.py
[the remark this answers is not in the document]
LPanosTT: The `FxImporter` carries a `ContextCache` object. To generate MLIR locations from the fx nodes, it queries the `ContextCache` using `get_node_location`. Since we currently lose the original file/line locations, the default `ContextCache.get_node_location` will return `loc(unknown)` for every node, which is undesirable.

tt_torch/dynamo/backend.py
```diff
@@ -492,12 +501,27 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
...
+    )
     if dump_intermediates:
         print("TTIR module", file=sys.stderr)
         print(ttir, file=sys.stderr)
 
+    def golden_callback(bin, callback_context, op_context):
```
AleksKnezevic, line 512: This probably belongs in the executor class.

tt_torch/dynamo/backend.py
LPanosTT: Seems to not be. I ran the compile of the simple mnist 1000 times with the debug info and 1000 times without.

1000 shlo --> ttir compiles with debug info took 2.58 seconds

1000 shlo --> ttir compiles without debug info took 2.51 seconds.

AleksKnezevic: try a larger graph.

LPanosTT: I had resnet50 compile from shlo all the way down to the flat buffer 5 times with debug info, and 5 times without.

With debug info the 5 compiles took 30.742 seconds, without debug info it took 30.681 seconds. So not a huge difference.

AleksKnezevic: Awesome, thanks for confirming. For some reason, in tt-mlir the debug strings took >95% of entire compilation time.

PR #215, opened 2025-01-21
Update ttnn md files through github actions

ddilbazTT:
Updated `.github/workflows/generate-ttnn-md.yml` so that generated md/json files are pushed after each nightly test.

This is a draft PR because I ran this branch with nightly tests using only 2 models so that it wouldn't clog jobs. In the branch commit history, you can see the md/json files being deleted then pushed after generated using nightly test. [link](https://github.com/tenstorrent/tt-torch/compare/main...refs/heads/ddilbaz/pages-workflow ) In the final PR, the only file changed will be `.github/workflows/generate-ttnn-md.yml`.

image


.github/workflows/generate-ttnn-md.yml
```diff
@@ -37,14 +42,37 @@ jobs:
...
+
       - name: Generate TTNN MD Files
         shell: bash
         run: |
           source env/activate
-          echo "${{ steps.strings.outputs.work-dir }}/tt_torch/tools/generate_md.py"
-          echo "${{ steps.strings.outputs.work-dir }}/results/models_op_per_op.xlsx"
           python ${{ steps.strings.outputs.work-dir }}/tt_torch/tools/generate_md.py --excel_path ${{ steps.strings.outputs.work-dir }}/results/models_op_per_op.xlsx --md_dir ${{ steps.strings.outputs.work-dir }}/docs/ops/ttnn --json_dir ${{ steps.strings.outputs.work-dir }}/docs/ops/ttnn
 
+      - name: Add and commit new files
```
AleksKnezevic, line 66: I don't think we should remove and commit each evening. That will clog git history and give us too many, mostly empty commits. I thin a better way would be to remove the old files, re-generate and use `git status` to see if there are any changes, and if there are commit and push.

AleksKnezevic (changes requested):

.github/workflows/generate-ttnn-md.yml
ddilbazTT: I changed it to something similar to what you recommended, I'm testing it with a single model now: [link](https://github.com/tenstorrent/tt-torch/actions/runs/12894068664)

vmilosevic (approved):

PR #220, opened 2025-01-23
Ddilbaz/op by op shlo

ddilbazTT:
Hi all,

This is a POC for incorporating the shlo op-by-op compiler to our existing backend. I would appreciate any and all inputs. 

- shlo compiler is based on `tt_torch/shlo_compile/shlo_compile.py` however, this file will not exist in the final draft
- `backend.py` only has calls to `_shlo_backend `or `_base_backend`, everything else is moved under torch_compile.py
- current `Op` class does not translate well with shlo compiler, so implemented an `StableHLOOp` Class which resembles it. 
- currently, we only support `COMPILE_OP_BY_OP`. As of writing, I realize I could also add `COMPILE` support but I want to know if we still want to divide shlo module into sub modules for each op if we follow that route.

Thank you.

### DIFF
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -4,16 +4,19 @@
 import torch
 import os
 import tt_mlir
+import sys
 
 from tt_torch.dynamo.torch_backend import (
     TorchFXExecutor,
     import_graph,
 )
-from tt_torch.dynamo.shlo_backend import StableHLOExecutor
+from tt_torch.dynamo.shlo_backend import (
+    StableHLOExecutor,
+    parse_module_from_str,
+)
 from tt_torch.dynamo.passes import pass_pipeline
 from tt_torch.tools.utils import (
     CompilerConfig,
-    FrontEnd,
     CompileDepth,
 )
 
@@ -41,7 +44,7 @@ def _shlo_backend(module_str, options=None):
     return executor
 
 
-def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
+def _torch_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
     # Apply environment overrides at start of compilation to allow overriding what was set in the test
     compiler_config.apply_environment_overrides()
     with torch.no_grad():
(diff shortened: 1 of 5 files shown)


tt_torch/dynamo/backend.py
```diff
@@ -499,17 +86,27 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
...
-    # fake_tensor_mode.allow_non_fake_inputs = True
-    # aten = make_fx(gm, tracing_mode="symbolic", decomposition_table={}, _allow_non_fake_inputs=True)(*example_inputs)
-    # return _base_backend(aten, example_inputs)
-    return _base_backend(gm, example_inputs, compiler_config=options)
-
-
-# backend = aot_autograd(fw_compiler=_base_backend)
+    if options.compiler_front_end == FrontEnd.STABLEHLO:
+        if not isinstance(gm_or_module, str):
+            exit(1)
```
mmanzoorTT, line 94: Please add a message before calling `exit`

mmanzoorTT:
Can you provide an for its usage?

tt_torch/dynamo/backend.py
```diff
@@ -499,17 +86,27 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
     return executor
 
 
-def backend(gm, example_inputs, options=None):
```
AleksKnezevic, line 502: Thanks for refactoring this! I think you should change it around a bit. IMO, we should have three flows:

1. Base either SHLO module or torch fx graph that we want to execute fully on device
2. SHLO op-by-op either compile or execute
3. Torch op-by-op either compile or execute

Does this make sense? Is there a reason to keep SHLO and torch separate when running the whole graph?

tt_torch/dynamo/backend.py
```diff
@@ -499,17 +86,27 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
...
+module {
+  func.func @main(%arg0: tensor<1x128xf32>, %arg1: tensor<128xf32>) -> tensor<1x128xf32> {
+    %0 = stablehlo.broadcast_in_dim %arg0, dims = [0, 1] : (tensor<1x128xf32>) -> tensor<1x128xf32>
+    %1 = stablehlo.broadcast_in_dim %arg1, dims = [1] : (tensor<128xf32>) -> tensor<1x128xf32>
+    %2 = stablehlo.add %0, %1 : tensor<1x128xf32>
+    return %2 : tensor<1x128xf32>
+  }
+}
+"""
+compiler_config = CompilerConfig()
```
AleksKnezevic, line 109: Should this be here?

tt_torch/dynamo/backend.py
ddilbazTT: We might need to sync up about this to make sure I am not mistaken. I assume right now I implemented half of 2, and 3 is just moved to a different file. Torch and stablehlo have separate executors and backends. Do we want to merge the executors and/ or backend? Or what I did right now could stay but we need minor tweaks? 

Also small nit, I think shlo execution will come some time later since I haven't figured out how to pass inputs and run and get the result from shlo mlir.

PR #231, opened 2025-01-28
Move problematic models to their own job and always generate artifact spreadsheeet

AleksKnezevic:


LPanosTT (approved):

ddilbazTT (approved):
There's a risk we might not catch any model that starts unexpectedly failing though it was passing before. But I can't think of a better/ easier way of implementing a fix than this pr either.

Issue #123, opened 2024-12-16
Enable torchvision models test


AleksKnezevic:
Torchvision models tests are currently skipped. Please enable them. You'll need to modify `ThisTester` class as defined in the test to use the same class as used in the other model tests.

PR #235, opened 2025-01-29
Fixed model tester setup for torchvision tests

LPanosTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-torch/issues/123)

### Problem description
The torchvision model tests are currently skipped due to an issue with the ModelTester

### What's changed
Instead of using passing just the model name, and test mode to the model tester, we also needed to pass weight names. So we needed to override the constructor. Beforehand we did this without calling `super().__init__(...)` so critical initialization did not occur. This has been fixed.


AleksKnezevic (approved):
Nice, that was simple.

PR #238, opened 2025-01-29
Divided models into compile/run, added models

AleksKnezevic:
### Problem description
It was tough to see what was actually running on CI.

### What's changed
Separate push model tests into compile/execute
Remove compile depth overrides from tests

### Checklist


ddilbazTT:
Sorry, do we end up using "op_by_op"? I couldn't follow how we pass it to the tests.

AleksKnezevic:
@ddilbazTT, these are push tests, no op_by_op, this is what compiles or runs the entire model e2e.

PR #240, opened 2025-01-30
Add support for OpModel pydantic model in downstream parser

AleksKnezevic:
### Problem description
We want to switch to OpModel pydantic model in tt-github-actions. This PR should allow us to do that.

### DIFF
--- a/tt_torch/tools/utils.py
+++ b/tt_torch/tools/utils.py
@@ -35,12 +35,41 @@ class OpCompilationStatus(IntEnum):
     EXECUTED = 7
 
 
+class Tensor:
+    def __init__(self, shape):
+        constrained_shape = []
+        for dim in shape:
+            if isinstance(dim, int):
+                constrained_shape.append(dim)
+            else:
+                constrained_shape.append(-1)
+        self.shape = constrained_shape
+
+        self.data_type = ""
+        self.buffer_type = ""
+        self.layout = ""
+        self.grid_shape = []
+
+    def to_dict(self):
+        return {
+            "shape": self.shape,
+            "data_type": self.data_type,
+            "buffer_type": self.buffer_type,
+            "layout": self.layout,
+            "grid_shape": self.grid_shape,
+        }
+
+
 class Op:
-    def __init__(self, torch_name, input_shapes):
-        self.torch_name = torch_name
+    def __init__(self, torch_name, input_shapes, model_name):
+        self.framework_op_name = torch_name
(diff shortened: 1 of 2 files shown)


tt_torch/tools/utils.py
```diff
@@ -78,10 +125,20 @@ def scrub_nan_inf(value):
...
         atol = scrub_nan_inf(self.atol)
 
+        if len(self.input_tensors) == 0:
+            self.input_tensors = [Tensor(shp) for shp in self.input_shapes]
+
+        if len(self.output_tensors) == 0:
+            self.output_tensors = [Tensor(shp) for shp in self.output_shapes]
+
         return {
-            "torch_name": self.torch_name,
```
nsmithtt, line 82: Currently tt-github-actions is still accessing this field
https://github.com/tenstorrent/tt-github-actions/blob/a394b3793a96bea711ce350c2a61dd631943a8ac/.github/actions/collect_data/src/parsers/tt_torch_model_tests_parser.py#L98

For a grace period can we leave this one and once tt-github-actions repo has flipped the switch we can remove the unused fields.

nsmithtt (approved):
Aside from the `torch_name` field, looks great!

tt_torch/tools/utils.py
AleksKnezevic: makes sense. I've added it back.

tt_torch/tools/utils.py
```diff
@@ -35,12 +35,41 @@ class OpCompilationStatus(IntEnum):
...
+    def __init__(self, shape):
+        constrained_shape = []
+        for dim in shape:
+            if isinstance(dim, int):
+                constrained_shape.append(dim)
+            else:
+                constrained_shape.append(-1)
+        self.shape = constrained_shape
+
+        self.data_type = ""
```
ddilbazTT, line 48: I think `None` might be better than `""`

ddilbazTT (approved):

tt_torch/tools/utils.py
AleksKnezevic: Yeah, I agree it's a little odd, but they're strings because that's what the pydantic model that consumes them expects.

nsmithtt: Yeah we should follow how pydantic has spec'd them.

ddilbazTT: Sorry I thought they would be enums my bad

Issue #265, opened 2025-02-04
Runtime error message extraction


mmanzoorTT:
Error messages are not extracted during the execution of models. Please see the nightly tests model excel sheet. https://github.com/tenstorrent/tt-torch/actions/runs/13106153515

PR #266, opened 2025-02-04
Update capturing of runtime strck dump process

mmanzoorTT:
### Ticket
#265 

### Problem description
Runtime stack dump is not captured for execution of model tests. 

### What's changed
Stack dump is captured during execution of the runtime instead of entire sub process which fails due to utilizing single/same process to execute multiple graphs to avoid process creation overhead.

### Checklist

### DIFF
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -37,6 +37,7 @@ import os
 import multiprocessing as mp
 import time
 import faulthandler
+from pathlib import Path
 import re
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


tt_torch/dynamo/backend.py
```diff
@@ -127,6 +137,15 @@ def __init__(
         self.execute_sender = None
         self.execute_receiver = None
 
+        # Create a temp file which will be used by sub process to dump stderr
+        # output during execution of the mlir graph on TT device.
+        self.f_stderr = tempfile.NamedTemporaryFile(mode="w+t", delete=False)
```
AleksKnezevic, line 142: This shouldn't happen when we create the executor, as this is at compile time. The user could compile the model, do some other stuff, and then finally run. This would interfere with stderr outside of op_by_op runtime. I think this should happen when we're running the first op, somewhere around line 345. ie
```python
if not self.stderror_redirected:
 ...
```

tt_torch/dynamo/backend.py
```diff
@@ -127,6 +137,15 @@ def __init__(
...
+
+    # Class destructor
+    def __del__(self):
```
AleksKnezevic, line 145: Similar as above, this should happen when the op_by_op execution finishes.

tt_torch/dynamo/backend.py
mmanzoorTT: Thanks for the suggestion. Now the temp file is created before execution of first op (while running op-by-op) and removed after execution of all the ops.

tt_torch/dynamo/backend.py
mmanzoorTT: fixed.

AleksKnezevic (approved):

PR #268, opened 2025-02-04
Create bfloat16 tensor for boolean program inputs

### LINKED ISSUE #271: Failure for subgraph inputs for op by op execution
TT devices does not support all the torch data types; so we convert them to supported data types. Currently we only cast the models input at the beginning of executing a model. subgraphs are generated for op by op execution and may require unsupported data types as input which causes failure during execution of an op.

mmanzoorTT:
closes #271 

### Ticket
#271 

### Problem description
subgraphs (generated for op by op execution) may require inputs of unsupported data types. which causes failure for op execution.

### What's changed
Add type conversion of inputs to supported data types for subgraph inputs for op by op execution.

### Checklist

### DIFF
--- a/tt_torch/csrc/bindings.cpp
+++ b/tt_torch/csrc/bindings.cpp
@@ -37,6 +37,10 @@ static tt::target::DataType torch_scalar_type_to_dt(torch::ScalarType st) {
   // case torch::ScalarType::Bool:
   case torch::ScalarType::BFloat16:
     return tt::target::DataType::BFloat16;
+  case torch::ScalarType::Bool:
+    // tt-metal does not support boolean data type; so bfloat16 data type is
+    // used instead.
+    return tt::target::DataType::BFloat16;
   default:
     break;
   }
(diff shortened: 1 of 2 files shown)


tt_torch/csrc/bindings.cpp
```diff
@@ -37,6 +37,10 @@ static tt::target::DataType torch_scalar_type_to_dt(torch::ScalarType st) {
   // case torch::ScalarType::Bool:
   case torch::ScalarType::BFloat16:
     return tt::target::DataType::BFloat16;
+  case torch::ScalarType::Bool:
```
AleksKnezevic, line 40: don't we cast these in backend.py?

mmanzoorTT: backend.py handles the inputs to the model. We can still have boolean inputs for sub graphs (in case of op_by_op execution) which fails unsupported data type.

mmanzoorTT: Memory requirements are different for different data types. Casting ensures the input memory requirements matches with the data type and proper amount of data is moved between host and device. We cannot remove the casting from backend.py.

AleksKnezevic (approved):

Issue #142, opened 2024-12-21
yolov5 test is not working


AleksKnezevic:
Please investigate

PR #274, opened 2025-02-04
yolov5 fixes

ddilbazTT:
Fix libg1 dependency issue by switching from opencv-python to opencv-contrib-python. Current output 0 atol is too high.

### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-torch/issues/142)

### Problem description
libg1 is missing. `opencv-contrib-python` can overcome this, but installing `opencv-contrib-python` makes `opencv-python` redundant. Furthermore, yolov5 returns a list of tensor list, which needs to be flattened. Afterwards, atol and pcc are not as wanted.

I would appreciate if you could review and give suggestions about atol/ pcc. 

image


AleksKnezevic (approved):

ddilbazTT:
@AleksKnezevic I would need to overwrite atol/pcc defaults to run the test since they would fail the test otherwise. I will do this through ModelTester if we want to add to nightly test. Is that ok?

AleksKnezevic:
@ddilbazTT you can provide an override in the test itself, can you not. i.e. in test_yolov5.py

ddilbazTT:
> @ddilbazTT you can provide an override in the test itself, can you not. i.e. in test_yolov5.py

Yes that's what I meant, I will overwrite required_pcc and required_atol:
```
tester = ThisTester(model_name, mode, compiler_config=cc)
tester.required_pcc = 1.0
tester.required_atol = 12
```

AleksKnezevic:
you can set `tester.required_atol = 12`, don't set the required pcc to 1.0 since that will not pass on hw.

AleksKnezevic (approved):

Issue #256, opened 2025-02-03
Improve record_property usage


AleksKnezevic:
We want to use record_property in our model tests to record status and pass to superset. We want to be able to populate the Test(BaseModel) pydantic model seen [here](https://github.com/tenstorrent/tt-github-actions/blob/main/.github/actions/collect_data/src/pydantic_models.py). 

Please also add tags for PCC and ATOL.

PR #283, opened 2025-02-05
Improve record_property usage (#256)

jameszianxuTT:
### Ticket
#256 

### Problem description
Current Pytest **record_property** usage has a lot of code duplication and does not expose model evaluation results (PCC, ATOL). This populates some missing fields needed for compliance with the [Pydantic Test model spec](https://github.com/tenstorrent/tt-github-actions/blob/main/.github/actions/collect_data/src/pydantic_models.py).

### What's changed
- Create a tag cache to store "tag" records, and flush them with tester.finalize() at the end of each test
- Extract actual PCC and ATOLs from verify_against_golden
- Move record_property calls into ModelTester, as side effects of other internal class functions to reduce code duplication
  - This pattern is less functionally pure but reduces the overhead for creating new tests, and adds flexibility when adding or changing the telemetry schema (as defined in the Pydantic Test model spec)

### Checklist
- [ ] New/Existing tests provide coverage for changes (in progress)

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -237,7 +237,6 @@ class ModelTester:
             self.required_pcc,
             self.required_atol,
             self.relative_atol,
-            return_pccs_atols=True,
         )
         self.record_tag_cache["pccs"] = pccs
         self.record_tag_cache["atols"] = atols
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -17,6 +17,7 @@ import json
 from onnx import version_converter
 from pathlib import Path
 from tt_torch.tools.verify import verify_against_golden
+from datetime import datetime, timezone
 
 
 class ModelTester:
@@ -61,12 +62,13 @@ class ModelTester:
         self.compiler_config = compiler_config
         self.compiler_config.model_name = model_name
         self.record_property = record_property_handle
-        self.record_tag_cache = (
-            {}
-        )  # Stores all tags to be pushed to the test report at test completion
+        self.record_tag_cache = {}  # Holds for tags to be written out at finalize()
 
         self.record_property("model_name", model_name)
-        self.record_property("mode", mode)
+        self.record_property("frontend", "tt-torch")
+
+        # configs should be set at test start, so they can be flushed immediately
+        self.record_property("config", {"compiler_config": compiler_config.to_dict()})
 
     def _load_model(self):
         raise NotImplementedError(
@@ -265,15 +267,19 @@ class ModelTester:
         return outputs
 
     def test_model(self, on_device=True):
+        if self.mode not in ["train", "eval"]:
+            raise ValueError(f"Current mode is not supported: {self.mode}")
+
         ret = None
+        self.record_property("start_timestamp", datetime.now(timezone.utc).isoformat())
+
         if self.mode == "train":
             ret = self.test_model_train(on_device)
         elif self.mode == "eval":
             ret = self.test_model_eval(on_device)
-        else:
-            raise ValueError(f"Current mode is not supported: {self.mode}")
 
-        self.record_property("torch_ttnn", (self, ret))
+        self.record_property("end_timestamp", datetime.now(timezone.utc).isoformat())
+
         return ret
 
     def flush_tag_cache_to_record(self):
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -17,7 +17,6 @@ import json
 from onnx import version_converter
 from pathlib import Path
 from tt_torch.tools.verify import verify_against_golden
-from datetime import datetime, timezone
 
 
 class ModelTester:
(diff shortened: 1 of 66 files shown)


jameszianxuTT:
There is an existing codepath to use the Pydantic Test spec in [tt-github-actions](https://github.com/tenstorrent/tt-github-actions/blob/0afa9aa4ad80f302aaa2d40e7cab9ff9d68f3c3e/.github/actions/collect_data/src/parsers/python_pytest_parser.py#L48), which auto-populates some fields from existing metadata. 

Is the purpose of this issue to align tt-torch's junitxml output to be consumed by the tt-github-actions API via python_pytest_parser (or similar?)

- in which case, that additional context will inform which Test fields need to be populated by tt-torch, and which will be populated by the consuming framework.

jameszianxuTT:
This draft is incomplete, I just wanted some feedback before committing to this implementation. 

I only changed the **MNIST** testcase as a reference with added PCC/ATOL tags, to demonstrate the minimal changes required to each model test function.

tests/utils.py
```diff
@@ -253,9 +265,19 @@ def test_model_eval(self, on_device=True):
...
         if self.mode == "train":
-            return self.test_model_train(on_device)
+            ret = self.test_model_train(on_device)
         elif self.mode == "eval":
-            return self.test_model_eval(on_device)
+            ret = self.test_model_eval(on_device)
         else:
             raise ValueError(f"Current mode is not supported: {self.mode}")
+
+        self.record_property("torch_ttnn", (self, ret))
```
AleksKnezevic, line 276: we don't need "torch_ttnn"

tests/utils.py
```diff
@@ -59,6 +60,13 @@ def __init__(
             compiler_config = CompilerConfig()
         self.compiler_config = compiler_config
         self.compiler_config.model_name = model_name
+        self.record_property = record_property_handle
+        self.record_tag_cache = (
+            {}
+        )  # Stores all tags to be pushed to the test report at test completion
+
+        self.record_property("model_name", model_name)
+        self.record_property("mode", mode)
```
AleksKnezevic, line 69: "mode" is incompatible, these properties need to be parsed into this pydantic [schema](https://github.com/tenstorrent/tt-github-actions/blob/0afa9aa4ad80f302aaa2d40e7cab9ff9d68f3c3e/.github/actions/collect_data/src/pydantic_models.py#L14)

AleksKnezevic (changes requested):

tests/utils.py
```diff
@@ -253,9 +267,23 @@ def test_model_eval(self, on_device=True):
         return outputs
 
     def test_model(self, on_device=True):
+        if self.mode not in ["train", "eval"]:
+            raise ValueError(f"Current mode is not supported: {self.mode}")
+
+        ret = None
+        self.record_property("start_timestamp", datetime.now(timezone.utc).isoformat())
```
AleksKnezevic, line 274: I thought this was already be managed by junit test?

AleksKnezevic (approved):

tests/utils.py
jameszianxuTT: It is managed by JUnitTest, but the way timestamps and test durations are written into the XML is incompatible with how the [parser consumes it](https://github.com/tenstorrent/tt-github-actions/blob/0afa9aa4ad80f302aaa2d40e7cab9ff9d68f3c3e/.github/actions/collect_data/src/parsers/python_pytest_parser.py#L73-L78). 

(For reference) The automatic JUnitTest timestamps are written out as a starting timestamp attribute for the testsuite, followed by testcases with test durations as attributes.

```xml
<testsuite name="pytest" errors="0" failures="0" skipped="1" tests="2" time="14.112" timestamp="2025-02-06T22:00:18.311804+00:00" hostname="yyz-lab-72-special-jameszianxu-for-reservation-19509">
	<testcase classname="tests.models.mnist.test_mnist" name="test_mnist_train[full-train]" time="0.001">
         [...]
	<testcase classname="tests.models.mnist.test_mnist" name="test_mnist_train[full-eval]" time="13.552">

```

Changing the parser to accept the existing format is possible but would either:
1. Cause a breaking change that might affect other repos using it downstream. 
2. Require additional complexity to handle the multiple "places" the timestamp could be and parsing logic to extract out starting/ending timestamps from the provided format

I figure the redundancy of providing this information is acceptable given the work/cruft required in the alternative case.

AleksKnezevic: Makes sense. I was a little surprised that other fe's manually record this info, but turns out they [do](https://github.com/tenstorrent/tt-forge-fe/blob/fb10a81b6852a035f5ea1e534a4f7aae4fff905d/conftest.py#L14). Might be a little nicer to do in conftest (as in the link) as that would get all of the time (including, i.e. loading model or pre-processing inputs).

AleksKnezevic (approved):

jameszianxuTT:
Related PR to the upstream parser was created in tt-github-actions: https://github.com/tenstorrent/tt-github-actions/pull/21

tt_torch/tools/verify.py
```diff
@@ -103,6 +104,9 @@ def verify_against_golden(
             )
             passed_atol = False
 
+    if return_pccs_atols:
```
AleksKnezevic, line 107: Sorry, just saw this now. I'd rather the function only have one return signature, and we ignore the unneeded returns elsewhere.

jameszianxuTT: Fixed in [latest commit](https://github.com/tenstorrent/tt-torch/pull/283/commits/925d6486bd6b4dc59a57c89fdc3ac2df201ae592).

AleksKnezevic (approved):

PR #287, opened 2025-02-06
Run nightly tests in parallel

AleksKnezevic:
### Problem description
Nightly tests time out due to running too long

### What's changed
Split up into many more tests (~30) and spread onto N150 and N300. Overhead of initiating a test is ~2 minutes, so this shouldn't use up too much more resources, but will allow nightlies to finish faster.


.github/workflows/run-op-by-op-model-tests.yml
```diff
@@ -15,70 +15,183 @@ jobs:
...
+              "
+          },
+          {
+            runs-on: wormhole_b0, name: "mgp-str", tests: "
+              tests/models/mgp-str-base/test_mgp_str_base.py::test_mgp_str_base
+              "
+          },
+          {
+            runs-on: wormhole_b0, name: "musicgen", tests: "
+              tests/models/musicgen/test_musicgen.py::test_musicgen
```
jameszianxuTT, line 180: I think this should be tests/models/musicgen_small/test_musicgen_small.py::test_musicgen_small ?

jameszianxuTT (commented):
I went through the changelist and I think it's just the musicgen test that was not correctly moved

.github/workflows/run-op-by-op-model-tests.yml
AleksKnezevic: Awesome, thanks for catching this!

jameszianxuTT (approved):

mmanzoorTT (approved):

.github/workflows/run-op-by-op-model-tests.yml
jameszianxuTT: Just as a callback to this, I added [a test](https://github.com/tenstorrent/tt-torch/blob/e8bdcf9846bc547aa263347065005658cb449ccd/tt_torch/tools/ci_verification.py#L13-L18) to onPR CI to scan workflow matrices for invalid tests and duplicated, based on the work I did for benchmarking.

PR #311, opened 2025-02-18
New tt-mlir installation support

nsmithtt:
Adopt the new way that tt-mlir builds and links:
- Packages all of tt-mlir + downstream tt-metal libs and files into installation prefix for easy deployment
- Statically links `libTTMLIRCompiler.so` and `libTTMLIRRuntime.so`, now FE's only need to link against these 2 binaries and no other deps.

See tt-mlir PR for more information:
https://github.com/tenstorrent/tt-mlir/pull/2108

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-set(TT_MLIR_VERSION "947aed03c54a92612c9e3bb2f172c252f867b13f")
+set(TT_MLIR_VERSION "bfb88ddfac2d3aaf2ab0525ff177bdeadb2d34e9")
 
 if (BUILD_TOOLCHAIN)
     cmake_minimum_required(VERSION 3.20)
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-set(TT_MLIR_VERSION "fa8c589842b557042f8c554120b24fdae11d5fd6")
+set(TT_MLIR_VERSION "947aed03c54a92612c9e3bb2f172c252f867b13f")
 
 if (BUILD_TOOLCHAIN)
     cmake_minimum_required(VERSION 3.20)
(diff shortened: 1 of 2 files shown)


third_party/CMakeLists.txt
```diff
@@ -56,8 +56,9 @@ else()
           -DTT_RUNTIME_ENABLE_TTMETAL=OFF
           -DTTMLIR_ENABLE_STABLEHLO=ON
           -DTTMLIR_ENABLE_RUNTIME=ON
+          -DTTMLIR_ENABLE_BINDINGS_PYTHON=OFF
```
AleksKnezevic, line 59: is this going to break our TTRT build?

third_party/CMakeLists.txt
```diff
@@ -56,8 +56,9 @@ else()
...
           -DTT_RUNTIME_DEBUG=${TT_RUNTIME_DEBUG}
-          -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
+          -DCMAKE_INSTALL_PREFIX=${TTTORCH_SOURCE_DIR}/install
```
AleksKnezevic, line 61: When running a whl build skbuild uses a different `CMAKE_INSTALL_PREFIX`. Why don't you pass it straight to tt-mlir?

third_party/CMakeLists.txt
nsmithtt: I don't think so, the python bindings are for MLIR python bindings.

third_party/CMakeLists.txt
nsmithtt: So ExternalProjectAdd is like a subprocess that goes off and runs an entirely separate cmake process.  It runs this "install" step during your tt-torch build, i.e. _not_ tt-torch install.  So the pattern I used for the other FEs too is to temporarily install tt-mlir to a separate location, and then add a copy step from that separate loc to the real CMAKE_INSTALL_PREFIX.

AleksKnezevic: ok, sounds good to me, thanks!

AleksKnezevic (approved):

vmilosevic (approved):

Issue #313, opened 2025-02-18
Fix op by op parser (Feb 17)


ddilbazTT:
op by op parser is [broken](https://github.com/tenstorrent/tt-torch/actions/runs/13383022786/job/37421660488)
To replicate issue:
- get latest main branch
- untar test results: 
```
  rm artifact.tar
  for a in `ls -1 *.tar`; do tar -xvf $a; rm $a; done
```
- parse tests:
```
  source env/activate
  export LD_LIBRARY_PATH="/opt/ttmlir-toolchain/lib/:/__w/tt-torch/tt-torch/install/lib:/__w/tt-torch/tt-torch/build/lib:./lib/:${LD_LIBRARY_PATH}"
  mkdir -p /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir-build/bin/
  cp /__w/tt-torch/tt-torch/install/lib/ttmlir-opt /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir-build/bin/
  chmod +x /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir-build/bin/ttmlir-opt
  python results/parse_op_by_op_results.py
```

AleksKnezevic:
Caused by restarting test, closing, will re-open if necessary.

PR #314, opened 2025-02-18
Interim fix for op-by-op parser [#313]

ddilbazTT:
Cannot find root cause - will add safeguards to make sure failure happens without interfering with other jobs

### Ticket
#313 

### Problem description
Feb 17 nightly tests rerun attempts 2 and 3 failed during tt-torch/results/parse_op_by_op_results.py -- generate_status_report Rerun attempt 1 has succeded during generate_status_report. Problem cannot be reproduced but added safeguard so that generate_status_report does not interfere with the execution of nightly tests. 

### What's changed
- Wrapped generate_status_report in a try/ except block
- Wrapped json file parsing in a try/except block
- Added print statements for easier debugging
- Skipping parsing if json input is not loaded in a dictionary format

### DIFF
--- a/results/parse_op_by_op_results.py
+++ b/results/parse_op_by_op_results.py
@@ -24,92 +24,110 @@ def find_json_files(directory="results"):
 
 
 def generate_status_report():
-    json_files = find_json_files()
-    status_report = {}
-    model_names = []
+    try:
+        json_files = find_json_files()
+        status_report = {}
+        model_names = []
 
-    for json_file in json_files:
-        # Get model name from the parent directory of the json file
-        model_name = (
-            json_file.strip("_unique_ops.json")
-            .split("/")[-1]
-            .split(" ")[0]
-            .split("test")[-1]
-        )
-        if len(model_name) > 28:
-            model_name = model_name[:28]
-        id = 1
-        while model_name in model_names:
-            model_name = model_name + f"_{id}"
-            id += 1
+        for json_file in json_files:
+            try:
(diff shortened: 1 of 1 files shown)


results/parse_op_by_op_results.py
```diff
@@ -843,5 +861,8 @@ def default(shlo_op, md_data):
 
 
 if __name__ == "__main__":
-    generate_status_report()
```
AleksKnezevic, line 846: is the order not reversed now? Does that make a difference?

results/parse_op_by_op_results.py
```diff
@@ -24,92 +24,110 @@ def find_json_files(directory="results"):
...
+                            status_percentages[status] = count / total_ops * 100
+                        elif status == 7:
+                            status_percentages[status] = count / total_ops * 100
+                        else:
+                            # For other statuses, calculate as usual
+                            status_percentages[status] = (
+                                (count / total_ops) * 100 if total_ops > 0 else 0
+                            )
+                status_report[model_name] = status_percentages
+            except Exception as e:
```
AleksKnezevic, line 91: We should still fail the job if this doesn't happen, perhaps we can add a test to see if any jsons failed and if they did, assert at the end.

results/parse_op_by_op_results.py
ddilbazTT: It is reversed because I do not want process_json_files to be killed by generate_status_report. try-except block should handle this but added it for extra measure.

PR #318, opened 2025-02-19
Wheel build

AleksKnezevic:
### Problem description
We need a wheel build for the release

### What's changed
Added wheel build
Changed CI to use wheel
Copy only wheels instead of entire venv

### Checklist


AleksKnezevic:
The changes to cmake files are needed for skbuild. It relies on a pre-defined location to install all libraries so it can package them together in a wheel.

LPanosTT (approved):

.github/workflows/generate-model-report.yml
```diff
@@ -55,17 +55,14 @@ jobs:
       working-directory: ${{ steps.strings.outputs.install-output-dir }}
       run: |
         tar xvf artifact.tar
+        mkdir -p ../dist
```
ddilbazTT, line 58: I think it's better to use reusable strings in paths than relative paths.

.github/workflows/run-build.yml
```diff
@@ -95,36 +83,18 @@ jobs:
...
-      working-directory: ${{ steps.strings.outputs.install-output-dir }}
-      run: |
-        mkdir env
-        cp -r ${{ steps.strings.outputs.work-dir }}/env/* env
-
-    - name: 'Tar install directory'
-      shell: bash
-      working-directory: ${{ steps.strings.outputs.install-output-dir }}
-      run: tar cvf artifact.tar .
+        mkdir artifacts
```
ddilbazTT, line 90: Again, I think it's better to use reusable strings and point to artifacts with them.

.github/workflows/run-e2e-tests.yml
```diff
@@ -65,24 +65,18 @@ jobs:
...
-      working-directory: ${{ steps.strings.outputs.install-output-dir }}
-      run: tar xvf artifact.tar
-
-    - name: make tt-metal and env directories
       shell: bash
       working-directory: ${{ steps.strings.outputs.install-output-dir }}
       run: |
-        mkdir -p ${{ steps.strings.outputs.work-dir }}/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal
-        mkdir -p ${{ steps.strings.outputs.work-dir }}/env
+        tar xvf artifact.tar
```
ddilbazTT, line 71: I think after tar, we should have a step where we print out all that has been extracted. This helps greatly for debugging for future references. Then everything with mkdir etc could come as a next step. I think one of the workflows has an example for this

.github/workflows/run-tests.yml
```diff
@@ -106,70 +100,69 @@ jobs:
...
-           --junit-xml=${{ steps.strings.outputs.test_report_path_models }} \
-           --cov=tests/models --cov-append
-
-    - name: Upload Test Report Models
-      uses: actions/upload-artifact@v4
-      if: success() || failure()
-      with:
-        name: test-reports-models-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
-        path: ${{ steps.strings.outputs.test_report_path_models }}
+    # - name: Run Supported Models
```
ddilbazTT, line 103: Now the tests appear to be commented out, is that intended?

AleksKnezevic: Yeah, there are a couple of commits that will be reverted. I wanted to be able to quickly iterate/run nightlies.

.github/workflows/run-e2e-tests.yml
AleksKnezevic: That's a good idea. github has a way of enabling additional [debug info](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/enabling-debug-logging). I'll add that in a subsequent PR.

.github/workflows/run-build.yml
AleksKnezevic: I find the reusable strings very clunky and hard to follow (since the names are very verbose (i.e. `steps.strings.outputs.install-output-dir`) and they're defined far from where they're used), and was planning on modifying all the ci yamls to a similar flow to this once this goes in. That said, that's just one person's opinion, so I'm happy to go with whatever the consensus is. I'll start a slack thread.

AleksKnezevic: The people have spoken and they agree with you 😄, I'll push an update.

setup.py
```diff
@@ -1,41 +1,108 @@
...
+                "tt-mlir",
+                "src",
+                "tt-mlir-build",
+                "bin",
+                "ttmlir-opt",
+            )
         )
-        subprocess.check_call(
-            ["cmake", "--build", "."] + build_args, cwd=self.build_temp
+        self.copy_file(ttmlir_opt, install_path)
```
nsmithtt, line 39: Why do you need ttmlir opt in the wheel? It should be a dev only tool

setup.py
```diff
@@ -1,41 +1,108 @@
...
+                    os.path.join(relative_folder, item).startswith(ignore)
+                    for ignore in tt_metal_folders_to_ignore
+                )
+            ]
+
+            return ignored_items
+
+        install_path = os.path.join(self.install_dir, "tt_torch_tt_metal_libs")
+        os.makedirs(install_path, exist_ok=True)
+        shutil.copytree(metal_dir, install_path, dirs_exist_ok=True, ignore=ignore)
```
nsmithtt, line 77: You shouldn't need to copy metal tree, this is handled by tt-mlir install target, see comments below.

third_party/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ else()
           -DTTMLIR_ENABLE_RUNTIME=ON
           -DTTMLIR_ENABLE_BINDINGS_PYTHON=OFF
           -DTT_RUNTIME_DEBUG=${TT_RUNTIME_DEBUG}
-          -DCMAKE_INSTALL_PREFIX=${TTTORCH_SOURCE_DIR}/install
+          -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
```
nsmithtt, line 61: I don't think this will work as expected, this will end up creating an install dir + mlir installation during your torch build which is not desirable.  I should have added a step below, same as forge-fe has, that installs this tmp install dir. 

Lines 79 + 80 below should read:
```cmake
install(DIRECTORY ${TTMLIR_INSTALL_PREFIX}/ DESTINATION "${CMAKE_INSTALL_PREFIX}" USE_SOURCE_PERMISSIONS)
```

This will correctly install all tt-mlir libs + tt-metal libs + tt-metal sources that are needed.

See forge-fe for reference: https://github.com/tenstorrent/tt-forge-fe/blob/main/third_party/CMakeLists.txt

setup.py
AleksKnezevic: we use it to extract errors from the compiler in post-processing, that is, we store the latest graph compiled to (i.e. shlo or ttir) and we run the subsequent pipeline and record the error. This was necessary because tt-torch wasn't always great at recovering from compile errors, but it is much more robust now, so we can try doing it at runtime, as opposed to post-processing.

Issue #254, opened 2025-02-03
op-by-op override clunky


AleksKnezevic:
Running individual tests op-by-op from the command line requires both setting the argument as well as providing the override, i.e.:
```bash
pytest -svv tests/models/torchvision/test_torchvision_object_detection.py::test_torchvision_object_detection[op_by_op-retinanet_resnet50_fpn-eval] --op_by_op
```

 This is needlessly clunky, and should work with simply:

```bash
pytest -svv tests/models/torchvision/test_torchvision_object_detection.py::test_torchvision_object_detection[op_by_op-retinanet_resnet50_fpn-eval]
```

Please fix conftest to allow this.

PR #319, opened 2025-02-20
Fix op_by_op override clunky

jameszianxuTT:
### Ticket
#254

### Problem description
The --op_by_op override previously ran only op_by_op tests if present, and only full model (non op_by_op) tests if absent. 

### What's changed
Now, there will be no effect from omitting the op_by_op option instead of the previous implied behaviour, which would assume you wanted to run full models only and NOT run any op_by_op tests when the op_by_op option was absent.

### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,34 +39,24 @@ def pytest_addoption(parser):
 
 def pytest_collection_modifyitems(config, items):
     # If --op_by_op flag is set, filter out tests with op_by_op=False
+
     selected_items = []
+    using_op_by_op = config.getoption("--op_by_op")
+
     for item in items:
         # Check if the test has a parameter called 'op_by_op'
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
There are two behaviours that were previously implemented:
- Filter out op_by_op tests when --op_by_op was absent (negative filter)
- Filter only op_by_op tests when --op_by_op was present (positive filter)

The positive filter was used in CI and its behaviour is retained. This means you can specify a set of tests to run, set --op_by_op, and expect that only the op_by_op tests in that set to be run. 

The negative filter was not used and its behaviour is removed in this PR. The two places in CI that execute full model tests specify only the full model test cases explicitly, and leave nothing to filter out.

LPanosTT (approved):

AleksKnezevic (approved):

PR #321, opened 2025-02-20
Uplift LLVM and add torch-mlir as third-party

LPanosTT:
### What's changed
Uplift tt-mlir to version which contains uplifted llvm
Add torch-mlir as third party as we may be developing on it soon


mmanzoorTT (approved):

requirements.txt
```diff
@@ -36,3 +34,4 @@ opencv-contrib-python
 setuptools
 wheel
 scikit-build
+nanobind
```
jameszianxuTT, line 37: What requires this dependency? It looks like an alternative to pybind but we already include that (For my own knowledge)

brataTT (approved):

jameszianxuTT (approved):

PR #327, opened 2025-02-20
First pass - CICD testcase schema change preparation

jameszianxuTT:
### Ticket
#256 

### Problem description
CICD testcase schema will change and obscure model name and frontend values.

### What's changed
- Pack model/fe metadata into tags in preparation for CICD testcase schema change
- This retains record_property calls for model name and frontend for now. This will be 
removed in a future pass once the pydantic model and python pytest parser is amended.

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -66,6 +66,9 @@ class ModelTester:
         self.record_property("model_name", model_name)
         self.record_property("frontend", "tt-torch")
 
+        self.record_tag_cache["model_name"] = model_name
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
I am avoiding any breaking changes and doing this in a two-pass PR. 

1. I will duplicate the frontend and model name into the tags so that they will land without a chance of breaking CI
2. Once the pydantic model and pytest parser in tt-github-actions is amended, I will remove the old duplicate changes

AleksKnezevic (approved):

Issue #203, opened 2025-01-15
op-by-op SHLO compiler


AleksKnezevic:
Please crate some infrastructure to compile a stableHLO graph op-by-op through tt-mlir.

ddilbazTT:
The way to install stablehlo is:

```
pip install --pre torch-mlir torchvision
pip install stablehlo -f https://github.com/openxla/stablehlo/releases/expanded_assets/dev-wheels
```
The way to use stablehlo is:

```
import tt_mlir
import mlir.ir as ir
import mlir.dialects.stablehlo as stablehlo
from mlir.ir import Context, Location

with ir.Context() as ctx:
    stablehlo.register_dialect(ctx)
    module = ir.Module.parse(MODULE_STRING)
```

ddilbazTT:
Hi @AleksKnezevic This is the basic implementation so far:

```
# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
#
# SPDX-License-Identifier: Apache-2.0

### PRE-REQUISITES
#   pip install --pre torch-mlir torchvision
#   pip install stablehlo -f https://github.com/openxla/stablehlo/releases/expanded_assets/dev-wheels

import mlir
import tt_mlir
import mlir.ir as ir
from mlir.ir import Context, Location, Module
import mlir.dialects.stablehlo as stablehlo
from typing import List, Dict, Any

modules = {}


def get_module_from_str(module_str: str):
    module = None
    with ir.Context() as ctx:
        stablehlo.register_dialect(ctx)
        module = Module.parse(module_str)
    return module


def lower_stablehlo_to_ttnn(stablehlo_ir: str):
    module = get_module_from_str(stablehlo_ir)
    try:
        ttir = tt_mlir.compile_stable_hlo_to_ttir(module.operation.get_asm())
        print("ttir done")
        try:
            binary, ttnn = tt_mlir.compile_ttir_to_bytestream(ttir)
            print("ttnn done")
            return ttnn
        except Exception as e:
            print("Error: ", e)
            return None
    except Exception as e:
        print("Error: ", e)
        return None


def get_ops_in_module(module: mlir.ir.Module):

    # Iterate through all functions in the module
    for func_op in module.body.operations:
        for block in func_op.regions[0].blocks:
            for op in block.operations:
                inputs = {}
                result_type = None
                if not op.name.startswith(("func.", "return")):
                    if (
                        op.name == "stablehlo.pad"
[...]

PR #338, opened 2025-02-21
Stablehlo compiler

ddilbazTT:
### Ticket
#203 

### Problem description
Currently, we only support compilation or execution of torch graphs. We do not support compilation of stablehlo graphs because this would fail at the first unsupported stablehlo op. The solution is to create a module for each op. This can be done through stablehlo support.

### What's changed
| Input\Compile Depth              | Compile Op by Op     | Execute Op by Op     | Execute        | Compile Stablehlo Op by Op      | Other                             |
|----------------------------------|----------------------|----------------------|----------------|---------------------------------|-----------------------------------|
| Stablehlo String                 | _shlo_backend        | _shlo_backend        | _base_backend  | Overwrite to CompileDepth.COMPILE_OP_BY_OP + _shlo_backend                   | No support           |                      
| Torch Graph                      | _torch_backend       | _torch_backend       | _base_backend  | torch_to_shlo + _shlo_backend    | _torch_backend                    |

- `_base_backend` --> `_torch_backend`
- new `_base_backend` is for execution, for other compile depths using `_torch_backend` or `_shlo_backend`
- divided `tt_torch/dynamo/backend.py` into `tt_torch/dynamo/backend.py`, `tt_torch/dynamo/shlo_backend.py`, `tt_torch/dynamo/torch_backend.py`
- old `tt_torch/dynamo/backend.py` contents are mostly moved to `tt_torch/dynamo/torch_backend.py`, `tt_torch/dynamo/executor.py`
- `tt_torch/dynamo/executor.py` provides base executor class to be used by both stablehlo and torch backends
- `env/activate` edited to add support for dependencies

Testing strategy:
- Please see commit history for how I'm testing stablehlo input. I will attach output log when finished.


env/activate
```diff
@@ -40,8 +39,12 @@ else
 
     cp dist/*.whl $TT_TORCH_HOME/dist
     cd $TT_TORCH_HOME
+
   fi
   pip install $TT_TORCH_HOME/dist/torchvision*.whl
+  pip install --pre torch-mlir torchvision
```
AleksKnezevic, line 45: what's torchvision for here?

tt_torch/tools/utils.py
```diff
@@ -275,7 +276,7 @@ def post_init(self):
         else:
             torch._dynamo.config.inline_inbuilt_nn_modules = True
 
-    def save_unique_ops(self):
+    def save_unique_ops(self, mode=None):
```
AleksKnezevic, line 279: perhaps default to `mode="torch"` as opposed to None.

tt_torch/dynamo/backend.py
```diff
@@ -566,19 +53,104 @@ def verify_golden_callback(binary, callback_context, op_context):
...
+    elif isinstance(shlo, str):
+        executor = StablehloExecutor(module_str=shlo, compiler_config=compiler_config)
+    else:
+        print("Compiler input not valid", file=sys.stderr)
+        exit(1)
+    if gm is not None:
+        executor.add_gm(gm, graph_constants)
+    return executor
+
+
```
AleksKnezevic, line 70: This section is duplicated. Can you move it into a helper function?

tt_torch/tools/utils.py
ddilbazTT: I could do that. Just want to confirm, by default we want the files to be named  `{self.results_path}{pytest_test}_torch_unique_ops.json` or `{self.results_path}{pytest_test}_unique_ops.json` then? Right now, the way it is:
default --> `{self.results_path}{pytest_test}_unique_ops.json`
torch --> `{self.results_path}{pytest_test}_torch_unique_ops.json`
stablehlo --> `{self.results_path}{pytest_test}_stablehlo_unique_ops.json`

tt_torch/dynamo/shlo_backend.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def add_gm(self, gm: torch.fx.GraphModule, graph_constants):
+        assert (
+            self.compiler_config.compile_depth
+            == CompileDepth.COMPILE_STABLEHLO_OP_BY_OP
+        ), "gm can only be added in COMPILE_STABLEHLO_OP_BY_OP mode"
+        self.gm = gm
+        self.graph_constants = tuple(graph_constants)
+
+    def gm_op_by_op(self, *inputs):
```
AleksKnezevic, line 207: maybe `shlo_op_by_op`

tt_torch/dynamo/shlo_backend.py
```diff
@@ -0,0 +1,430 @@
...
+            if input_type in self.type_conversion.keys():
+                new_inputs = new_inputs + (
+                    (input.to(dtype=self.type_conversion[input_type])),
+                )
+                continue
+
+            # No conversion required.
+            new_inputs = new_inputs + ((input),)
+        inputs = new_inputs
+        if self.compiler_config.compile_depth == CompileDepth.EXECUTE:
```
AleksKnezevic, line 411: isn't this handled by base backed?

tt_torch/dynamo/shlo_backend.py
ddilbazTT: So this is actually the same function from torch_backend, but stripped. This is actually going through a torch graph op by op, and only called during COMPILE_STABLEHLO_OP_BY_OP, since it's implied the input is a torch graph. Doing so, we create 2 json dumps - one for stablehlo and one for torch, during one run.

tt_torch/dynamo/shlo_backend.py
ddilbazTT: This is base_backend, which returns executor. This is in parallel with torch executor.

```
def _base_backend(gm_or_shlo, example_inputs, compiler_config):
    # Called during EXECUTE
    # input is a torch graph
    if isinstance(gm_or_shlo, torch.fx.GraphModule):
        shlo, executor, gm, graph_constants = torch_to_shlo(
            gm_or_shlo, example_inputs, compiler_config
        )
    # input is a stablehlo string module
    elif isinstance(gm_or_shlo, str):
        shlo = parse_module_from_str(gm_or_shlo)
        executor = StablehloExecutor(
            parsed_module=shlo, compiler_config=compiler_config
        )
    else:
        assert False, "Compiler input not valid"

    binary = shlo_to_flatbuffer(shlo, compiler_config)
    executor.set_binary(binary)
    return executor
```

tt_torch/dynamo/backend.py
ddilbazTT: I think my latest commit should address this. By complicated, I assumed you meant the `isinstance` checks. I agree. I now check and assign parsed_module when initializing the Executor object.

ddilbazTT:
### **Latest Refactoring:**

I talked with Aleks on Friday and there was some refactoring decisions made.

**Base vs Torch vs Stablehlo Backends:**
- Torch and Stablehlo Backends refer to `COMPILE_OP_BY_OP`/ `EXECUTE_OP_BY_OP` 
- Base backend refers to Compile Depths: `TORCH_FX`, `STABLEHLO`, `TTNN_IR`, `EXECUTE`

**Executor/ OpByOpExecutor/ StablehloExecutor/ TorchExecutor**
- `Executor` is the base class which runs the binary or the graph, depending on compile depth. 
- `OpByOpExecutor` inherits from base `Executor`. Allows op-by-op compilation/ execution of stablehlo or torch ops through `compile_op` / `run_op` functions. 
- `StablehloExecutor` inherits from `OpByOpExecutor`. It parses stablehlo module and creates a module for each op  in `get_ops_in_module` to allow op-by-op compilation/ execution. Implements `get_stable_hlo_graph` and `shlo_op_by_op`. 
- `TorchExecutor` inherits from `OpByOpExecutor`. Implements `get_stable_hlo_graph` and `gm_op_by_op`. 

**Shared Functions between StablehloExecutor and TorchExecutor**
- `transform_input` , `pre_process_inputs`: helper functions for input manipulation/ preparation 
- `compile_op`: lowers a node/ op to stablehlo, then passes through pybinded functions to lower to ttnn/ binary. 
- `run_op`: given a binary and inputs, runs the binary with the inputs
- `get_stable_hlo_graph`: this function is implemented differently in both backends. This is expected since the starting point of TorchExecutor is a torch node, whereas for `StablehloExecutor` is a Stablehlo module. They return an mlir stablehlo `module` object and an `op` element. 

**shlo_op_by_op/ gm_op_by_op**
- These functions are very similar to one another, but one runs a stablehlo graph op by op, the other runs a torch graph op by op. Both call `compile_op` and `run_op` commands. 
- For stablehlo, I implemented a random input generator to be able to run the op. `generate_random_inputs_for_shlo`

@mrakitaTT @nsmithtt @kmitrovicTT re the thread on Slack (slack is down atm for me) 
We had discussed implementing a shlo parser in tt-mlir similar to one I wrote here. I started working on that, but I had difficulty pip installing in tt-mlir. So I don't mind if you guys want to copy over my code there or if you want me to do it. But I don't think we should be blocked by that for tt-torch. 

For reference, this is the code block which breaks a stablehlo module in sub modules:
```
def get_ops_in_module(self, module):
        for func_op in module.body.operations:
            for block in func_op.regions[0].blocks:
                for op in block.operations:
                    if op.name.startswith(("func.", "return")):
                        continue

                    inputs = {
                        operand.get_name(): str(operand.type) for operand in op.operands
                    }
                    args_str = ", ".join(f"{key}: {typ}" for key, typ in inputs.items())

                    # Handle multiple results in the operation
                    result_names = [str(result.get_name()) for result in op.results]
                    result_types = [str(result.type) for result in op.results]

                    # Construct the function signature based on the number of results
                    if len(result_names) == 1:
                        result_str = f"{result_types[0]}"
                        return_stmt = f"return {result_names[0]} : {result_types[0]}"
                    else:
                        result_str = f"({', '.join(result_types)})"
                        return_stmt = f"return ({', '.join(result_names)}) : ({', '.join(result_types)})"
                    # Build the new module string
                    new_module_str = f"""module {{
        func.func @main({args_str}) -> {result_str} {{
            {str(op)}
            {return_stmt}
        }}
    }}"""

                    opObj = StablehloOp(
                        model_name=self.compiler_config.model_name,
                        op_id=", ".join(result_names),
                        original_shlo=str(op),
                    )
                    opObj.add_stable_hlo_graph(new_module_str)
                    self.sub_ops.append(opObj)

```

After rebasing this branch and concluding testing, I will turn this draft PR into an actual PR.

kmitrovicTT:
@ddilbazTT definitely don't block tt-torch effort on extracting the common stuff in tt-mlir. If you see something can be put in common repo without many difficulties, then do it. 

If I add more common infra in tt-mlir, I'll integrate your code in it and ping you for review.

PR #360, opened 2025-02-21
Calculate Model Metric Aggregates

jameszianxuTT:
### Ticket
#256 

### Problem description
Some models produce multiple output tensors, and thus multiple
pcc / atol values which are pushed into tags as as JSON list. 

It is possible, but adds significant complexity to construct a SQL query to 
parse and aggregate over a JSON list.

### What's changed
- Add pcc and atol MIN and AVG values and push aggregate metrics into tag cache

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -281,11 +281,16 @@ class ModelTester:
         metric_list = self.record_tag_cache.get(metric_key, [])
         avg_metric = default_value
         min_metric = default_value
+        max_metric = default_value
+
         if metric_list:  # check null or empty list
             avg_metric = sum(metric_list) / len(metric_list)
             min_metric = min(metric_list)
+            max_metric = max(metric_list)
+
         self.record_tag_cache["avg_" + metric_key] = avg_metric
         self.record_tag_cache["min_" + metric_key] = min_metric
+        self.record_tag_cache["max_" + metric_key] = max_metric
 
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
The existing [PCC verification logic](https://github.com/tenstorrent/tt-torch/blob/3120d48984bd569c462e068c2a990066060cf7c0/tt_torch/tools/verify.py#L14) requires all output tensors to pass a PCC check, and fails the entire model test on PCC if any checks fail. 

Taking the MIN (PCCs) as the "single PCC metric" for a model test is reductive, but reflects this verification logic.

tests/utils.py
```diff
@@ -275,8 +275,27 @@ def test_model(self, on_device=True):
...
 
+    def record_aggregate_model_metric(self, metric_key, default_value=-1):
+        # read a metric from the tag cache and write out the average and min values
+
+        metric_list = self.record_tag_cache.get(metric_key, [])
+        avg_metric = default_value
+        min_metric = default_value
+        if metric_list:  # check null or empty list
+            avg_metric = sum(metric_list) / len(metric_list)
+            min_metric = min(metric_list)
```
AleksKnezevic, line 286: for atol we probably want max, right?

jameszianxuTT: Good point, I'll add that in.

AleksKnezevic (approved):

PR #361, opened 2025-02-21
Create getting_started.md

LPanosTT:
Create getting_started.md
 - Contains instructions on how to install the wheel, and torch-mlir
 - Instructions on how to compile a model
   - includes basic CompilerConfig example

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -1,0 +1,88 @@
+# Getting Started
+
+## Creating a Virtual Environment (skip if you already have one)
+
+Create a virtual environment if you do not already have one in your project:
+```sh
+python -m venv myvenv
+```
+This will create a virtual environemnt in the folder `myvenv` in the current directory.
+
+Activate the environemnt:
+```sh
+source myvenv/bin/activate
+```
+
+## Installing tt_torch
+
+### Installation Notes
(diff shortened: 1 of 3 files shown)


docs/src/getting_started.md
```diff
@@ -0,0 +1,67 @@
...
+model = torch.compile(model, backend=backend)
+
+inputs = ...
+
+outputs = model(inputs)
+```
+
+### Configuring the Compiler
+
+Some features must be enabled using `CompilerConfig`. Here is how to enable one of those features, constant folding:
```
mmanzoorTT, line 43: This seems odd `Some features must be enabled`; if user must have to enable the features then this should be the default option instead of asking user to enable them. I will suggest to write something like this
```
tt-torch supports multiple flags for compiler configuration using `CompilerConfig` and it is preferred to enable these features.
```

mmanzoorTT (approved):

docs/src/getting_started.md
LPanosTT: Sounds good. I'll add this.

AleksKnezevic: I would actually remove this completely from this file and add it to docs/src/controlling.md. You don't have to describe each option, just say that it exists, where it's defined and that it can be used to control compile flow.

AleksKnezevic (approved):

PR #374, opened 2025-02-26
Call op::verify() methods after importing the inital graph

LPanosTT:
### Problem description
When the FX graph is initially imported to TorchFX IR, the graph isn't verified.

### What's changed
Walk the graph and verify the ops after we import the module.

### DIFF
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -43,6 +43,15 @@ import sys
 import tempfile
 
 
+def verify_ir(module):
+    def verify_op(op):
+        if hasattr(op, "verify"):
+            op.verify()
+        return torch_mlir.ir.WalkResult.ADVANCE
+
+    module.operation.walk(verify_op)
+
+
 class TTContextCache(ContextCache):
     def get_node_location(self, node: torch.fx.Node) -> Optional[Location]:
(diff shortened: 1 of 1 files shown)


brataTT (approved):

tt_torch/dynamo/backend.py
```diff
@@ -587,6 +596,8 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
         dump_info = dump_debug or dump_intermediates == "INFO"
 
     module = import_graph(gm.graph)
+    verify_ir(module)
```
ddilbazTT, line 599: Shouldn't this happen after pass_pipeline, before returning executor?

LPanosTT: This verifies MLIR modules only and this is the first instance of an MLIR module in the compile phase. I implemented it here because `import_stateless_graph` assumes we actually give it an fx graph which is stateless. But if we don't it will still make an IR but it will be invalid and will cause compile issues in later lowering steps.

LPanosTT: I ran into this trying to bring up a model and it took me a long time to realize that there wasn't a bug in torch-mlir, we were just giving it an invalid graph to begin with.

Issue #346, opened 2025-02-21
monilenetv2 onnx


AleksKnezevic:
Add mobilenetv2 onnx test, and add to op-by-op once infra is ready

ddilbazTT:
@LPanosTT this issue is on you right?

kmabeeTT:
Test was added March 10th, and added to full-model-execute on March 26th:

```
git l test_MobileNetV2_onnx.py
2025-03-26 9b1bbed by Brata Choudhury (Author achoudhury@tenstorrent.com) : Fix onnx model tester regressions and add onnx mobilenet to On PR
2025-03-25 155fb9a by GitHub (Author jameszianxu@tenstorrent.com) : Model name deduplication and priority group marking (#502)
2025-03-10 47eec94 by Lewis Panos (Author lpanos@tenstorrent.com) : Added OnnxModelTester. Added onnx mobilenetv2 test
```

Anything left to do here other than getting it running in op-by-op flow for good measure? Interestingly it fails with below error (known issue?):

```
WARNING: Input 0 is not contiguous. Converting to contiguous in-place.
                 Always |    FATAL | Program input size mismatch: 2 != 1
LOG_ASSERT @ /localdev/kmabee/tt-torch/third_party/tt-mlir/src/tt-mlir/runtime/lib/ttnn/program.cpp:94: program->inputs()->size() == programInputs.size()
backtrace:
 --- tt::runtime::ttnn::runProgram(tt::tt_metal::distributed::MeshDevice&, tt::runtime::Binary, unsigned int, std::vector<tt::runtime::Tensor, std::allocator<tt::runtime::Tensor> >&)
 --- tt::runtime::ttnn::submit(tt::runtime::Device, tt::runtime::Binary, unsigned int, std::vector<tt::runtime::Tensor, std::allocator<tt::runtime::Tensor> >&)
 --- tt::runtime::submit(tt::runtime::Device, tt::runtime::Binary, unsigned int, std::vector<tt::runtime::Tensor, std::allocator<tt::runtime::Tensor> >&)
 --- /localdev/kmabee/tt-torch/install/lib/tt_mlir.cpython-311-x86_64-linux-gnu.so(+0xc890b) [0x7f863197a90b]
 --- /localdev/kmabee/tt-torch/install/lib/tt_mlir.cpython-311-x86_64-linux-gnu.so(+0xcc615) [0x7f863197e615]
```

kmabeeTT:
[...]

PR #386, opened 2025-03-03
Added OnnxModelTester, other onnx qol infra. Added onnx mobilenetv2 test

LPanosTT:
### Ticket
Mobilenetv2 onnx issue: https://github.com/tenstorrent/tt-torch/issues/346#issue-2869513608

### What's changed
- `verify_module` now accepts a `ModelProto` instead of a filename to a serialized onnx model. 
- Added an OnnxModelTester for making onnx model tests
- Added mobilenetv2 onnx test (SHLO depth)
- `verify_against_golden` now handles assertions on PCC/ATOL checks
    - Prints ❎ instead of ❌ if the check fails but we decided against asserting PCC/ATOL

### DIFF
--- a/tests/models/MobileNetV2/test_MobileNetV2_onnx.py
+++ b/tests/models/MobileNetV2/test_MobileNetV2_onnx.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import onnx
+import requests
+import torchvision.models as models
+from torchvision import transforms
+from PIL import Image
+import os
+
+import pytest
+from tests.utils import OnnxModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth
+
+
+class ThisTester(OnnxModelTester):
+    def _load_model(self):
+        # Load the MobileNetV2 model with updated API
+        self.weights = models.MobileNet_V2_Weights.DEFAULT
+        model = models.mobilenet_v2(weights=self.weights)
+        torch.onnx.export(model, self._load_torch_inputs(), f"{self.model_name}.onnx")
+        model = onnx.load(f"{self.model_name}.onnx")
+        os.remove(f"{self.model_name}.onnx")
+        return model
+
+    def _load_torch_inputs(self):
+        # Define a transformation to preprocess the input image using the weights transforms
+        preprocess = self.weights.transforms()
+
+        # Load and preprocess the image
+        url = "http://images.cocodataset.org/val2017/000000039769.jpg"
+        image = Image.open(requests.get(url, stream=True).raw)
+        img_t = preprocess(image)
+        batch_t = torch.unsqueeze(img_t, 0)
+        return (batch_t,)
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize("op_by_op", [True, False], ids=["op_by_op", "full"])
+def test_MobileNetV2(record_property, mode, op_by_op):
+    model_name = "MobileNetV2"
+    cc = CompilerConfig()
+    cc.compile_depth = CompileDepth.STABLEHLO
+    if op_by_op:
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        assert_pcc=False,
+        assert_atol=False,
+        compiler_config=cc,
+        record_property_handle=record_property,
+    )
+    results = tester.test_model()
+    if mode == "eval":
+        # Print the top 5 predictions
+        _, indices = torch.topk(results[0], 5)
+        print(f"Top 5 predictions: {indices[0].tolist()}")
+
+    tester.finalize()
+
+
+# Empty property record_property
+def empty_record_property(a, b):
+    pass
+
+
+# Main
+if __name__ == "__main__":
+    test_MobileNetV2(empty_record_property)
--- a/tests/onnx/test_basic.py
+++ b/tests/onnx/test_basic.py
@@ -23,11 +23,10 @@ def test_abs():
         model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 21)])
         return model
 
-    onnx.save(create_abs_model(), "abs.onnx")
     cc = CompilerConfig()
     cc.compile_depth = tt_torch.tools.utils.CompileDepth.EXECUTE_OP_BY_OP
     cc.op_by_op_backend = tt_torch.tools.utils.OpByOpBackend.STABLEHLO
-    verify_module("abs.onnx", compiler_config=cc)
+    verify_module(create_abs_model(), compiler_config=cc)
 
 
 def test_add():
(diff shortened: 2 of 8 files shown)


tests/onnx/test_basic.py
```diff
@@ -2,11 +2,13 @@
...
 # SPDX-License-Identifier: Apache-2.0
 import torch
+import onnx
 from torch import nn
 import pytest
+import os
 
 from tt_torch.tools.verify import verify_module
-from tt_torch.tools.utils import CompilerConfig
+from tt_torch.tools.utils import CompilerConfig, CompileDepth
```
brataTT, line 11: nit: Don't see CompileDepth being used here

tt_torch/tools/utils.py
```diff
@@ -18,12 +18,14 @@
...
-    EXECUTE_OP_BY_OP = 6
-    EXECUTE = 7
+    TORCH_ONNX_IR = 2
+    TORCH_FX_IR = 3
+    TORCH_BACKEND_IR = 4
+    STABLEHLO = 5
+    TTNN_IR = 6
+    COMPILE_OP_BY_OP = 7
+    EXECUTE_OP_BY_OP = 8
+    EXECUTE = 9
```
brataTT, line 28: I think we refer to these in our op_by_op parser.
Those may need to be updated.
Not asking for the update to be in the same PR unless this change breaks the parser.

ddilbazTT:
Should we also consider adding v1? It is available on onnx model zoo: https://github.com/onnx/models/tree/main/validated/vision/object_detection_segmentation/ssd-mobilenetv1/model @AleksKnezevic @LPanosTT

AleksKnezevic:
@ddilbazTT, v2 is a customer requested model. We can add v1 if it's available, but it's low priority, so unless it `just works` I would skip it.

tests/models/MobileNetV2/test_MobileNetV2_onnx.py
```diff
@@ -0,0 +1,85 @@
...
+    def _load_model(self):
+        # Load the MobileNetV2 model with updated API
+        self.weights = models.MobileNet_V2_Weights.DEFAULT
+        model = models.mobilenet_v2(weights=self.weights)
+        torch.onnx.export(model, self._load_torch_inputs(), f"{self.model_name}.onnx")
+        model = onnx.load(f"{self.model_name}.onnx")
+        os.remove(f"{self.model_name}.onnx")
+        return model
+
+    def _load_torch_inputs(self):
```
ddilbazTT, line 27: I don't quite understand why `_load_torch_inputs` and `_load_numpy_inputs` are needed, and how they work. It seems `_load_numpy_inputs` gets the inputs from the session, is that how it should work for other tests?

tests/models/MobileNetV2/test_MobileNetV2_onnx.py
```diff
@@ -0,0 +1,85 @@
...
+
+import pytest
+from tests.utils import OnnxModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth
+
+
+class ThisTester(OnnxModelTester):
+    def _load_model(self):
+        # Load the MobileNetV2 model with updated API
+        self.weights = models.MobileNet_V2_Weights.DEFAULT
```
ddilbazTT, line 20: Why don't we use `self.model` instead of `self.weights`? We don't use `weights` in `OnnxModelTester` but we use `model`.

tests/utils.py
```diff
@@ -305,3 +290,82 @@ def finalize(self):
...
+    # Pass this function so we can use superclass __init__ without failure
+    def _load_inputs(self):
+        pass
+
+    def _load_torch_inputs(self):
+        raise NotImplementedError(
+            "This method should be implemented in the derived class"
+        )
+
+    def _load_numpy_inputs(self):
```
ddilbazTT, line 335: I think we should have only `_load_inputs` which takes in torch input, but creates the numpy input based on the session, and uses it in `test_model_eval` under the hood. `_load_numpy_inputs` seem generic to any input where we have the torch input already created by the user.

tt_torch/dynamo/backend.py
```diff
@@ -65,7 +73,9 @@ def _torch_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
     return executor
 
 
-def torch_to_shlo(gm: torch.fx.GraphModule, example_inputs, compiler_config):
+def generate_torch_fx_ir(
```
ddilbazTT, line 76: I think you've reverted torch_to_shlo to the previous backend. I think the torch_to_shlo pipeline made it easier to follow, now we need to call generate_torch_fx_ir and torch_fx_ir_to_torch_backend_ir quite a few times, and base backend is more convoluted. Is there a reason you broke it up?

tt_torch/tools/verify.py
```diff
@@ -175,7 +197,7 @@ def _verify_onnx_module(
     do_assert,
 ):
 
-    sess = InferenceSession(filename)
+    sess = InferenceSession(model_proto.SerializeToString())
```
ddilbazTT, line 200: This is a really useful change, before we were creating a file for each onnx test.

tests/models/MobileNetV2/test_MobileNetV2_onnx.py
LPanosTT: So the `ModelProto` requires a dict of (str, np.ndarray). So we get the proper names from the session the the values are provided by the test. I probably could implement `_load_numpy_inputs` in the base `OnnxModelTester` class since its just a typecast (if torch is bf16) and `.to_numpy`

tests/models/MobileNetV2/test_MobileNetV2_onnx.py
LPanosTT: I copied this from the regular PyTorch model test. The `self.weights` is really just used as a local variable in this function and using `weights` as the variable name instead would have the same effect, except I guess we can drop the reference to it instead of keeping it in the class.

tests/utils.py
LPanosTT: I responded to another comment that basically says this. Read it first 😆

tt_torch/dynamo/backend.py
LPanosTT: I broke it up because I thought it would make it easier to follow. The original `torch_to_shlo` takes a graph module and gives an SHLO one. But there's really a number of compile phases in there. Torch FX graph to torch FX IR, torch FX IR to torch backend IR, torch backend IR to SHLO. I illustrated the compile flow from start to finish in `tt_torch/tools/utils.py line 19` and I thought a simple helper for each step in that illustration would be simpler. Especially since ONNX -> SHLO is different than Torch -> SHLO, but afterward its the same.

LPanosTT: Also because I added a couple extra compile depths to capture those phases. I thought it would be clearer to have a one func per compile phase, which is also one per compile depth.

ddilbazTT: I think we should keep torch_to_shlo, but torch to shlo could call the functions you wrote. It made the backends much easier to follow

tt_torch/onnx_compile/onnx_compile.py
```diff
@@ -23,60 +29,122 @@
...
-from tt_torch.dynamo.backend import dump_module
+from tt_torch.dynamo.backend import (
+    dump_module,
+    torch_backend_ir_to_stablehlo,
+    stablehlo_to_ttir,
+    ttir_to_ttnn_and_binary,
+)
+
+
+class OnnxExecutor:
```
ddilbazTT, line 40: This could inherit from the base executor or we can change the base executor to accommodate onnx.

tt_torch/dynamo/backend.py
ddilbazTT: We discussed this offline. A refactoring done when adding shlo compiler was to remove unneeded compile depths. There could be an onnx specific compile depth not used by the torch backend. Since TORCH_BACKEND_IR will not be needed for torch models, I personally don't think we need to insert a new compile depth here for it to match onnx compile.

ddilbazTT (approved):

Issue #378, opened 2025-02-27
Nightly Tests don't run models in EXECUTE compile depth


jameszianxuTT:
Nightly Tests only run two categories of test, sliced on Compile Depth:
- Execute Op By Op 
- Compile to TTNN_IR 

Full model execution, which is the highest standard for "working" model, should be captured in nightly tests. It is only currently captured in the On PR and On Push workflows.

This also aligns to simplify the query for a unified frontend dashboard status per request in https://github.com/tenstorrent/tt-forge/issues/46

Note - full model execution testcases are characterized by compile_depth EXECUTE with full model evaluation mode

PR #392, opened 2025-03-04
Full Model Execution in Nightlies Improvements

jameszianxuTT:
### Ticket
#378 

### Problem description
- Nightly tests would have failed due to some bugs in #368 
- Some model execution tests were disabled
- Multi-testcase test jobs were self-overwriting a shared
XML and only showing the last test result

### What's changed
- Fixed omissions from #368
- Enabled torchvision tests and split into 3 jobs
- Fixed XML overwrite issue

### Checklist


AleksKnezevic (approved):

jameszianxuTT:
As validation for this change, here is an instance of the [nightly tests running all full model execution](https://github.com/tenstorrent/tt-torch/actions/runs/13659904027).

It is verified that these tests run successfully and produce valid XML representing all testscases run within a test job.

PR #394, opened 2025-03-04
Shlo compiler

ddilbazTT:
**Backend**
- Entry point for torch graph
- Decided that if input is onnx, it should call _shlo_backend or the executors separately from `compile_onnx`

**Base vs Torch vs Stablehlo Backends:**
- Torch and Stablehlo Backends refer to `COMPILE_OP_BY_OP`/ `EXECUTE_OP_BY_OP` 
- Base backend refers to Compile Depths: `TORCH_FX`, `STABLEHLO`, `TTNN_IR`, `EXECUTE`

**Executor/ OpByOpExecutor/ StablehloExecutor/ TorchExecutor**
- `Executor` is the base class which runs the binary or the graph, depending on compile depth. 
- `OpByOpExecutor` inherits from base `Executor`. Allows op-by-op compilation/ execution of stablehlo or torch ops through `compile_op` / `run_op` functions. 
- `StablehloExecutor` inherits from `OpByOpExecutor`. It parses stablehlo module and creates a module for each op  in `get_ops_in_module` to allow op-by-op compilation/ execution. Implements `get_stable_hlo_graph` and `shlo_op_by_op`. 
- `TorchExecutor` inherits from `OpByOpExecutor`. Implements `get_stable_hlo_graph` and `gm_op_by_op`. 

**Shared Functions between StablehloExecutor and TorchExecutor**
- `transform_input` , `pre_process_inputs`: helper functions for input manipulation/ preparation 
- `compile_op`: lowers a node/ op to stablehlo, then passes through pybinded functions to lower to ttnn/ binary. 
- `run_op`: given a binary and inputs, runs the binary with the inputs
- `get_stable_hlo_graph`: this function is implemented differently in both backends. This is expected since the starting point of `TorchExecutor` is a torch node, whereas for `StablehloExecutor` is a Stablehlo module. They return an mlir stablehlo `module` object and an `op` element. 

**shlo_op_by_op/ gm_op_by_op**
- These functions are very similar to one another, but one runs a stablehlo graph op by op, the other runs a torch graph op by op. Both call `compile_op` and `run_op` commands. 
- For stablehlo, I implemented a random input generator to be able to run the op. `generate_random_inputs_for_shlo`

**onnx support**
- Created `onnx_to_stablehlo` function which can be used to pipe to stablehlo backend
- Restructured `_verify_onnx_module` to run with stablehlo backend
- Added many onnx tests to test_basic.py
- Restructured `compile_onnx` function:

```
def compile_onnx(module: onnx.ModelProto, example_inputs, options=None):

    if options is None:
        options = CompilerConfig()
    options.op_by_op_backend = OpByOpBackend.STABLEHLO
    options.typecast_inputs = False
    options.apply_environment_overrides()
    if (
        options.compile_depth == CompileDepth.COMPILE_OP_BY_OP
        or options.compile_depth == CompileDepth.EXECUTE_OP_BY_OP
    ):
        module = onnx_to_stablehlo(module, options)
        executor = StablehloExecutor(module=module, compiler_config=options)
        return executor(*example_inputs)
    elif options.compile_depth == CompileDepth.EXECUTE:
        module = onnx_to_stablehlo(module, options)
        executor = Executor(gm=None, graph_constants=None, compiler_config=options)
        ttir = tt_mlir.compile_stable_hlo_to_ttir(
            module.operation.get_asm(enable_debug_info=True)
        )
        dump_module(module=ttir, name="TTIR module", compiler_config=options)
        binary, ttnn = tt_mlir.compile_ttir_to_bytestream(ttir)
        dump_module(module=ttnn, name="TTNN module", compiler_config=options)
        executor.set_binary(binary)
        return executor(*example_inputs)
```

### DIFF
--- a/tests/models/MobileNetV2/test_MobileNetV2.py
+++ b/tests/models/MobileNetV2/test_MobileNetV2.py
@@ -9,7 +9,7 @@ from PIL import Image
 
 import pytest
 from tests.utils import ModelTester
-from tt_torch.tools.utils import CompilerConfig, CompileDepth
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
 
 
 class ThisTester(ModelTester):
@@ -35,7 +35,11 @@ class ThisTester(ModelTester):
     "mode",
     ["eval"],
 )
-@pytest.mark.parametrize("op_by_op", [True, False], ids=["op_by_op", "full"])
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
 def test_MobileNetV2(record_property, mode, op_by_op):
     model_name = "MobileNetV2"
     cc = CompilerConfig()
@@ -43,6 +47,8 @@ def test_MobileNetV2(record_property, mode, op_by_op):
     cc.consteval_parameters = True
     if op_by_op:
         cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
 
     tester = ThisTester(
         model_name,
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -2,567 +2,34 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import torch
-
-from torch._dynamo.backends.common import aot_autograd
-from torch.fx.experimental.proxy_tensor import make_fx
-from torch._functorch.compile_utils import strip_overloads
-import operator
-
-from tt_torch.dynamo.passes import pass_pipeline
-from tt_torch.tools.utils import (
-    CompilerConfig,
-    CompileDepth,
-    Op,
-    Tensor,
-    OpCompilationStatus,
-    calculate_atol,
-    calculate_pcc,
-)
-
+import os
 import tt_mlir
-from tt_mlir import is_runtime_debug_enabled
+import sys
 import torch_mlir
-from torch_mlir.ir import Context, Location
-from torch_mlir.extras.fx_importer import FxImporter, ContextCache
 
-from torch_mlir.dialects import torch as torch_dialect
-
+from tt_torch.dynamo.torch_backend import (
+    TorchExecutor,
+    import_graph,
+    verify_ir,
+    lower_to_stable_hlo,
+)
+from tt_torch.dynamo.shlo_backend import (
+    generate_random_inputs_for_shlo,
+    parse_module_from_str,
+    StablehloExecutor,
+)
 from torch_mlir.compiler_utils import (
     OutputType,
     run_pipeline_with_repro_report,
     lower_mlir_module,
 )
-from typing import List, Tuple, Union, Optional
-import os
-import multiprocessing as mp
-import time
-import faulthandler
-from pathlib import Path
-import re
-import sys
-import tempfile
-
-
-def verify_ir(module):
-    def verify_op(op):
-        if hasattr(op, "verify"):
-            op.verify()
-        return torch_mlir.ir.WalkResult.ADVANCE
-
-    module.operation.walk(verify_op)
-
-
-class TTContextCache(ContextCache):
-    def get_node_location(self, node: torch.fx.Node) -> Optional[Location]:
-        return Location.name(node.name, context=self._c)
-
-
-def import_graph(graph: torch.fx.GraphModule):
-    context = Context()
-    torch_dialect.register_dialect(context)
-    importer = FxImporter(context=context)
-    importer._cc = TTContextCache(
-        importer._c, py_attr_tracker=importer._py_attr_tracker
-    )
-    importer.import_stateless_graph(graph)
-    return importer.module
-
-
-def lower_to_stable_hlo(module, op=None, enable_ir_printing=False):
-    run_pipeline_with_repro_report(
-        module,
-        f"builtin.module(torchdynamo-export-to-torch-backend-pipeline)",
-        "Lowering TorchFX IR -> Torch Backend IR",
-        enable_ir_printing,
-    )
-    if op is not None:
-        op.compilation_status = OpCompilationStatus.CONVERTED_TO_TORCH_BACKEND_IR
-
-    lower_mlir_module(False, OutputType.STABLEHLO, module)
-    if op is not None:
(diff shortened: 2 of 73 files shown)


tt_torch/dynamo/backend.py
```diff
@@ -575,19 +42,28 @@ def verify_golden_callback(binary, callback_context, op_context):
...
+    executor = StablehloExecutor(module=shlo, compiler_config=compiler_config)
+    if gm is not None:
+        # original input is a torch graph
+        executor.add_gm(gm, graph_constants)
+    return executor
+
+
+def _torch_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
     # Apply environment overrides at start of compilation to allow overriding what was set in the test
     compiler_config.apply_environment_overrides()
```
AleksKnezevic, line 55: This is already called in `_base_backend`, no?

tt_torch/tools/utils.py
```diff
@@ -313,7 +321,16 @@ def save_unique_ops(self):
         pytest_test = pytest_test.replace("[", "_").replace("]", "_")
         for key, op in self.unique_ops.items():
             unique_op_dict[key] = op.to_dict()
-        output_file = Path(f"{self.results_path}{pytest_test}_unique_ops.json")
+        if mode is None:
```
AleksKnezevic, line 324: mode would be "torch" here, no?

AleksKnezevic: Also, I think we can just keep the pytest name without mode, the pytest should differentiate whether it's stable_hlo or torch.

tt_torch/onnx_compile/onnx_compile.py
```diff
@@ -55,6 +55,13 @@ def compile_onnx(module: onnx.ModelProto):
         print("StableHLO module", file=sys.stderr)
         module.print(large_elements_limit=0)
 
+    return module
+
+
+def compile_onnx(module: onnx.ModelProto):
+
+    module = onnx_to_stablehlo(module)
```
AleksKnezevic, line 63: What do you think of removing the whole `gm_or_shlo` thing from backend, and having backend.py only take in torch modules which it can do op-by-op in either shlo or torch, and having the shlo path come from here? then we would still have a clear divide between the pytorch and onnx compile flows?

AleksKnezevic (changes requested):

tt_torch/dynamo/backend.py
ddilbazTT: Yes, but `_torch_backend` and `_base_backend` are separate.

tt_torch/tools/utils.py
ddilbazTT: I see, so we should parameterize the `op_by_op_backend` in the test itself, right?

tt_torch/onnx_compile/onnx_compile.py
ddilbazTT: We could. I want to confirm the following would change to make this happen:
`backend`:
- assert that input is a torch graph
- Another option could be to accept both torch and onnx graphs as inputs
`_base_backend`:
- lower to stablehlo then execute, or execute torch graph as is, depending on op_by_op_backend
`compile_onnx`:
- lowers to stablehlo, then calls `_stablehlo_backend`

AleksKnezevic: Yeah: 

- `backend` remains more-or-less unchanged, jsut calls `_base_backend`
- `_base_backend` either lowers to shlo then executes, or executes depending on `op_by_op_backend`
- `compile_onnx` lowers to shlo, then calls `_stablehlo_backend`

AleksKnezevic: There'll probably be some additional refactoring/cleanup once you mode all that code.

tt_torch/dynamo/backend.py
```diff
@@ -624,12 +98,20 @@ def _base_backend(gm: torch.fx.GraphModule, example_inputs, compiler_config):
         print("StableHLO module", file=sys.stderr)
         module.dump()
 
+    return module, gm, graph_constants
+
+
+def shlo_to_flatbuffer(executor, module, compiler_config):
+    dump_intermediates = os.environ.get("TT_TORCH_IR_LOG_LEVEL")
```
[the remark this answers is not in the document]
ddilbazTT: Can I also add dump_intermediates, dump_info and dump_debug to compiler_config? So that we don't have to set dump_info every time, for exmaple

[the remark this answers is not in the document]
AleksKnezevic: Do we use dump_debug anywhere? To pass into the pipelines I guess?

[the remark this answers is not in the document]
AleksKnezevic: but yeah, this can be hidden in compiler_cfg

tt_torch/dynamo/executor.py
```diff
@@ -0,0 +1,342 @@
...
+            new_inputs = new_inputs + ((input),)
+        return new_inputs
+
+    def set_binary(self, binary):
+        self.binary = binary
+
+    def __call__(self, *inputs):
+        if self.compiler_config.compile_depth != CompileDepth.EXECUTE:
+            assert (
+                self.compiler_config.op_by_op_backend == OpByOpBackend.TORCH
```
AleksKnezevic, line 127: is `OpByOpBackend` a valid config for compiledepths that aren't op by op?

tests/models/MobileNetV2/test_MobileNetV2.py
```diff
@@ -36,13 +36,16 @@ def _load_inputs(self):
     ["eval"],
 )
 @pytest.mark.parametrize("op_by_op", [True, False], ids=["op_by_op", "full"])
-def test_MobileNetV2(record_property, mode, op_by_op):
+@pytest.mark.parametrize("stablehlo", [True, False], ids=["stablehlo", "default"])
```
AleksKnezevic, line 39: This will change the name of all the tests, do tests need to be updated with the new names?

ddilbazTT: Yes, I just realized that and made the changes in the workflows. I added `default` to wherever needed, hope I didn't miss any

ddilbazTT: From off-line convo, decided to convert this to: `@pytest.mark.parametrize("op_by_op", [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None], ids=["op_by_op_stablehlo", "op_by_op_torch", "full"])`

AleksKnezevic (approved):

AleksKnezevic:
@LPanosTT, can you please take a look at this PR as well.

LPanosTT (approved):
LGTM

ddilbazTT:
@mrakitaTT @kmitrovicTT FYI

Issue #401, opened 2025-03-06
Unify FE Reporting


jameszianxuTT:
Use the standard format for reporting full model execution test results, to align with other FEs. 

This involves changes to how the junitxml is reported, requiring some name remapping and additional data exported through tags.

PR #402, opened 2025-03-06
FE unification - Amend XML reporting to align with other FEs

jameszianxuTT:
### Ticket
#401 

### Problem description
Frontends adopted a unified reporting format, tt-torch does not comply.

### What's changed
Add and remap extant fields to match FE standardized format. 

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -31,6 +31,7 @@ class ModelTester:
         assert_pcc=True,
         assert_atol=True,
         record_property_handle=None,
+        model_group="generality",
     ):
         if mode not in ["train", "eval"]:
             raise ValueError(f"Current mode is not supported: {mode}")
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -293,6 +296,22 @@ def record_aggregate_model_metric(self, metric_key, default_value=-1):
         self.record_tag_cache["min_" + metric_key] = min_metric
         self.record_tag_cache["max_" + metric_key] = max_metric
 
+    def remap_compile_depth(self, compile_depth, min_pcc):
+        compile_depth_translation_table = {
+            CompileDepth.TORCH_FX: "FAILED_TTMLIR_COMPILATION",
+            CompileDepth.TORCH_MLIR: "FAILED_RUNTIME",
+            CompileDepth.STABLEHLO: "FAILED_RUNTIME",
+            CompileDepth.TTNN_IR: "FAILED_RUNTIME",
+            CompileDepth.COMPILE_OP_BY_OP: "INCORRECT_RESULT",
```
jameszianxuTT, line 305: There isn't a standard enum value for op-by-op tests so this is arbitrary. 
Existing compile depth in the config will remain as is.

This is only used for reporting E2E model tests and CompileDepth.EXECUTE is the only relevant field in that case.

AleksKnezevic (approved):

Issue #412, opened 2025-03-10
Op by Op test_case_names have become excessively verbose


jameszianxuTT:
Some change on/around March 6th has caused op by op test reporting (likely in data ingestion) to break. 

aten::cat used to be reported as `aten::cat_[torch.Size([1, 13, 64]), torch.Size([1, 13, 64])]_-1`

[...]

PR #414, opened 2025-03-10
Properly unpack inputs to get_input_shapes_and_constants

jameszianxuTT:
### Ticket
#412 

### Problem description
Inputs are implicitly wrapped in a tuple when not using `*`
to pass them into get_input_shapes_and_constants.

### What's changed
Add `*` to unpack inputs as individual args, not as a single
element so the outer `for inp in inputs` loop runs multiple times.

### Checklist

### DIFF
--- a/tt_torch/dynamo/torch_backend.py
+++ b/tt_torch/dynamo/torch_backend.py
@@ -116,7 +116,7 @@ class TorchExecutor(OpByOpExecutor):
 
     def get_stable_hlo_graph(self, node, inputs, **kwargs):
 
-        input_shapes_and_constants = self.get_input_shapes_and_constants(inputs)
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
I'm going to rerun a bunch of the op by op tests to see if this actually does fix everything; it looks to work locally for openpose. See discussion in the linked issue #412

jameszianxuTT:
I don't see any comparable logic in the shlo_backend, but you may want to check that there are no other refactored-out-to-methods that break in the same way in the [shlo_compiler changeset](https://github.com/tenstorrent/tt-torch/pull/394/files#diff-65fb0e547d4443ea220ddd7ea493a10549292fed30d9cd2bdf56288c0e34583e). This does seem unlikely since everything else seems to work. @ddilbazTT

ddilbazTT (approved):
Good catch!

PR #415, opened 2025-03-10
Runtime stitching bringup: Added granular runtime APIs e2e, added distilbert multi iteration test.

### LINKED ISSUE #400: Granularize tt_mlir.run into multiple APIs
## Problem Statement
Currently in tt_mlir.run, we
* Do some input preprocessing
* Open the device
* Create runtime tensors
* Submit the workload
* Copy out the tensors
* Close the device

This is problematic because:
* We open and close device on every separate run, which can impact performance and no device state can be preserved between runs
  * We also don't have a way for the user to configure device settings (like async mode)
* All input preprocessing alloc/dealloc is performed within the API - users do not have a way to access these input handles to use for later runs (which is required for runtime stitching)
* User has limited control due to some parameters being hardcoded, like device ids, program ids etc.

## Proposal

I propose we granularize `tt_mlir.run` by splitting it into multiple APIs such that the user has more control over each step and can preserve handles to intermediates as needed.

---
```
tt::runtime::Binary create_binary_from_bytestream(py::bytes byte_stream);
```

Create a binary from byte stream and return the binary handle.

---
```
tt::runtime::Device open_device(vector<int> device_ids, bool enable_async_mode, int l1_small_size etc.);
```

This will just be a wrapper around runtime openDevice, and will return a device handle to the user.

---
```
std::vector<tt::runtime::Tensor> preprocess_inputs(std::vector<at::Tensor> &inputs, tt::runtime::Binary binary, int program_idx);
```

Preprocesses the torch tensors based on the layout described in the binary and returns the runtime tensors. The binary can be generated from the bytesteam via the new `create_binary_from_bytestream` API. This is a core API for runtime stitching such that we can pre-convert tensors to their desired layouts and keep them there.

---
```
std::vector<at::Tensor> run(tt::runtime::Device device, std::vector<tt::runtime::Tensor> preprocessed_inputs, tt::runtime::Binary binary, int program_index);
```
This represents the middle portion of the current tt_mlir.run API. It'll now take the preprocessed_inputs and run them on device. This may need modification in the future to:
* Return the outputs as runtime tensors instead of copying them to torch tensors directly, but for now I'm assuming we don't need the outputs to persist on device to feed a consumer program (though this may be the case for LLMs to accumulate the tokens so please correct me if I'm wrong).
* Accept non processed inputs (activations) alongside pre-processed inputs (weights). Since runtime currently moves everything to device initially regardless, we probably don't need this for now.
---
```
void close_device(tt::runtime::Device device);
```

Closes the runtime device. Should just be a wrapper around `tt::runtime::CloseDevice`. With this the user has control over when to open/close the devices.

jnie-TT:
### Ticket
Closes #400 

### Problem description
As described in #400, we needed more granular APIs to be able to cache tensors on device between runs. 

### What's changed
Broke down the previous `tt_mlir.run` API to multiple APIs. Updated the base `Executor` class to use the granular flow, as well as various options in `CompilerConfig` to configure it. Added a `tt_mlir.run_end_to_end` API that functions the same way as the previous `tt_mlir.run`, which isolates all runtime details in one place. 

Added a multi-iteration distilbert test that runs on pre-cached inputs.

Tested perf on llama-3b, got a 1.5x perf improvement over 128 iterations, however consteval has some issues on llama-3b so adding distilbert as a sanity test instead.

### Checklist

### DIFF
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -39,6 +39,7 @@ jobs:
                   tests/models/openpose/test_openpose_v2.py::test_openpose_v2[full-eval]
                   tests/models/resnet/test_resnet.py::test_resnet[full-eval]
                   tests/models/resnet50/test_resnet50.py::test_resnet[full-eval]
+                  tests/models/distilbert/test_distilbert.py::test_distilbert_multiloop[full-distilbert-base-uncased-eval-64]
             "
           },
           {
--- a/tests/models/distilbert/test_distilbert.py
+++ b/tests/models/distilbert/test_distilbert.py
@@ -56,3 +56,50 @@ def test_distilbert(record_property, model_name, mode, op_by_op):
         print(f"Model: {model_name} | Input: {tester.text} | Output: {results}")
 
     tester.finalize()
+
+
+@pytest.mark.parametrize(
+    "num_loops",
+    [64],
+)
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize("model_name", ["distilbert-base-uncased"])
+@pytest.mark.parametrize(
+    "op_by_op",
+    [None],
+    ids=["full"],
+)
+def test_distilbert_multiloop(record_property, model_name, mode, op_by_op, num_loops):
+    import time
+
+    cc = CompilerConfig()
+    cc.enable_consteval = True
+    cc.consteval_parameters = True
+    cc.enable_async = True
+    cc.cache_preprocessed_constants = True
+    cc.initialize_device()
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        assert_pcc=False,
+        assert_atol=False,
+        compiler_config=cc,
+        record_property_handle=record_property,
+    )
+    model = tester.compile_model(tester.get_framework_model(), tester.compiler_config)
+
+    with torch.no_grad():
+        start_time = time.time()
+        for _ in range(num_loops):
+            results = tester.run_model(model, tester.inputs)
+        end_time = time.time()
+
+        print(f"Model: {model_name} | Input: {tester.text} | Output: {results}")
+        print(f"{num_loops} iterations took {(end_time - start_time)} seconds")
+
+    tester.finalize()
+    cc.cleanup_device()
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -238,21 +238,25 @@ class ModelTester:
         self.record_tag_cache["pccs"] = pccs
         self.record_tag_cache["atols"] = atols
 
-    @torch.no_grad()
-    def test_model_eval(self, on_device=True):
+    def get_framework_model(self):
         model = (
             self.framework_model.eval()
             if hasattr(self.framework_model, "eval")
             else self.framework_model
         )
+        return model
+
+    @torch.no_grad()
+    def test_model_eval(self, on_device=True):
+        model = self.get_framework_model()
         golden = self.get_golden_outputs(model, self.inputs)
 
         if on_device == True:
             model = self.compile_model(model, self.compiler_config)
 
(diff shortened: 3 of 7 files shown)


jameszianxuTT (changes requested):

tests/models/distilbert/test_distilbert.py
```diff
@@ -56,3 +56,46 @@ def test_distilbert(record_property, model_name, mode, op_by_op):
         print(f"Model: {model_name} | Input: {tester.text} | Output: {results}")
 
     tester.finalize()
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize("model_name", ["distilbert-base-uncased"])
```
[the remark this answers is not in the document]
jameszianxuTT: Suggest changing this to `distilbert_base_uncased`

.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -39,6 +39,7 @@ jobs:
                   tests/models/openpose/test_openpose_v2.py::test_openpose_v2[full-eval]
                   tests/models/resnet/test_resnet.py::test_resnet[full-eval]
                   tests/models/resnet50/test_resnet50.py::test_resnet[full-eval]
+                  tests/models/distilbert/test_distilbert.py::test_distilbert_multiloop
```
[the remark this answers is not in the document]
jameszianxuTT: Feel free to ignore this nit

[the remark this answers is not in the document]
jnie-TT: Fixed

tests/models/distilbert/test_distilbert.py
```diff
@@ -56,3 +56,51 @@ def test_distilbert(record_property, model_name, mode, op_by_op):
...
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        assert_pcc=False,
+        assert_atol=False,
+        compiler_config=cc,
+        record_property_handle=record_property,
+    )
+    model, _ = tester.get_model_and_golden()
```
AleksKnezevic, line 93: why not just `model = tester.framework_model`

tests/utils.py
```diff
@@ -238,21 +238,25 @@ def verify_outputs(self, golden, outputs):
         self.record_tag_cache["pccs"] = pccs
         self.record_tag_cache["atols"] = atols
 
-    @torch.no_grad()
-    def test_model_eval(self, on_device=True):
+    def get_model_and_golden(self):
```
AleksKnezevic, line 241: if you're going to keep this API, see comment above, I would call this `get_framework_model` and have a separate `get_golden`

tt_torch/csrc/bindings.cpp
```diff
@@ -124,23 +124,60 @@ static torch::Tensor create_torch_tensor(const tt::runtime::Tensor &tensor,
...
+}
+
+std::string bytestream_to_json(py::bytes byte_stream) {
   std::string data_str = byte_stream;
   auto binary_ptr = std::shared_ptr<void>(
       new char[data_str.size()],
       [](void *ptr) { delete[] static_cast<char *>(ptr); } // Custom deleter
   );
   // Copy data into the allocated memory
+  std::memset(binary_ptr.get(), 0, data_str.size());
```
AleksKnezevic, line 162: nit, this should be hidden behind a debug flag.

tests/models/distilbert/test_distilbert.py
jnie-TT: Makes sense! Fixed

tt_torch/csrc/bindings.cpp
jnie-TT: Removed this entirely since it's not needed

AleksKnezevic (approved):

jameszianxuTT (approved):

PR #416, opened 2025-03-10
Use Torch ExportedModule to import initial MLIR module

LPanosTT:
Use a `torch.export.ExportedProgram` to generate the initial MLIR module.

This requires us to create an `ExportedProgram` from the initial `GraphModule`.

Benefits:
- We can use the torch-mlir's official entrypoint
    - This handles in-place ops for us
- We can run decompositions and **keep** location data
    - This location data will stick around throughout the compile process

Issues:
- `aten.clamp` is decomposed by torch-mlir to `maximum(minimum(input, max), min)`. `ttnn.maximum` requires that the operand which needs to be broadcasted is on the RHS. Currently, in tt-mlir the `PartiallyBroadcastable` op trait only enforces that the broadcasted operand is on the LHS
    - tt-torch issue: https://github.com/tenstorrent/tt-torch/issues/431
    - tt-mlir issue: https://github.com/tenstorrent/tt-mlir/issues/2458
- Graph parameters are inlined as constants in the graph. To have the `FxImporter` treat them as graph inputs we need to edit the `ExportedModule`s `ExportedGraphSignature` and force all "parameter" types to "user inputs"
    - This is a hack as the `ExportedGraphSignature` is meant to be a private member of `ExportedProgram`
    - Ideally we can configure the `FxImporter` to _not_ inline the parameters if we pass a flag of some sort. Perhaps a future contribution to torch-mlir.

Other Info:
- We need to upgrade to PyTorch 2.6.0 as it contains crucial changes which allow us to use custom decompositions (necessary to support interpolation)
    - AdaptiveAvgPool2d is lowered AvgPool2d and eventually to `stablehlo.reduce_window **even in the case where the op is equivalent to a global average**. Since we do not have support for lowering a sum_pool in `StablehloToTTIRPatterns.cpp` (sum because the division is afterward), I've temporarily added a custom decomposition of `aten.avg_pool2d` which will convert to a mean over the spatial dimensions when the `avg_pool2d` is equivalent to it.
    - `aten.split` is no longer lowered to a series of `narrow` ops. Instead it is now lowered to a series of `as_strided` ops.
        - `narrow` is lowered to `slice`, which can be lowered to `stablehlo.slice`. `as_strided` cannot be lowered from Torch Backend IR to Stablehlo. I've temporarily added back the old decomposition from PyTorch 2.5.0 which uses narrow as a custom decomposition.
        - I've made a PR which adds a lowering of `AtenAsStridedOp` to `stablehlo::SliceOp` in our fork of torch-mlir: https://github.com/tenstorrent/llvm-torch-mlir/pull/4
     - The tracer which generates the `GraphModule` which is passed to `backend` does not account for control flow, I believe in PyTorch 2.5.0 a graph break would be triggered during `.generate` methods in `transformers` LLMs. It does not anymore and so `.generate` will run until the max length is reached.
         - **this means that the entire generation becomes  one program**
         - Once the first EOS token is generated, the rest of the length is filled with padding. We cannot compare the golden output to the result from the `GraphModule` as the output shapes are different.
             - Since the output of `.generate` graphs are integers PCC/atol verification is not quite useful but does return `True` when the outputs are _identical_
             - The tokenizer can decode the outputs and strip padding.
             - I've added a flag to `ModelTester` that informs the `ModelTester` it is testing a `.generate` call. It will decode the output tokens and we compare the resulting strings.
         - PyTorch has an experimental `torch.cond` which they seem to intend to use to trace data-dependent control-flow. There's a note in the `transformers` source that says they intend to use it when it is no longer experimental
- When the graph is compiled, the user inputs are placed **at the end** of the arguments passed to the program rather than the front. That is graph constants first, then inputs.
- I needed to implement an `FxImporter` hook for importing literals to the graph. By default it will make all non-scalars `DenseElementsResourceAttr`s, however, this causes the process to hang upon cleanup whether the test fails or not. So the hook just uses `DenseElementsAttr` for all literals.
    - Someone has mentioned this problem in an IREE issue as well: https://github.com/iree-org/iree/issues/20102
    - They've traced it down to this PR in llvm that adds a GIL acquire when destroying the `DenseElementsResourceAttr`: https://github.com/llvm/llvm-project/pull/124832

### DIFF
--- a/tests/models/mamba/test_mamba.py
+++ b/tests/models/mamba/test_mamba.py
@@ -69,7 +69,11 @@ def test_mamba(record_property, model_name, mode, op_by_op):
             cc.op_by_op_backend = OpByOpBackend.STABLEHLO
 
     tester = ThisTester(
-        model_name, mode, compiler_config=cc, record_property_handle=record_property
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        is_token_output=True,
     )
     results = tester.test_model()
 
--- a/tests/torch/test_basic.py
+++ b/tests/torch/test_basic.py
@@ -198,21 +198,6 @@ def test_linear_with_bias():
     verify_module(Basic(), input_shapes=[(32, 32)])
 
 
-def test_linear_with_bias_no_embedded_constants():
-    class Basic(nn.Module):
-        def __init__(self):
-            super().__init__()
-            self.linear_a = nn.Linear(32, 32)
-
-        def forward(self, x):
-            x = self.linear_a(x)
-            return x
-
-    cc = CompilerConfig()
-    cc.remove_embedded_constants = True
-    verify_module(Basic(), input_shapes=[(32, 32)], compiler_config=cc)
-
-
 @pytest.mark.parametrize(
     ("input_type"),
     [
--- a/tests/torch/test_constant_fold.py
+++ b/tests/torch/test_constant_fold.py
@@ -50,7 +50,6 @@ def test_interp():
     inH = 5
     inW = 5
     inC = 1
-    scale_factor = 3
 
     input_shape = (1, inC, inH, inW)
     small = (
--- a/tt_torch/dynamo/decompositions.py
+++ b/tt_torch/dynamo/decompositions.py
@@ -37,7 +37,7 @@ def _current(scope: str) -> DecompositionTable:
     if stack:
         return dict(stack[-1])
     else:
-        return dict(DEFAULT_DECOMPOSITION_TABLE)
+        return dict(CUSTOM_DECOMPOSITION_TABLE)
 
 
 @contextlib.contextmanager
@@ -46,7 +46,7 @@ def _extend_context_manager(
     *,
     from_current: bool = True,
     add_ops: Optional[DecompositionOpsList] = None,
-    remove_ops: Optional[DecompositionOpsList] = None
+    remove_ops: Optional[DecompositionOpsList] = None,
 ):
     table: DecompositionTable
     if from_current:
@@ -76,7 +76,9 @@ def _extend_context_manager(
 # torch's upsample_bilinear2d when align_corners=True.
 # This logic was derived from @brentyi's implementation in:
 #    https://github.com/jax-ml/jax/issues/11206#issuecomment-1423140760
-def compute_bilinear_weight(input_size, output_size, scale, align_corners, dtype):
+def compute_linear_weight(input_size, output_size, scale, align_corners, dtype):
+    if input_size == 1:
+        return torch.ones(1, output_size, dtype=dtype)
     translation = 0
     if align_corners:
         scale = (output_size - 1) / (input_size - 1)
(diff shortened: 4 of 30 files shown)


jameszianxuTT (approved):

tests/torch/test_basic.py
```diff
@@ -208,9 +208,7 @@ def forward(self, x):
             x = self.linear_a(x)
             return x
 
-    cc = CompilerConfig()
-    cc.remove_embedded_constants = True
-    verify_module(Basic(), input_shapes=[(32, 32)], compiler_config=cc)
+    verify_module(Basic(), input_shapes=[(32, 32)])
```
AleksKnezevic, line 211: This test is redundant now and should be removed.

tests/torch/test_constant_fold.py
```diff
@@ -38,6 +38,10 @@ def forward(self, x):
 
 
 def test_interp():
+    pytest.skip(
```
AleksKnezevic, line 41: remove

tt_torch/dynamo/executor.py
```diff
@@ -175,12 +179,14 @@ def __call__(self, *inputs):
...
-            preprocessed_inputs += self.preprocessed_graph_constants
+            preprocessed_inputs = (
+                self.preprocessed_graph_constants + preprocessed_inputs
+            )
 
         outputs = tt_mlir.run(device, binary, program_idx, preprocessed_inputs)
 
-        self._cache_constants_if_needed(preprocessed_inputs[activations_len:])
-        self._cleanup_resources(preprocessed_inputs[:activations_len], device)
+        self._cache_constants_if_needed(preprocessed_inputs[:-activations_len])
```
ddilbazTT, line 188: It seems we have to call `self._cache_constants_if_needed` to set `self.preprocessed_graph_constants` but we check if `self.preprocessed_graph_constants` is None before [self._cache_constants_if_needed](url) is called - is that how we want it to work? I don't know who pushed in that change but we're continuing it here, so wanted to ask. `self.preprocessed_graph_constants` is always going to be `None` in the previous checks, no?

tt_torch/dynamo/torch_backend.py
```diff
@@ -45,18 +56,84 @@ def verify_op(op):
...
 def import_graph(graph: torch.fx.GraphModule):
-    context = Context()
-    torch_dialect.register_dialect(context)
-    importer = FxImporter(context=context)
-    importer._cc = TTContextCache(
-        importer._c, py_attr_tracker=importer._py_attr_tracker
-    )
-    importer.import_stateless_graph(graph)
-    return importer.module
+    with Context() as context:
```
ddilbazTT, line 79: do we use this module anywhere after changes? Seems removed. If removed, we should also remove it from imports.

LPanosTT: This method of importing the graph is still used in the torch op-by-op flow to generate modules from the single op graphs.

tt_torch/dynamo/executor.py
jnie-TT: @ddilbazTT The idea is that the executor could execute the graph multiple times in a loop. In the first loop `self.preprocessed_graph_constants` will be None, therefore we will cache the constants. Then in subsequent loops, since we cached constants in the first loop, `self.preprocessed_graph_constants` will not be None anymore and we can use them directly instead of preprocessing them again, thus saving performance cycles

ddilbazTT: I think eventually we need to propagate this to all executors. In the base executor, we're running the model end to end but when we're running op-by-op, we would probably need to do something similar. That's beyond the scope of this PR though, thanks for explaining! @jnie-TT

tests/models/mamba/test_mamba.py
```diff
@@ -69,7 +69,11 @@ def test_mamba(record_property, model_name, mode, op_by_op):
             cc.op_by_op_backend = OpByOpBackend.STABLEHLO
 
     tester = ThisTester(
-        model_name, mode, compiler_config=cc, record_property_handle=record_property
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        is_transformers_generation=True,
```
AleksKnezevic, line 76: Maybe `is_string_output` is a better name, but that's pretty nit-picky.

tests/models/resnet50/test_resnet50.py
```diff
@@ -30,7 +30,7 @@ def _load_inputs(self):
         img_t = preprocess(image)
         batch_t = torch.unsqueeze(img_t, 0)
         batch_t = batch_t.to(torch.bfloat16)
-        return batch_t
+        return (batch_t,)
```
AleksKnezevic, line 33: Why is this needed? Do all models need to return a tuple?

tests/torch/test_constant_fold.py
```diff
@@ -50,7 +50,6 @@ def forward(self, x):
     inH = 5
     inW = 5
     inC = 1
-    scale_factor = 3
```
AleksKnezevic, line 53: Was this intended?

LPanosTT: Yes this variable isn't used.

AleksKnezevic (approved):

tt_torch/dynamo/decompositions.py
```diff
@@ -76,7 +76,9 @@ def _extend_context_manager(
 # torch's upsample_bilinear2d when align_corners=True.
 # This logic was derived from @brentyi's implementation in:
 #    https://github.com/jax-ml/jax/issues/11206#issuecomment-1423140760
```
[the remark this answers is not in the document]
LPanosTT: Yes but I just added a new special case, the normal implementation is inspired by that source

tests/models/mamba/test_mamba.py
LPanosTT: Yeah, I'm thinking `is_token_output` might be better since it's an integer tensor. Also I'll verify that `self.tokenizer` was added if this is the case.

tests/models/resnet50/test_resnet50.py
LPanosTT: It's not needed anymore. Remanent from earlier commits when it was necessary

Issue #291, opened 2025-02-07
Deepseek qwen test fails in op-by-op


AleksKnezevic:
Please debug and re-enable.

ddilbazTT:
deepseek qwen passes in commit hash 2256f17
Steps to reproduce:
```
git checkout -b deepseek_qwen_2256f17 2256f17
deactivate; rm -rf env/venv build install third_party/tt-mlir/
source env/activate
cmake -G Ninja -B build
cmake --build build
cmake --install build
pytest -s tests/models/deepseek/test_deepseek_qwen.py --op_by_op | tee 2256f17_deepseek_qwen.log

```
Branch: `deepseek_qwen_2256f17`

ddilbazTT:
Debugged further. The error is happening when compiling node.idx == 9552. The following is some of the debug info:

```
(Pdb) p op.stable_hlo_graph
'module {\n  func.func @main(%arg0: tensor<10x5120xbf16>, %arg1: tensor<5120x152064xbf16>) -> tensor<10x152064xbf16> {\n    %0 = stablehlo.dot_general %arg0, %arg1, contracting_dims = [1
] x [0] : (tensor<10x5120xbf16>, tensor<5120x152064xbf16>) -> tensor<10x152064xbf16>\n    return %0 : tensor<10x152064xbf16>\n  }\n}\n'
(Pdb) p op.torch_ir_graph
'module {\n  func.func @main(%arg0: !torch.vtensor<[10,5120],bf16>, %arg1: !torch.vtensor<[5120,152064],bf16>) -> !torch.vtensor<[10,152064],bf16> {\n    %0 = torch.aten.mm %arg0, %arg1
: !torch.vtensor<[10,5120],bf16>, !torch.vtensor<[5120,152064],bf16> -> !torch.vtensor<[10,152064],bf16>\n    return %0 : !torch.vtensor<[10,152064],bf16>\n  }\n}\n'

```

The error happens during
`calculated, runtime_stack_dump = self.run_op(binary, *args)`

AleksKnezevic:
you can single step through run_op to see where it happens. What I've done previously, to speed up debug, is conditionally compile/execute only the op I was interested in. than you can repro in a few seconds. 

My guess is that it's in the json generation

ddilbazTT:
The error doesn't happens after compilation, during execution. So I don't think it's json generation.

AleksKnezevic:
right, makes sense.

PR #440, opened 2025-03-15
Fix Deepseek Qwen Fatal Bus Error [#291]

ddilbazTT:
### Ticket
#291 

### Problem description
Deepseek Qwen op no. 9552 fails with out-of-memory error. This is expected to be caught by try-except block. However, it is not. Instead, the whole program crashes.

The error is related to multiprocessing in Python. There is no full isolation between parent and child processes. So, the child process running out of memory kills the parent process. I deduced this by running several tests:

- a single threaded `run_op` function - successfully passes
- a multi processed run_op function which only calls `tt_mlir.run_end_to_end`, does not do anything related to stderr piping etc - fails

Further research showed that the parent and child processes share disk space when using multiprocessing.

### What's changed
Replaced a multi-processed `run_op` with one that runs a subprocess, since subprocesses provide complete isolation.

### Some References:
[https://news.ycombinator.com/item?id=15009816#:~:text=Memory%20consumption%20is%20bloated%20and,used%20Python%20subprocesses%20that%20way.](https://news.ycombinator.com/item?id=15009816#:~:text=Memory%20consumption%20is%20bloated%20and,used%20Python%20subprocesses%20that%20way.)
[https://www.infoworld.com/article/2257425/python-threading-and-subprocesses-explained.html](https://www.infoworld.com/article/2257425/python-threading-and-subprocesses-explained.html)

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -3,6 +3,9 @@
 name: Nightly Tests
 
 on:
+  push:
+    branches:
+      - deepseek_qwen
   schedule:
     - cron: '0 4 * * *'  # Runs at 04:00 UTC every day
   workflow_dispatch:  # Manual trigger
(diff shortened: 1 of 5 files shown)


ddilbazTT:
I ran this test as a nightly to see if it would pass: [https://github.com/tenstorrent/tt-torch/actions/runs/13865104371](https://github.com/tenstorrent/tt-torch/actions/runs/13865104371)

jazpurTT:
@ddilbazTT @AleksKnezevic similar issue that I am having with the mistral model, I ran it with these changes and it solves the Bus Error issue, just seeing pcc/atol failure now.

AleksKnezevic:
Thanks @jazpurTT. @ddilbazTT has a different solution we think will also solve the problem.

Issue #261, opened 2025-02-03
enable object detection tests


AleksKnezevic:
All tests in `tests/models/torchvision/test_torchvision_object_detection.py` fail with torch compile issues. Fix and enable.

AleksKnezevic:
All the tests in that file  should be added to `.github/workflows/run-op-by-op-model-tests.yml`. You can split them up such that no test runs longer than 2-3 hours. 

Please note, when testing this change outside of a nightly pipeline on main, disable (comment out) all other tests otherwise CI runners get swamped.

jazpurTT:
Compilation failure happens at the "Torch Backend IR -> StableHLO Backend IR" pass. Looks like stableHLO expects a 1x256x37x37 output instead of the 1x256x38x38 derived from the Torch Backend IR.
The correct shape is 1x256x38x38. This is due to pytorch's ceil_mode param to use _ceil_ instead of _floor_ to compute the output shape. But, stableHLO does not have an equivalent parameter to use _ceil_ for the output shape calculation.

`error: 'stablehlo.reduce_window' op inferred type(s) 'tensor<1x256x37x37xf32>' are incompatible with return type(s) of operation 'tensor<1x256x38x38xf32>'`

In the compilation pass torch-mlir only checks if the ceil mode is a constant boolean, but the variable is not used/considered. A solution could be to check if ceil_mode is True and pad an extra [+1, +1] to get the right output shape when passed down to stableHLO. But, adding the extra padding could cause the output to mismatch.

jazpurTT:
The padding should actually be the size of the stride on each dimension. So for a each dim `i `the padding should be `stride[i] / 2 `on each end to ensure the padding is symmetrical which seems to be a requirement. But, if the stride is odd, if we either ceil or floor `stride[i] / 2`it can cause the same issue in some cases.

Another solution would be to insert a padding op before and maybe we can pad the tensor by `stride[i]` without it needing to be symmetrical. But I am not sure if we would want to add this to the _torch ir -> shlo_ pass.

AleksKnezevic:
Sorry about this @jazpurTT, I pasted the wrong command above, it should be `pytest -svv tests/models/torchvision/test_torchvision_object_detection.py::test_torchvision_object_detection[op_by_op_torch-ssd300_vgg16-eval]`. It's not the actual failure of the shlo conversion that we want to debug, it's why pytorch fails in issuing us this graph. The expected failure is:

```bash
[...]

PR #445, opened 2025-03-17
t5 models golden and output shape mismatch

jazpurTT:
### Ticket
#261 
### Problem description
Output of t5 models has a tail of 0s in the output that the golden does not have, this "trimming" of the golden seems to come from some post-processing that is not part of the graph. For a golden of size [1, N] all N first values from the calculated output matches with the golden.

### What's changed
Added a flag to the tester to skip validating the generated output with the golden for cases like these.

### Checklist
- [x] Tested changes with all t5 model: https://github.com/tenstorrent/tt-torch/actions/runs/13901004579

### DIFF
--- a/tests/models/flan_t5/test_flan_t5.py
+++ b/tests/models/flan_t5/test_flan_t5.py
@@ -55,6 +55,7 @@ def test_flan_t5(record_property, mode, op_by_op):
         record_property_handle=record_property,
         assert_pcc=False,
         assert_atol=False,
+        verify_with_golden=False,
     )
     results = tester.test_model()
     if mode == "eval":
--- a/tests/models/t5/test_t5.py
+++ b/tests/models/t5/test_t5.py
@@ -52,6 +52,7 @@ def test_t5(record_property, model_name, mode, op_by_op):
         record_property_handle=record_property,
         assert_pcc=False,
         assert_atol=False,
+        verify_with_golden=False,
     )
     results = tester.test_model()
     if mode == "eval":
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -30,6 +30,7 @@ class ModelTester:
         compiler_config=None,
         assert_pcc=True,
         assert_atol=True,
+        verify_with_golden=True,
         record_property_handle=None,
         model_group="generality",
     ):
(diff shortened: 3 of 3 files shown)


tests/models/flan_t5/test_flan_t5.py
```diff
@@ -55,6 +55,7 @@ def test_flan_t5(record_property, mode, op_by_op):
         record_property_handle=record_property,
         assert_pcc=False,
         assert_atol=False,
+        verify_with_golden=False,
```
AleksKnezevic, line 58: nit: perhaps `do_verify` instead of `verify_with_golden`

AleksKnezevic (approved):
I left a couple of nit-picky comments, feel free to merge after addressing.

LPanosTT:
@AleksKnezevic @jazpurTT In my exported model branch I have a fix for this too... I thought my upgrade to PyTorch 2.6.0 had caused the problem. Long story short I use the tokenizer to decode the golden and calculated outputs and compare the string instead.

AleksKnezevic:
@LPanosTT, interesting, so even without the changes we're generating the full text? @jazpurTT, please see @LPanosTT's change, better to actually compare if possible than not...

LPanosTT:
> @LPanosTT, interesting, so even without the changes we're generating the full text? @jazpurTT, please see @LPanosTT's change, better to actually compare if possible than not...

The same text is still generated (unless there is a separate issue with the specific model of course). The compiled model just pads to `max_seq_len` and golden does not. You can ask the tokenizer to leave out special characters (like `<pad>`) while decoding and so if the actual generated tokens are the same then so will this output string.

jazpurTT:
> @LPanosTT, interesting, so even without the changes we're generating the full text? @jazpurTT, please see @LPanosTT's change, better to actually compare if possible than not...

Just checked, yeah that should work, and I agree, comparing would be better. Only suggestion I would make is to move the tokenized comparison inside the verify_outputs( .. ) function given that is what we are doing. @AleksKnezevic should I close this PR?

AleksKnezevic:
@LPanosTT, will your changes be in today? If not, @jazpurTT, I'd like to get nightly tests passing again, so you can either pull in @LPanosTT's changes here, or merge this PR and he can add them when he gets his branch merged.

LPanosTT:
@AleksKnezevic I hope so but I'm just waiting for you and/or some other to take a look again because it only has one approval and things have changed a decent amount since that approval was given in the first place.

AleksKnezevic:
@LPanosTT, well played with the uno reverse! I'll review now.

Issue #443, opened 2025-03-17
Model Benchmarking Workflow


jameszianxuTT:
- Create a workflow to attempt running all models in **CompileDepth.EXECUTE** and record their actual achieved compile depth, while gracefully exiting on failure. 
- Make sure the XML record is accurate and generated, regardless of the failure mode of the actual test. 
- Create a per-model compile depth reference JSON, and a job to compare the measured compile depths and reference compile depths
- Create a way to get the model list, perhaps by consuming all other tester YMLs, and finding a way to run them in execute (string replacement?)

jameszianxuTT:
There are too many ways that something can fail, especially during the EXECUTE phase. This could cauase a variety of signals to terminate the python process, including some that can't really be caught like SIGSEGV / SIGKILL.
- Post-crash XML report flushing does not appear too plausible - that XML may be malformed or incompletely generated in multiple ways that would be tricky to parse. Hijacking the junitxml plugin seems like it could cause unintended side effects
- As an alternative, maintain a separate, frequently flushed, incremental checkpoint log file that's parsed at the end of the tests (say, in a separate process to the pytest runner) to reconstruct an fake junitxml, as a crash-safe way to record logs, just for these tests. 
- Fields to reconstruct are [detailed in the Pydantic test model](https://github.com/tenstorrent/tt-github-actions/blob/main/.github/actions/collect_data/src/parsers/python_pytest_parser.py).

jameszianxuTT:
More complexity arises from the probability of test timeout and hangs. Ideally, an in-stages parallel test can be used to quarantine test jobs after they timeout but CI limitations make that complicated, and may require aggregating a report across multiple workflows.

A simpler alternative of running each test in its own job (and runner), and only running discovered tests with depth < compile, makes the possibility of intra-job conflict impossible, but introduces the possibility of runner starvation due to hung jobs. Statistical simulation of runner scheduling suggests this is highly improbable with the jobs we have currently.

jameszianxuTT:
Closing this as "complete", as much as it will be, based on existing model benchmarking PRs in tt-torch. The infra to perform benchmarking and dynamic collection of compile depth does now exist.

PR #446, opened 2025-03-17
Second pass for schema change - stop recording frontend info

jameszianxuTT:
### Ticket
#256 

### Problem description
Redundant model_name and frontend fields are recorded into the XML; which the Pytest result pydantic model was recently changed to omit.

### What's changed
Remove redundant reporting of the model name and frontend now that it won't break the pydantic model ingestion

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -64,11 +64,8 @@ class ModelTester:
         self.record_property = record_property_handle
         self.record_tag_cache = {}  # Holds for tags to be written out at finalize()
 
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
Never mind, this is actually useful behavior that I will exploit for #443. 

As long as the pydantic model doesn't get too strict, we should pack it here because model_name / frontend is added at the _start_ of the test; so in the event the test fails, we can recover the properties/model_name and pack it into the tags while reconstructing our xml

PR #452, opened 2025-03-18
Crash Safe JunitXML Writer

jameszianxuTT:
### Ticket
#443 

### Problem description
We only run model tests as a regression test, and we lack a way to run a benchmark on a models, and capture their actual achieved compile depth.

### What's changed
- Create a crashsafe, transactional XML logger that creates an XML to mirror the output of junitxml and optionally overload record_property when passing --crashsafe to the pytest as a CLI option
- Create utils to extract all pytests under tests/models that are full-eval related and write them to a parallel build matrix
- Create utils to rewrite the crashsafe XML into a pydantic model compliant format and pack max compile depth regardless of the termination mode of the original test

### Checklist

### DIFF
--- a/.github/workflows/depth-benchmarks.yml
+++ b/.github/workflows/depth-benchmarks.yml
@@ -1,0 +1,31 @@
+# This workflow automates running model depth benchmarks
+
+name: Depth Benchmarks
+
+on:
+  schedule:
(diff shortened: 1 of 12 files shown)


jameszianxuTT:
Closing for now to not spam CI

PR #453, opened 2025-03-18
Few op-by-op flow improvements (xfail output shown, better tar-results error) and fixes (use self.tokenizer) to models

### LINKED ISSUE #451: Few xfail op-by-op flow tests (whisper, musicgen, ...)  failing earlier with self.tokenizer missing errors since 0decda6
FYI @jameszianxuTT - what we talked about offline today

Superset shows op-by-op tests passing EXECUTE dropped from 104 to 99 on latest nightly. Quick look and few new failing tests in [latest](https://github.com/tenstorrent/tt-torch/actions/runs/13915163923) tt-torch nightly, they are xfail tests failing with this cryptic error in "Tar Results" job:

```
Run tar cvf musicgen_small_38937056031.tar .
OCI runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/__w/tt-torch/tt-torch/results/models/tests/") set in config.json failed: no such file or directory: unknown
```

It just means the model doesn't produce valid test output directory, and we don't see output in logs unless running with `--runxfail` flag locally (or removing xfail tag), example:

```
pytest -svv --runxfail tests/models/musicgen_small/test_musicgen_small.py::test_musicgen_small --op_by_op_torch

<snip>

        if is_token_output and not hasattr(self, "tokenizer"):
>           raise ValueError(
                "is_token_output is set to True. Please set `self.tokenizer` inside _load_model method."
            )
E           ValueError: is_token_output is set to True. Please set `self.tokenizer` inside _load_model method.

tests/utils.py:44: ValueError
```

A few AIs here:

1. Fix tests affected (thx @LPanosTT for suggested change)
2. Better error message in "Tar results" step when $pwd is not present
3. Maybe consider deploying `--runxfail` flag to track output in per-test logs that are marked xfail instead of it being silent in CI logs.

kmabeeTT:
### Ticket
See #451 for details - Closes #451

### Problem description
Address a few concerns from this ticket.

 - A couple categories of test fixes after Lewis' change at 0decda6 that prevented models from running (use self.tokenizer, and is_token_output). There are other issues exposed by whisper, musicgen that will require followup debug/fixes.
 - Add better error message in "Tar results" when $pwd (test output dir) used as working-directory doesn't exist (was occuring in these xfail tests that failed to run)
  - Add `--runxfail` arg to op-by-op flow so we can at least see xfail test output in log files instead of nothing (useful for when signature changes, but we don't know, happened today)
  
### Checklist

codegen model is back to passing in op-by-op flow with these changes.

### DIFF
--- a/.github/workflows/run-op-by-op-model-tests.yml
+++ b/.github/workflows/run-op-by-op-model-tests.yml
@@ -313,7 +313,7 @@ jobs:
 
           pytest_log="test_${counter}.log"
 
-          pytest -svv "$test_case" --op_by_op_torch > "$pytest_log" 2>&1
+          pytest -svv --runxfail "$test_case" --op_by_op_torch > "$pytest_log" 2>&1
           exit_code=$?
 
           echo "====== BEGIN LOG: $test_case ======" >> full_job_output.log
@@ -340,9 +340,17 @@ jobs:
     - name: Tar results
       if: success() || failure()
       shell: bash
-      working-directory: ${{ steps.strings.outputs.test-output-dir }}
       run: |
-        tar cvf ${{ matrix.build.name }}_${{ steps.fetch-job-id.outputs.job_id }}.tar .
+        TEST_DIR="${{ steps.strings.outputs.test-output-dir }}"
+        OUTPUT_TAR="${{ matrix.build.name }}_${{ steps.fetch-job-id.outputs.job_id }}.tar"
+
+        if [ ! -d "$TEST_DIR" ]; then
+          echo "ERROR: Test output dir '$TEST_DIR' does not exist. Please check if test ran properly."
+          exit 1
+        fi
+
+        cd "$TEST_DIR"
+        tar cvf "$OUTPUT_TAR" .
 
     - name: Upload test folder to archive
       if: success() || failure()
--- a/tests/models/codegen/test_codegen.py
+++ b/tests/models/codegen/test_codegen.py
@@ -50,7 +50,7 @@ def test_codegen(record_property, mode, op_by_op):
         mode,
         compiler_config=cc,
         record_property_handle=record_property,
-        is_transformers_generation=True,
+        is_token_output=True,
     )
     results = tester.test_model()
 
--- a/tests/models/musicgen_small/test_musicgen_small.py
+++ b/tests/models/musicgen_small/test_musicgen_small.py
@@ -11,14 +11,14 @@ from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
 
 class ThisTester(ModelTester):
     def _load_model(self):
-        self.processor = AutoProcessor.from_pretrained("facebook/musicgen-small")
+        self.tokenizer = AutoProcessor.from_pretrained("facebook/musicgen-small")
         model = MusicgenForConditionalGeneration.from_pretrained(
             "facebook/musicgen-small"
         )
         return model.generate
 
     def _load_inputs(self):
-        inputs = self.processor(
+        inputs = self.tokenizer(
             text=[
                 "80s pop track with bassy drums and synth",
                 "90s rock song with loud guitars and heavy drums",
--- a/tests/models/whisper/test_whisper.py
+++ b/tests/models/whisper/test_whisper.py
@@ -12,7 +12,7 @@ from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
 class ThisTester(ModelTester):
     def _load_model(self):
         # load model and processor
-        self.processor = WhisperProcessor.from_pretrained(
+        self.tokenizer = WhisperProcessor.from_pretrained(
             "openai/whisper-small", torch_dtype=torch.bfloat16
         )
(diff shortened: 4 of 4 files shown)


jameszianxuTT (approved):

tests/models/segformer/test_segformer.py
```diff
@@ -26,7 +26,7 @@ def _load_model(self):
     def _load_inputs(self):
         url = "http://images.cocodataset.org/val2017/000000039769.jpg"
         image = Image.open(requests.get(url, stream=True).raw)
-        inputs = self.processor(images=image, return_tensors="pt")
+        inputs = self.tokenizer(images=image, return_tensors="pt")
```
AleksKnezevic, line 29: Did you run this locally? Is the image processor also called tekenizer?

AleksKnezevic (approved):
Let a comment on one of the changes, as long as you ran locally, it's good to go!

tests/models/segformer/test_segformer.py
AleksKnezevic: Is this test not in ci? It's in `.github/workflows/run-op-by-op-model-tests.yml`

AleksKnezevic: Also, feel free to merge in the validated changes if you want. Or hold off on all of them, either way is good with me.

kmabeeTT: Oh man thanks for pointing that out.. turns out a grep I ran earlier to check had a typo and they were all in CI. I didn't need to update the ones outside of these (those that call model.generate in `_load_model()` is a clue, from talking to Lewis)...

```
 tests/models/codegen/test_codegen.py               | 2 +-
 tests/models/musicgen_small/test_musicgen_small.py | 4 ++--
 tests/models/whisper/test_whisper.py               | 6 +++---
 ```
 
 Reducing set of changes and will merge.

**Review by AleksKnezevic — APPROVED** [review]

File: .github/workflows/run-op-by-op-model-tests.yml [hunk 2]
```diff
@@ -340,9 +340,17 @@ jobs:
...
       shell: bash
-      working-directory: ${{ steps.strings.outputs.test-output-dir }}
       run: |
-        tar cvf ${{ matrix.build.name }}_${{ steps.fetch-job-id.outputs.job_id }}.tar .
+        TEST_DIR="${{ steps.strings.outputs.test-output-dir }}"
+        OUTPUT_TAR="${{ matrix.build.name }}_${{ steps.fetch-job-id.outputs.job_id }}.tar"
+
+        if [ ! -d "$TEST_DIR" ]; then
+          echo "ERROR: Test output dir '$TEST_DIR' does not exist. Please check if test ran properly."
+          exit 1
```
jameszianxuTT, line 349: Just going through nightlies today. 

RMGB is skipped as of Lewis commit two days ago and therefore generate no logs (the logs mark it as PASSED since a skipped test has exit code 0 and that's how the workflow assesses pass/fail)  https://github.com/tenstorrent/tt-torch/actions/runs/13938641379/job/39011557096

This would be a valid case in which no logs are created for an ostensibly passing test and therefore no logs are produced. I think **this should exit 0**. Though, maybe I should think more about the cases in which it is possible for a test to run but produce no output, where that is an indication of an error uncaught by the test execution. 

I think if there is a problem, it should cause the previous step to exit with nonzero exit code, and we should not try to catch such a problem by observing the symptom that no output logs are created. However, I am not sure how such a failure could arise.

kmabeeTT: Thanks a lot @jameszianxuTT for your reply. Good points, I think this topic does require a bit more thinking, though I am inclinded to leave this change of mine alone for now (due to first bullet below)...

The link you shared is almost the case my change was aiming to improve (and kind of does, was thinking about xfail tests, but skip tests also are related), [here](https://github.com/tenstorrent/tt-torch/actions/runs/13915163923/job/38937054477#step:12:1) was the RMGB test before my change where "Tar results" just hits ugly error because it cannot chdir to working-directory 

```
OCI runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/__w/tt-torch/tt-torch/results/models/tests/") set in config.json failed: no such file or directory: unknown
```

Some current thinking...

- my change right now doesn't change pass/fail frequency AFAIK, just replaces a ~unhelpful message with a slightly more helpful message when test output dir doesn't get created.

- Ideally do want to continue catch when "xfail" tests (that at least ran before) regress to stop producing output directory... thankfully it is ~caught in "tar results" step currently when test output directory not created

- Ideally want to find a way to avoid having "skip" tests failing in tar-results stage (they were before my change too though, no change here). It could be worked around by returning warning and "exit-0" in tar-results step if test output directory doesn't exist but then we can't catch xfail regressions in prev bullet, so that's not great.

- Might make sense... and Aleks was thinking this too... to have assert in pytest/model itself that results json file is created to catch regressions (which wouldn't occur for skipped tests). Though this problem is more about directory not existing than the .json files, maybe it has to do with when all tests in a single group (like RMBG) are skipped, not sure.

PR #456, opened 2025-03-19
Fix Deepseek Qwen Fatal Bus Error [#291]

ddilbazTT:
Deepseek Qwen op no. 9552 fails with out-of-memory error. This is expected to be caught by try-except block. Instead, the whole program crashes.

The error is related to multiprocessing in Python. There is no full isolation between parent and child processes. So, the child process running out of memory kills the parent process.

Thus, now there is a memory limit to inputs that we pass on to the child process, and if exceeded, we push the inputs to disk and send a reference to disk address to child process. This is not the ideal solution, but it works using python's multiprocessing module.

- helper functions such as `get_tensor_size`, `get_inputs_size` help determine size of inputs in bytes
- `run_op` determines if inputs exceed 1 GB or not
- if exceeding, inputs are dumped to a temporary file using pickle
- `execute_process` uses `pickle.load()` if inputs exceed memory limit
- file where inputs are stored is deleted at the end of `run_op`

As a note, PyTorch also faces a similar issue, and seems like they have extended torch.multiprocessing to handle large tensor sharing over processes.

### Ticket
#291

### DIFF
--- a/tt_torch/dynamo/executor.py
+++ b/tt_torch/dynamo/executor.py
@@ -6,10 +6,12 @@ import onnx
 import onnxruntime as ort
 import tt_mlir
 import time
+import pickle
 import faulthandler
 import sys
 import re
 import tempfile
+import os
 import multiprocessing as mp
 from tt_torch.tools.utils import (
     CompileDepth,
@@ -19,6 +21,29 @@ from tt_torch.tools.utils import (
     OpCompilationStatus,
 )
 from typing import Union
+
+
+def get_tensor_size(tensor):
+    """Calculate the memory size of a tensor in bytes."""
+    if isinstance(tensor, torch.Tensor):
+        return tensor.element_size() * tensor.nelement()
+    return 0
+
+
+def get_inputs_size(inputs):
+    """Calculate the total memory size of inputs in bytes."""
+    total_size = 0
+
+    if isinstance(inputs, torch.Tensor):
+        total_size += get_tensor_size(inputs)
+    elif isinstance(inputs, (list, tuple)):
+        for item in inputs:
+            total_size += get_inputs_size(item)
+    elif isinstance(inputs, dict):
+        for item in inputs.values():
+            total_size += get_inputs_size(item)
+
+    return total_size
 
 
 def compile_process(receiver, sender, ttir_event, ttnn_event, json_event):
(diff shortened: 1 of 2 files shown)


ddilbazTT:
I have tested if `deepseek_qwen` would run on a nightly test and it successfully did: [https://github.com/tenstorrent/tt-torch/actions/runs/13933783498](https://github.com/tenstorrent/tt-torch/actions/runs/13933783498)

tt_torch/dynamo/executor.py
```diff
@@ -21,6 +23,29 @@
 from typing import Union
 
 
+def get_tensor_size(tensor):
+    """Calculate the memory size of a tensor in bytes."""
```
jazpurTT, line 27: I've seen that for short comments we usually use # for comments, do we have a standard?

tt_torch/dynamo/executor.py
```diff
@@ -21,6 +23,29 @@
...
+
+    if isinstance(inputs, torch.Tensor):
+        total_size += get_tensor_size(inputs)
+    elif isinstance(inputs, (list, tuple)):
+        for item in inputs:
+            total_size += get_inputs_size(item)
+    elif isinstance(inputs, dict):
+        for item in inputs.values():
+            total_size += get_inputs_size(item)
+
```
jazpurTT, line 45: error when it is neither of the options?

tt_torch/dynamo/executor.py
```diff
@@ -381,7 +443,38 @@ def run_op(self, binary, *inputs):
             self.file_stderr = tempfile.NamedTemporaryFile(mode="w+t", delete=False)
             self.stderror_redirected = True
 
-        obj = {"binary": binary, "inputs": inputs, "dump_file": self.file_stderr.name}
+        inputs_size = get_inputs_size(inputs)
+
+        large_input = inputs_size > 1 * 1024 * 1024 * 1024  # 1GB limit
```
jazpurTT, line 448: can we set a global variable? instead of hardcoding the value here

tt_torch/dynamo/executor.py
ddilbazTT: I used it as the function description, not a comment.

tt_torch/dynamo/executor.py
ddilbazTT: Should we have a global value for 1073741824 (1024 * 1024 * 1024) or a function like gb_to_bytes where we pass in 1 in this instance?

ddilbazTT: I decided having this limit in the op_by_op executor instead of a global since it would achieve same result. Added a helper function to convert gb to bytes

AleksKnezevic (changes requested):

tt_torch/dynamo/executor.py
```diff
@@ -381,7 +443,38 @@ def run_op(self, binary, *inputs):
...
+        obj = {
+            "binary": binary,
+            "dump_file": self.file_stderr.name,
+            "large_input": large_input,
+        }
+
+        inputs_file_path = None
+        if large_input:
+            try:
+                temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".pkl")
```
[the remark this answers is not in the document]
ddilbazTT: Makes sure file isn't deleted after closed: [https://docs.python.org/3/library/tempfile.html](https://docs.python.org/3/library/tempfile.html)

tt_torch/dynamo/executor.py
```diff
@@ -381,7 +443,38 @@ def run_op(self, binary, *inputs):
...
+
+        obj = {
+            "binary": binary,
+            "dump_file": self.file_stderr.name,
+            "large_input": large_input,
+        }
+
+        inputs_file_path = None
+        if large_input:
+            try:
```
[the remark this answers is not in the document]
ddilbazTT: I assume you mean at the except block, we should set `obj["inputs"] = None`. Like this:

 ```
       if large_input:
            try:
                temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".pkl")
                inputs_file_path = temp_file.name
                temp_file.close()

                with open(inputs_file_path, "wb") as f:
                    pickle.dump(inputs, f)

                obj["inputs_file_path"] = inputs_file_path
            except Exception as e:
                print(f"Error saving inputs to disk: {e}")
                if inputs_file_path and os.path.exists(inputs_file_path):
                    try:
                        os.remove(inputs_file_path)
                    except OSError:
                        pass
                large_input = False
                obj["inputs"] = None
        else:
            obj["inputs"] = inputs
```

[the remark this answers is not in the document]
AleksKnezevic: no, even if saving to disk succeeds, inputs should be None

AleksKnezevic (approved):

jazpurTT (approved):

jazpurTT:
> @jazpurTT, can you please check your model with this change too?

I did, it is compiling op_by_op, no more Bus Error. There is just an ATOL error during verify_output.

PR #460, opened 2025-03-19
Update .gitignore to ignore all hidden files and folders

hshahTT:


.gitignore
```diff
@@ -1,4 +1,5 @@
 *.pyc
+.*
```
AleksKnezevic, line 2: Will this ignore `.github`?  We don't want that.

hshahTT: Yup it would've. James mentioned .github/ is how we run our daily pipelines, so I closed this PR. Will replace with specific folders and reopen it.

PR #462, opened 2025-03-19
Change the way unused outputs are calculated in op_by_op_mode

AleksKnezevic:
### Problem description
Unused outputs were not being detected correctly in op-by-op mode causing many ops to fail needlessly. The issue is that torch.export (introduces by @LPanosTT several days ago) places getitem nodes on node outputs whether they're needed or not.

### What's changed
Checking whether the user get_items have users restores original behaviour. 

### Checklist

### DIFF
--- a/tt_torch/dynamo/torch_backend.py
+++ b/tt_torch/dynamo/torch_backend.py
@@ -253,12 +253,10 @@ class TorchExecutor(OpByOpExecutor):
             getitem_nodes = []
             graph_node.meta["val"] = node.meta["val"]
 
+            # if the output of the getitem node is not used, we don't append it to the graph
+            unused_output = [len(user.users) == 0 for user in node.users]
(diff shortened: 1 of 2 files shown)


LPanosTT (approved):

kmabeeTT (approved):
looks good to my eyes, appreciate the extra info print about ops executed

brataTT (approved):
Thanks for fixing it 🙌 
I see test_basic.py::test_unused_output passes
We can also add an assert checking the assumption `len(node.users) == len(node.meta["tensor_meta"]` unless torch always constructs it this way

AleksKnezevic:
> Thanks for fixing it 🙌 
> I see test_basic.py::test_unused_output passes
> We can also add an assert checking the assumption `len(node.users) == len(node.meta["tensor_meta"]` unless torch always constructs it this way

Node users is always all of the outputs, as there are getitem nodes, so don't think this will work.

Issue #465, opened 2025-03-20
nightly test failure mar 20, 2025


ddilbazTT:
The nightly tests give an error similar to:

`Failed to execute 394/1636: aten.permute.default: stat: path should be string, bytes, os.PathLike or integer, not NoneType`

Coming from previous days merged PR: [https://github.com/tenstorrent/tt-torch/pull/456](https://github.com/tenstorrent/tt-torch/pull/456)

PR #467, opened 2025-03-20
Fix nightly test failure March 20, 2025 [#465]

ddilbazTT:
Fix "stat: path should be string, bytes, os.PathLike or integer, not NoneType" error. Check all mnist ops execute on PR.

### Ticket
#465 

### Problem description
Issue arose from 
`if os.path.isfile(inputs_file_path):`
when inputs_file_path is None, this gives an "stat: path should be string, bytes, os.PathLike or integer, not NoneType" error.
This was replaced with
`if inputs_file_path and os.path.isfile(inputs_file_path):`

+ added a workflow job that checks all mnist ops run on PR.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -64,6 +64,7 @@ jobs:
         echo "dist-dir=$(pwd)/dist" >> "$GITHUB_OUTPUT"
         echo "job-id=$JOB_ID" >> "$GITHUB_OUTPUT"
         echo "test_report_path_torch=report_torch_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
+        echo "test_report_path_mnist=report_mnist_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
         echo "test_report_path_models=report_models_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
         echo "test_report_path_onnx=report_onnx_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
 
@@ -120,6 +121,22 @@ jobs:
         name: test-reports-onnx-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
         path: ${{ steps.strings.outputs.test_report_path_onnx }}
 
+    - name: Run Mnist Op-By-Op and Check All Ops Execute
+      shell: bash
+      run: |
+        source env/activate
+        TT_TORCH_CHECK_ALL_OPS_RUN=1 pytest -svv tests/models/mnist/test_mnist.py --op_by_op_torch \
+            --junit-xml=${{ steps.strings.outputs.test_report_path_mnist }} \
+            --cov=tt_torch --cov-report term --cov-report xml:coverage.xml --cov-append
+
+    - name: Upload Test Report Mnist
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        name: test-reports-mnist-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
+        path: ${{ steps.strings.outputs.test_report_path_mnist }}
+
+
     - name: Show Test Report
       uses: mikepenz/action-junit-report@v5
       if: success() || failure()
@@ -145,6 +162,6 @@ jobs:
       continue-on-error: true
       uses: codecov/test-results-action@v1
       with:
-        files: ${{ steps.strings.outputs.test_report_path_torch }}, ${{ steps.strings.outputs.test_report_path_onnx }}
+        files: ${{ steps.strings.outputs.test_report_path_torch }}, ${{ steps.strings.outputs.test_report_path_onnx }}, ${{ steps.strings.outputs.test_report_path_mnist }}
         disable_search: true
         token: ${{ secrets.CODECOV_TOKEN }}
--- a/tt_torch/dynamo/torch_backend.py
+++ b/tt_torch/dynamo/torch_backend.py
@@ -379,7 +379,6 @@ class TorchExecutor(OpByOpExecutor):
         if self.stderror_redirected:
             os.unlink(self.file_stderr.name)
             self.stderror_redirected = False
-
         return outputs
 
     def __call__(self, *inputs):
--- a/tt_torch/tools/utils.py
+++ b/tt_torch/tools/utils.py
@@ -262,6 +262,7 @@ class CompilerConfig:
         self._enable_intermediate_verification = False
         self.dump_debug = False
         self.dump_info = False
+        self.check_all_ops_execute = False
         self._verify_op_by_op = False
         self.typecast_inputs = True
         self.runtime_device = None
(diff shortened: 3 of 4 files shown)


ddilbazTT:
I created a function to check if all ops ran:
```
    def verify_all_ops_run(self):
        if self._check_all_ops_run:
            assert (
                self.num_executed_ops == self.total_ops
            ), f"Expected all ops to run in {self.model_name}, only {self.num_executed_ops}/{self.total_ops} ops ran"
        print(f"Verified all ops ran in {self.model_name}")
```
I wasn't sure if this should be a standalone function or a part of `save_unique_ops`
```
    def save_unique_ops(self):
        unique_op_dict = {}
        pytest_test = os.environ.get("PYTEST_CURRENT_TEST")
        # 'PYTEST_CURRENT_TEST' is unavailable for the scripts executed/invoked
        # with python command; use 'sys.argv[0]' instead.
        if pytest_test is None:
            pytest_test = sys.argv[0]

        # Keep slashes, replace all non-alphanumeric characters with underscore.
        pytest_test = re.sub(r"[^A-Za-z0-9_/]", "_", pytest_test)

        for key, op in self.unique_ops.items():
            unique_op_dict[key] = op.to_dict()
        output_file = Path(f"{self.results_path}{pytest_test}_unique_ops.json")
        print(f"#####  Saving unique ops to {output_file}#####  ")
        output_file.parent.mkdir(exist_ok=True, parents=True)
        with open(output_file, "w") as f:
            json.dump(unique_op_dict, f)

        self.total_ops = len(unique_op_dict)
        self.num_executed_ops = 0
        for op in unique_op_dict.values():
            if op["compilation_status"] == OpCompilationStatus.EXECUTED:
                self.num_executed_ops += 1

        print(f"{self.num_executed_ops}/{self.total_ops} ops executed")
```

tt_torch/dynamo/torch_backend.py
```diff
@@ -379,7 +379,7 @@ def run_gm_op_by_op(self, *inputs):
         if self.stderror_redirected:
             os.unlink(self.file_stderr.name)
             self.stderror_redirected = False
-
+        self.compiler_config.verify_all_ops_run()
```
AleksKnezevic, line 382: This can go as part of `save_unique_ops`

tt_torch/tools/utils.py
```diff
@@ -256,13 +256,16 @@ def __init__(self):
...
+        self.num_executed_ops = 0
+        self.total_ops = 0
         self.op_by_op_backend = OpByOpBackend.TORCH
         self.enable_consteval = False
         self._consteval_parameters = False
         self._enable_intermediate_verification = False
         self.dump_debug = False
         self.dump_info = False
         self._verify_op_by_op = False
+        self._check_all_ops_run = False
```
AleksKnezevic, line 268: maybe `check_all_ops_execute`. I don't think you need to have a property with it too. compile_depth and check_all_ops_execute are set independently by the user, so one can be set before the other.

tt_torch/tools/utils.py
```diff
@@ -378,13 +398,20 @@ def save_unique_ops(self):
...
-        total_ops = len(unique_op_dict)
-        num_executed_ops = 0
+        self.total_ops = len(unique_op_dict)
+        self.num_executed_ops = 0
         for op in unique_op_dict.values():
             if op["compilation_status"] == OpCompilationStatus.EXECUTED:
-                num_executed_ops += 1
+                self.num_executed_ops += 1
+
+        print(f"{self.num_executed_ops}/{self.total_ops} ops executed")
```
AleksKnezevic, line 407: Here you can just:
```python
if self.check_all_ops_execute:
  assert num_executed_ops == total_ops
```

AleksKnezevic (approved):
I left a few comments, feel free to merge after addressing them.

AleksKnezevic:
> I created a function to check if all ops ran:
> 
> ```
>     def verify_all_ops_run(self):
>         if self._check_all_ops_run:
>             assert (
>                 self.num_executed_ops == self.total_ops
>             ), f"Expected all ops to run in {self.model_name}, only {self.num_executed_ops}/{self.total_ops} ops ran"
>         print(f"Verified all ops ran in {self.model_name}")
> ```
> 
> I wasn't sure if this should be a standalone function or a part of `save_unique_ops`
> 
yeah, I saw that. I think you can just leave it as part of save_unique_ops. Then we don't need another api for it.

tt_torch/tools/utils.py
ddilbazTT: should I remove the print statement?

.github/workflows/run-tests.yml
```diff
@@ -120,6 +121,22 @@ jobs:
         name: test-reports-onnx-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
         path: ${{ steps.strings.outputs.test_report_path_onnx }}
 
+    - name: Run Mnist Op-By-Op and Check All Ops Execute
+      shell: bash
+      run: |
+        source env/activate
+        TT_TORCH_CHECK_ALL_OPS_RUN=1 pytest -svv tests/models/mnist/test_mnist.py --op_by_op_torch \
+            --junit-xml=${{ steps.strings.outputs.test_report_path_mnist }} \
```
jameszianxuTT, line 129: This is probably out of scope of this PR but:

We don't really get much information from the junitxml reports, from models run in op_by_op. We [don't actually record these junitxml logs in the op by op pipeline](https://github.com/tenstorrent/tt-torch/blob/9d68c3f1d93aada7022a0238cc0ac37ab5d42933/.github/workflows/run-op-by-op-model-tests.yml#L316) and instead dump them into a log file for a different report generation process. 

The test reports generated by that process (using the [download-report](https://github.com/tenstorrent/tt-torch/blob/9d68c3f1d93aada7022a0238cc0ac37ab5d42933/.github/workflows/nightly-tests.yml#L45-L54) job to process the model result) would yield a richer report that shows clearly which ops make it to which compile depths.

At this point, a regression caught by this new test would:
- Cause CI to fail
- Generate a JUnitXML
- Require you to read the full logs of this workflow to figure out which ops fail, as dumped to stdout through -svv

A followup PR to improve this test coverage might be to:
1) Expand the number of tests for better op coverage
2) Run the download-report job or add that inline to this workflow to produce a clearer model summary and upload that as an artefact

jameszianxuTT (approved):

.github/workflows/run-tests.yml
ddilbazTT: I see your point but do we need download-report on each PR? Maybe as you said we can just dump a log file and upload it as an artifact.

ddilbazTT: However, I agree with perhaps expanding the number of models we test for full execution coverage. We could also have a separate workflow for it, I just made a band-aid solution under `.github/workflows/run-tests.yml`

jameszianxuTT: This is more of a nitpick than anything, I'm not sure how actually useful it will be to create a nicer report since you'll probably end up running MNIST locally anyways, so it won't really save you time.
 
Technically the logs written to stdout are already accessible through GH actions by going to "view raw logs" so there's no need to dump and upload it as an artifact. This improvement I'm suggesting will probably be a bit of work and complicate the workflow (eg. by adding download-report/splicing the parse-op-by-op.py invocation inline into this workflow) for little benefit than "niceness" of logs

PR #468, opened 2025-03-20
Some parse_op_by_op_results.py QOL updates (git info, totals, conditonal formatting)

kmabeeTT:
### Ticket
None

### What's changed
 - Move summary worksheet creation to function since it's growing, and create few helper functions
 - Add more totals for Compiled-to-TTNN and Executed-On-Device, and various conditional color formatting
 - See comment in PR for example of what it looks like before/after

### Checklist

### DIFF
--- a/results/parse_op_by_op_results.py
+++ b/results/parse_op_by_op_results.py
@@ -6,11 +6,34 @@ import os
 import json
 import csv
 import xlsxwriter
+from xlsxwriter.utility import xl_rowcol_to_cell
 from mdutils.mdutils import MdUtils
 from pathlib import Path
 
 import subprocess
 import re
+
+import datetime
+import subprocess
+
+# Function to get git branch and commit
+def get_git_info():
+    try:
+        branch = (
+            subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"])
+            .decode("utf-8")
+            .strip()
+        )
+        commit = (
+            subprocess.check_output(["git", "rev-parse", "HEAD"])
+            .decode("utf-8")
+            .strip()
+        )
+    except Exception as e:
+        branch = "N/A"
+        commit = "N/A"
+    return branch, commit
+
 
 # Script to parse the results of the unique ops json files and combine them into a spreadsheet
 # This script parses models compiled into stable hlo / TTIR op by op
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Screenshot 2025-03-20 at 3 33 12 PM

Here is comparison before/after with things as they stand right now (local subset of test reports downloaded from a recently nightly)

results/parse_op_by_op_results.py
```diff
@@ -175,6 +198,201 @@ def create_test_dirs():
...
+    # Get current date, time and git branch, commit info.
+    branch, commit = get_git_info()
+    now = datetime.datetime.now()
+    date_str = now.strftime("%Y-%m-%d")
+    time_str = now.strftime("%H:%M:%S")
+
+    info_str = (
+        f"Generated on {date_str} at {time_str} from Branch: {branch} Commit: {commit}"
+    )
+    worksheet.write(0, 0, info_str)
```
jameszianxuTT, line 285: nit: Putting this in 0,0 and then offsetting all the data in the actual spreadsheet might not be the best way to embed such metadata. This might have been better to embed in a separate worksheet, or even in the [document properties](https://xlsxwriter.readthedocs.io/example_doc_properties.html)

Dealing with this offset and the hardcoding of row/col indices will mean that future bulk adjustments of this data within this worksheet summary page are going to be pretty painful (eg. if you did want to reverse this and not embed the info_str and move all the actual data to 0,0, you'd have to change pretty much every index reference in this function). But adding enums for row/col data or computing row/col indices dynamically might be overengineering this; 

I guess it just depends on your vision for how much you'd want to modify this worksheet in the future

jameszianxuTT (approved):

results/parse_op_by_op_results.py
kmabeeTT: Thanks a lot - yeah, it's a valid concern, I feel the same way. Did want this particular line up front so it's easy to share a screenshot and compare screenshots side by side, but yeah, the column/row maintenance is not ideal.  Going to keep it for now, and maybe in future revisit with a cleanup task (feel free if you have a good idea there too).

kmabeeTT:
Fixed a small bug in the script (Total Ops per Compile Depth included the first row still, made it use a variable to prevent accidental including of wrong range, rebased/squashed/force-pushed).

PR #469, opened 2025-03-20
Ensure the strides we are providing to tt::runtime::createTensor desc…

LPanosTT:
### Problem description
Torch tensors which are contiguous may not always have a stride attribute which indicates that the tensor is contiguous. This occurs when the left-most dimension is 1. In such cases, when the left-most dimension is 1, the stride value for that dimension will never be used, and so they do not bother to compute it even when calling .contiguous().

I found this issue from 2017 to get this information: https://discuss.pytorch.org/t/tensor-stride-not-updated-by-contiguous/4549

### What's changed
Our runtime expects that this stride is accurate. So, we will require that this torch tensor is contiguous and then calculate a fully-accurate stride for it.
### Checklist

### DIFF
--- a/tt_torch/csrc/bindings.cpp
+++ b/tt_torch/csrc/bindings.cpp
@@ -85,10 +85,21 @@ static tt::runtime::Tensor create_tensor(const torch::Tensor &tensor) {
     shape.push_back(1);
   }
 
-  auto stride =
-      std::vector<uint32_t>(tensor.strides().begin(), tensor.strides().end());
-  if (stride.empty()) {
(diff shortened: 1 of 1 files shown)


AleksKnezevic (approved):

kmabeeTT:
Thanks for debug today with me on this and fix, all those errors are gone in the handful of models I just ran, autogenerated xlsx attached, looked through the stage-6 errors and none were this error anymore.

Screenshot 2025-03-20 at 6 35 26 PM

kmabeeTT (approved):

PR #470, opened 2025-03-20
Move tag flush to happen before the verify_against_golden asserts

jameszianxuTT:
### Ticket
Part of #443 

### Problem description
Tags are not flushed since the PCC / ATOL verification check happens as an assert and prevents tester finalization if the check fails and we are asserting PCC / ATOL

### What's changed
- Move the tag flushing to happen as soon as PCC and ATOLs are generated. This is the last thing to record into tags anyways (at this point) so it is valid to write out tags at this point.
- Remove tester.finalize() from every step since it happens inside golden verification

### Checklist

### DIFF
--- a/tt_torch/tools/verify.py
+++ b/tt_torch/tools/verify.py
@@ -20,6 +20,7 @@ def verify_against_golden(
     required_pcc=0.99,
     required_atol=None,
     relative_atol=None,
+    preflush_tags=None,
 ):
     assert (required_atol is not None) != (
         relative_atol is not None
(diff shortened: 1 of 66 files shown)


tt_torch/tools/verify.py
```diff
@@ -108,6 +109,9 @@ def verify_against_golden(
             )
             passed_atol = False
 
+        if preflush_tags is not None:
```
jameszianxuTT, line 112: It is a little messy to move the context of tags / logging into this verify_against_golden function, but the alternative of moving the assert around would I think be equally messy

jameszianxuTT:
Never mind, a more comprehensive refactor should be done to make verify_outputs and verify_against_golden purely functional.

PR #482, opened 2025-03-24
Create run-full-model-execution-tests-nightly.yml w/ 13 tests passing

kmabeeTT:
### Ticket
None

### Problem description
 - We have more passing full-model-execution coming online, and desire to keep them out of push/pr and only in nightly, so need to create a home for them.  Also, since we have a bunch of new ones passing now, would like to keep them protected. I suspect in the future this will be more regular occurrance and could benefit from automated flow to do this that is WIP.

### What's changed
 - Create new file run-full-model-execution-tests-nightly.yml used only by nightly-tests.yml that contains 13 tests passing
 - They are part of set of 23 tests passing op-by-op flow to execute but 10 other tests have runtime issues or pcc fails, so excluding
 - Ideally wanted these to be inside existing yaml and have matrix be dynamically configured for push vs nightly, but not simple.
 - 2 parallel jobs take about 10 minutes each

### Checklist

Link to passing run at TOTT before creating this branch and PR: https://github.com/tenstorrent/tt-torch/actions/runs/14024930263/job/39262196611 (I moved 2 tests from group 2 to group 1 after it ran to better balance load).

### DIFF
--- a/.github/workflows/run-full-model-execution-tests-nightly.yml
+++ b/.github/workflows/run-full-model-execution-tests-nightly.yml
@@ -1,0 +1,145 @@
+name: Run Full Model Execution Tests
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      docker-image:
+        description: 'Docker image to use for build'
+        required: true
+        type: string
+      run-codecov:
+        description: 'Run code coverage reports'
+        required: false
+        type: string # properly a boolean but autocast to string when passed in using 'with' inputs
+        default: 'true'
+  workflow_run:
+    workflows: [Build] # backref to run-build as dependency
+    types: [completed]
+
+jobs:
+  tests:
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        build: [
+          {
+            runs-on: wormhole_b0, name: "eval_1", tests: "
+                  tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vgg19]
+                  tests/models/dpr/test_dpr.py::test_dpr[full-eval]
+                  tests/models/albert/test_albert_masked_lm.py::test_albert_masked_lm[full-albert/albert-xlarge-v2-eval]
+                  tests/models/albert/test_albert_masked_lm.py::test_albert_masked_lm[full-albert/albert-large-v2-eval]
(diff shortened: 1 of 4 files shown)


.github/workflows/run-full-model-execution-tests-nightly.yml
```diff
@@ -0,0 +1,145 @@
...
+jobs:
+  tests:
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        build: [
+          {
+            runs-on: wormhole_b0, name: "eval_1", tests: "
+                  tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vgg19]
```
AleksKnezevic, line 29: Are any of these duplicated in the compile tests?

kmabeeTT: Yep a few. When that PR gets merged tonight will update this one to remove the duplicates from the e2e compile test list yaml.

jameszianxuTT:
I think that it is a bit risky to duplicate the full-model-execution tests just to support different subsets of tests. I'm imagining someone may accidentally modify one of these files and not make a corresponding modification to the other, which could lead to some errant behaviour. 

As you say:
> Ideally wanted these to be inside existing yaml and have matrix be dynamically configured for push vs nightly, but not simple.

I'll get on this soon, it would be a relevant feature to #443. There should be precedent in other repos' CI for this dynamic matrix configuration. For now, I think we could go with this stopgap solution until that dynamic test matrix generation infra can be set up.

jameszianxuTT (approved):

.github/workflows/run-full-model-execution-tests-nightly.yml
kmabeeTT: Done. Merged the PR to add bunch of tests to compile list, and rebased this PR to remove duplicates fromt the compile list (which was all 13 tests actually)... so this PR essentially promotes them from nightly compile list to new nightly execute list.  Resolving.

kmabeeTT:
thx @jameszianxuTT for review!

> I think that it is a bit risky to duplicate the full-model-execution tests just to support different subsets of tests. I'm imagining someone may accidentally modify one of these files and not make a corresponding modification to the other, which could lead to some errant behaviour.

Yeah, though to be clear it's not the tests themselves that are duplicated, just the body of this file and it's actions etc. Agreed though, don't love it.  Had a few different ideas to solve it, but ended up getting ugly/complicated and not as quick as this... let's discuss this week :)

PR #498, opened 2025-03-25
Pass in system desc during compilation

brataTT:
### Ticket
Link to Github Issue

### Problem description
No system desc was being passed in during compilation. This meant default system_desc was being created with assumed values. This change passes in a real system_desc.

### What's changed
Temporary file with system desc is created before compilation step where ttnn gets a system desc attribute inserted into it. Then the path is passed in.

### Checklist

### DIFF
--- a/tt_torch/csrc/tt-mlir-interface.cpp
+++ b/tt_torch/csrc/tt-mlir-interface.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <cstdlib>
+#include <filesystem>
 #include <iostream>
 
 #include "tt-mlir-interface.hpp"
@@ -131,6 +132,30 @@ compileTTIRToTTNN(std::string_view code) {
     pm.enableIRPrinting();
   }
   mlir::tt::ttnn::TTIRToTTNNBackendPipelineOptions options;
+
+  if (std::filesystem::path system_desc_path = std::getenv("SYSTEM_DESC_PATH");
+      !system_desc_path.empty() && std::filesystem::exists(system_desc_path)) {
+
+    // TODO: avoid copying to temp file and create system desc every time
+    // https://github.com/tenstorrent/tt-torch/issues/580
+    std::filesystem::path system_desc_temp_path =
+        std::filesystem::temp_directory_path() /
+        (system_desc_path.stem().string() + "_tmp" +
+         system_desc_path.extension().string());
+
+    std::filesystem::copy_file(
+        system_desc_path, system_desc_temp_path,
+        std::filesystem::copy_options::overwrite_existing);
+
+    options.systemDescPath = system_desc_temp_path;
+  } else { // SYSTEM_DESC_PATH is not set or the file does not exist
+    std::cout << "WARNING: "
+              << (system_desc_path.empty()
+                      ? "SYSTEM_DESC_PATH environment variable is not set"
+                      : "The file in SYSTEM_DESC_PATH does not exist")
+              << ". Default system description will be used instead.\n";
+  }
+
   mlir::tt::ttnn::createTTIRToTTNNBackendPipeline(pm, options);
 
   // Run the pass manager.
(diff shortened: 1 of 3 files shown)


tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -96,6 +97,13 @@ std::string compileStableHLOToTTIR(std::string_view code) {
   return buffer;
 }
 
+void createSystemDesc(const std::string &path) {
+  if (std::ifstream(path).good()) {
+    return;
+  }
+  tt::runtime::getCurrentSystemDesc().first.store(path.c_str());
```
brataTT, line 104: I'm creating system_desc just before the compilation step.
Would it be more appropriate to create it during build?
Didn't do it that way because mlir doesn't build system desc during build.

tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -131,6 +139,8 @@ compileTTIRToTTNN(std::string_view code) {
     pm.enableIRPrinting();
   }
   mlir::tt::ttnn::TTIRToTTNNBackendPipelineOptions options;
+  options.systemDescPath = "/tmp/system_desc.ttsys";
```
AleksKnezevic, line 142: Why is this hard-coded?

tt_torch/csrc/tt-mlir-interface.cpp
AleksKnezevic: Isn't the ttnn graph different because we don't pass is the system desc?

tt_torch/csrc/tt-mlir-interface.cpp
brataTT: Ah yes, we can also add an env var `SYSTEM_DESC_PATH` like mlir does
And define it in `env/activate`

tt_torch/csrc/tt-mlir-interface.cpp
brataTT: The systemDescAttr added to the graph is different
Passing it into `TTIRToTTNN` options makes sure the correct system_desc gets inserted into the ttnn graph

jnie-TT: @brataTT system desc has to be created during run, specifically before running ttir to ttnn pipeline - I think the location you're creating it is fine

tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -96,6 +97,13 @@ std::string compileStableHLOToTTIR(std::string_view code) {
   return buffer;
 }
 
+void createSystemDesc(const std::string &path) {
+  if (std::ifstream(path).good()) {
```
jnie-TT, line 101: Nit: I don't have a strong opinion here, but to prevent stale/outdated system descriptors, we might just want to always create it regardless of it's existence. It's pretty quick and doesn't add much overhead to compile.

mmanzoorTT:
@brataTT Question: How system descriptor was generated before this change?

AleksKnezevic (approved):

tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -131,6 +132,28 @@ compileTTIRToTTNN(std::string_view code) {
     pm.enableIRPrinting();
   }
   mlir::tt::ttnn::TTIRToTTNNBackendPipelineOptions options;
+
```
jnie-TT, line 135: NIt: create an issue for this and maybe explain why we need to make a copy

jnie-TT (approved):
Comment in line but otherwise looks good.

tt_torch/csrc/tt-mlir-interface.cpp
brataTT: Created issue https://github.com/tenstorrent/tt-torch/issues/580
Will follow up with tt-mlir commit to accept device as parameter in `getCurrentSystemDesc`

PR #499, opened 2025-03-25
changing mistral model to 7B

jazpurTT:
### Ticket
Follow up to #475 

### Problem description
Change mistral model to 7B version. 

### What's changed
- Updated mistral test to load 7B model
- Added mistral to op_by_op test pool

### Checklist
- [x] Tested model locally and verified we do not need HF_TOKEN to load and run model


tests/models/mistral/test_mistral_7b.py
```diff
@@ -0,0 +1,54 @@
...
+        inputs["input_ids"] = torch.randint(1, 30001, (1, 12))
+        inputs["attention_mask"] = torch.ones_like(inputs["input_ids"])
+        return inputs
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.xfail(reason="Mistral-7B is too large to fit on a single device")
```
AleksKnezevic, line 31: Can we pytest.skip it in the test only for the full model as opposed to xfailing each one?

jazpurTT: done

PR #505, opened 2025-03-25
Replace openDevice and closeDevice functions with new mesh API

hshahTT:
### Ticket
[493](https://github.com/tenstorrent/tt-torch/issues/493)

### Problem description
A new API for opening and closing devices was provided in this [commit](https://github.com/tenstorrent/tt-mlir/commit/436290947d53368774ab272c26405d90a3a540de) to the tt-mlir repo. The API we currently use will soon be deprecated

### What's changed
- Added PyBindings for the new API functions
- Replaced occurrences of `openDevice` and `closeDevice` with `openMeshDevice` and `closeMeshDevice`

### Checklist

### DIFF
--- a/tt_torch/tools/utils.py
+++ b/tt_torch/tools/utils.py
@@ -14,8 +14,12 @@ import math
 import sys
 import shutil
 
-# from tt_mlir import open_device, close_device, is_runtime_debug_enabled
-from tt_mlir import open_mesh_device, close_mesh_device, MeshDeviceOptions, is_runtime_debug_enabled
+from tt_mlir import (
+    open_mesh_device,
+    close_mesh_device,
+    MeshDeviceOptions,
+    is_runtime_debug_enabled,
+)
 
 
 """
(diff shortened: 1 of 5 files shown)


AleksKnezevic (approved):

tt_torch/tools/utils.py
```diff
@@ -269,6 +274,8 @@ def __init__(self):
         self.enable_async = False
         self.cache_preprocessed_constants = False
         self.inline_parameters = False
+        self.mesh_device_shape = [1, 1]
+        self.mesh_device_options = MeshDeviceOptions()
```
jnie-TT, line 278: I think we can remove `self.enable_async` and combine it with `MeshDeviceOptions`, so we can just set enable_async in mesh_device_options here instead of setting it in `initialize_device`

hshahTT: Good point, updated it. PTAL again.

jnie-TT: awesome, thanks for looking at this

jnie-TT (approved):

PR #507, opened 2025-03-26
Add fetch-depth:0 to all checkout actions to collect entire git history

jameszianxuTT:
### Ticket
None 
### Problem description
Recent ND crashes with actions/checkout@v4 causing 
random CI jobs to fail at checkout with git corruption-like
symptoms

### What's changed
Force checkout of entire history with fetch-depth:0 instead of 
default fetch-depth: 1

### Checklist


jameszianxuTT:
-

Issue #449, opened 2025-03-17
Add libgl1 and libglx-mesa0 to docker


AleksKnezevic:
Currently libgl1 and libglx-mesa0 are installed where needed by models (yolo). Add this to the docker so it can be done in one place.

AleksKnezevic:
Once in, remove it from all the tests that install it.

Issue #316, opened 2025-02-19
Change torchvision build to create a wheel


AleksKnezevic:
Copying the venv is too large, torchvision build should create a wheel instead.

PR #508, opened 2025-03-26
Adding libraries and wheel build to docker

jazpurTT:
### Ticket
#316 #449

### Problem description
- The torchvision wheel is built every time we run CI and we want to move it into the docker.
- libgl1 and libglx-mesa0 are being installed every op-by-op and full-model CI runs for the yolo models and we want to move it into the docker
- We have duplicate libraries in requirements.txt

### What's changed
- Added torchvision wheel build into dockerfile.ci (installed dependencies in dockerfile.base) and copied wheel in a /opt/dist/ dir to final stage so its available when generating new containers
- Added a check in env/activation to check if wheel is available in /opt/dist/ and install wheel from that file
- removed duplicated libraries from requirements.txt
- added install of libgl1 and libglx-mesa0 to dockerfile.base
- removed install of libgl1 and libglx-mesa0 from *.yml files

### Checklist
- [x] Tested the build of the new Docker image https://github.com/tenstorrent/tt-torch/actions/runs/14049828360
- [x] Checked that CI run does not build wheel from scratch for each job that runs source env/activate https://github.com/tenstorrent/tt-torch/actions/runs/14066665483
- [x]  Tested models that use torchvision https://github.com/tenstorrent/tt-torch/actions/runs/14067904743

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -32,7 +32,9 @@ RUN apt-get update && apt-get install -y \
     libboost-all-dev \
     jq \
     curl \
-    lcov
+    lcov \
+    libgl1 \
+    libglx-mesa0
 
 # Install python3.11 packages
 RUN python3.11 -m pip install \
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -11,6 +11,8 @@ RUN apt-get update && apt-get install -y \
     python3-dev \
     python3-venv \
     python3-pip \
+    python3.11-dev \
+    python3.11-venv \
     git \
     git-lfs \
     libhwloc-dev \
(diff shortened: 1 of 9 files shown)


env/activate
```diff
@@ -39,9 +43,12 @@ else
     TORCHVISION_USE_VIDEO_CODEC=0 TORCHVISION_USE_FFMPEG=0 CC=clang CXX=clang++ _GLIBCXX_USE_CXX11_ABI=1 USE_CUDA=OFF python setup.py bdist_wheel
```
AleksKnezevic, line 43: Do we still need to do this?

jazpurTT: I thought so, this would cover cases were someone activates the venv not in a container from the Docker image in our repo. We'd still want to build the wheel and copy it into /dist/ once so it doesn't have to be built every time the venv is activated

.github/Dockerfile.base
```diff
@@ -30,7 +32,15 @@ RUN apt-get update && apt-get install -y \
     libboost-all-dev \
     jq \
     curl \
-    lcov
+    lcov \
+    libgl1 \
+    libglx-mesa0
+
+# Install python3.11 packages
```
vmilosevic, line 39: We already have some pp dependencies in requirements.txt, how is this related? Maybe we should pre-install everything from equirements.txt?

jazpurTT: These are the dependencies needed to run the setup.py script for the torchvision wheel build in the .ci files. If I understand correctly requirements.txt is for the venv which should be separate from the docker.

AleksKnezevic:
> Do we need to provide torchvision wheel to customers, is that going to be part of release?

No, I don't think so. the activation script will build it if it's not available.

AleksKnezevic (approved):

PR #509, opened 2025-03-26
Fix onnx model tester regressions and add onnx mobilenet to On PR

brataTT:
### Ticket
None

### Problem description
Onnx model tester infra had regressions that went uncaught because nothing in CI was using it

### What's changed
Added fixes for regressions due to recent test infra changes:
- skip token mismatch commit 586dd1f
- model name deduplication commit 155fb9a

Added test using this infra to CI to catch future regressions

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -379,6 +379,7 @@ class OnnxModelTester(ModelTester):
         assert_pcc=True,
         assert_atol=True,
         record_property_handle=None,
+        model_group="generality",
     ):
 
         super().__init__(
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -379,6 +379,7 @@ def __init__(
         assert_pcc=True,
         assert_atol=True,
         record_property_handle=None,
+        model_group="generality",
```
AleksKnezevic, line 382: This is a priority model, model_group needs to be `"red"`

AleksKnezevic (changes requested):

tests/utils.py
brataTT: The onnx model is marked "red" in the [test file](https://github.com/tenstorrent/tt-torch/pull/509/files#diff-4a4010be5a21b199e0fe54f79c1d6176de4d7e5dbb34cc48b6009e07ce8bd458R61)
This line denotes the default for all models, so I'll keep it as is

AleksKnezevic (approved):

PR #511, opened 2025-03-27
Change pcc threshold for PASSED status to be model specific

jameszianxuTT:
### Ticket
None

### Problem description
We're using 0.99 as global threshold for marking a 
model's bringup_status as PASSED.

### What's changed
Use model specific required_pcc override to makr a 
model's bringup_status as PASSED. Constructor
default is 0.99, unless overridden.

### Checklist


kmabeeTT (approved):
Thanks !

Issue #330, opened 2025-02-20
Microsoft Phi


AleksKnezevic:
Add tests for phi 1, 1.5, 2, 3 and 4. Models [here](https://huggingface.co/microsoft)

AleksKnezevic:
Tests should run in op-by-op nightly pipelines.

hshahTT:
Added a PR for v1, 1.5, and 2: https://github.com/tenstorrent/tt-torch/pull/515

Will add v3 and v4 in another PR.

kmabeeTT:
Let me know if you're unable to loop back to v3,v4 variants in the next week or so.  I see these in another spreadsheet as what we're looking for, at least:

```
Microsoft Phi-3-mini
Microsoft Phi-3.5-MoE
Microsoft Phi-3.5-vision
Microsoft Phi-4
```

hshahTT:
Unassigning myself to focus on multi-device work

AleksKnezevic:
@ssaliceTT, can you please take a look at this?

PR #515, opened 2025-03-28
Add Microsoft Phi v1, v1.5, v2 models

hshahTT:
### Ticket
[330](https://github.com/tenstorrent/tt-torch/issues/330)

### Problem description
Add Phi v1, v1.5, v2 models in nightly op-by-op pipeline.

### What's changed
Added a single model test for these models since they all have the same usage format. v3 and v4 have a slightly different format so will add them later.

Nightly op-by-op pipeline succeeded: https://github.com/tenstorrent/tt-torch/actions/runs/14115345926

Note: these 3 models took ~2.5 hours to run (including building all dependencies). A single Phi model will take around ~30-40 min to run. We may want to consider just adding one of these instead of all 3.

### Checklist


jameszianxuTT:
Just curious - did you try this in [full-eval] (EXECUTE) mode? I would expect it to fail, but it would be a useful datapoint if you know how long it takes before crashing. Not necessary to close this PR though

jameszianxuTT (approved):

tests/models/phi/test_phi_1_1p5_2.py
```diff
@@ -0,0 +1,73 @@
...
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        is_token_output=True,
```
AleksKnezevic, line 60: since this is a customer priority model, `model_group` should be `red`. Once you add that it's good to go.

AleksKnezevic (approved):

kmabeeTT (approved):
Nice job! Thanks Het.

tests/models/phi/test_phi_1_1p5_2.py
hshahTT: Added it, will wait for the tests to pass again and submit

AleksKnezevic: I've you've already ran these in nightly once, no need to re-run to merge just for this change.

hshahTT: Sorry I meant I'm waiting for the automated all-green tests to pass. Github won't let me merge this PR until those succeed.

ddilbazTT:
I vaguely remember you bringing up that each of these models take 50 minutes, and @AleksKnezevic saying we could only run one of them instead of all of them. It seems we are running all in the op-by-op flow, just wanted to check that's what we expect.

AleksKnezevic:
@ddilbazTT, since this is a high priority model, let's run all of them (at least for the first few times) to ensure there's nothing surprising.

hshahTT:
Thanks for the reviews everyone!

@jameszianxuTT I ran the full-eval locally and it failed after about 30 min (I can rerun it and confirm the runtime). It fails after the model compiles and while it's comparing the decoded token output to the goldens (see screenshot)

image

PR #518, opened 2025-03-28
Benchmark Updates

jameszianxuTT:
### Ticket
#443 

### Problem description
- We don't have a test auto-discovery system right now to quarantine new tests and dynamically generate the test matrix
- The individual test timeout is set to 10 hours, meaning a test that hangs may cause the entire workflow to be killed by the 10h limit and for the benchmark report to not be autogenerated
- There is no translation table to insert bringup_status values for the cross-FE dashboard based on benchmarked max achieved compile depths

### What's changed
- Added utility functions to discover added and removed pytests, using set difference between pytests in tree and pytests listed in the benchmark YAML
- Reduced timeout from 600m -> 500m
- Added translation for bringup_status and injected that into tags during XML reconstruction

### Checklist


jameszianxuTT:
Some of the changes in this PR are set up steps for future CI improvements (the benchmark test duration CSV, the load balancer script), as they require more complex tooling to be fully integrated. 

The 3 actual changes here are :
- Manually adding in the new tests this week into quarantined test buckets
- Recoloring the spreadsheet cells for easy reading (tracking promotable/regression status)
- Reducing the test timeout to avoid the workflow-level T/O autokill

jameszianxuTT:
Highlight verification from [a test run](https://github.com/tenstorrent/tt-torch/actions/runs/14137062111)
image

These were all originally "No"s but I manually edited them to "Yes"

.github/benchmark_durations.csv
```diff
@@ -0,0 +1,136 @@
+full_test_name,achieved_compile_depth,model_name,test_execution_time
```
jameszianxuTT, line 1: This file is here as reference and to be consumed in a future CI infra improvement, but not currently used.

AleksKnezevic (approved):

kmabeeTT (approved):

PR #520, opened 2025-03-28
Add helper script results/download_artifacts.py to easily download+unzip GH artifacts from CI

kmabeeTT:
### Ticket
None

### What's changed
 - Add the new script, by default, download all from lastest tt-torch/nightly and unzip them
 - Fully configurable to also grab latest friom given repo, branch, workflow, or just give a specific run-id.  Also can filter results to only download some files. Or just list artifacts instead of dl.
 - Example: python3 results/download_artifacts.py --filter full-logs
- Use git rev-parse --short=8 HEAD when generating artifact names to get consistent 8 character hashes (was occasionally changing between reruns of same job on CI).

### DIFF
--- a/.github/workflows/generate-benchmark-report.yml
+++ b/.github/workflows/generate-benchmark-report.yml
@@ -58,7 +58,7 @@ jobs:
       shell: bash
       run: |
         branch=$(git rev-parse --abbrev-ref HEAD | sed 's/[":<>|*?\\\/]/_/g' | tr -d '\r\n')
-        commit=$(git rev-parse --short HEAD)
+        commit=$(git rev-parse --short=8 HEAD)
         date=$(date +'%Y-%m-%d')
         echo "branch=$branch" >> "$GITHUB_OUTPUT"
         echo "commit=$commit" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/generate-model-report.yml
+++ b/.github/workflows/generate-model-report.yml
@@ -105,7 +105,7 @@ jobs:
       shell: bash
       run: |
         branch=$(git rev-parse --abbrev-ref HEAD | sed 's/[":<>|*?\\\/]/_/g' | tr -d '\r\n')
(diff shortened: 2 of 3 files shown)


kmabeeTT:
@jameszianxuTT this has feedback you gave me on Monday and one other change:

- Added --run-id 1234 to specify run-id instead of grabbing latest from particular workflow
- Skipped makedirs if --list and adjust print to say the same
- Added recursive unzip, works for your case
- Also, if repeat artifacts are found with same name, just download the latest.

I might have introduced a bug right before I pushed, getting some unzip errors on some cases, will fix and re-force push, probably something small.

kmabeeTT:
Noticed that some artifacts (the xls file) was seeing different git commit hashes (7 characters instead of 8) when the same job is rerun multiple times, causing both xlsx to appear as unique artifacts (not desirable):

```
python3 results/download_artifacts.py  --filter xlsx --no-unzip
Fetching workflow runs for nightly-tests
Downloading artifacts for run 14121749737 (created on 20250328) into folder 'nightly-tests_artifacts_20250328_run_id_14121749737'.
Downloading artifact 'models_op_per_op_2025-03-28_main_d310a276.xlsx' (ID: 2837265935) to nightly-tests_artifacts_20250328_run_id_14121749737/models_op_per_op_2025-03-28_main_d310a276.xlsx.zip
Downloading artifact 'models_op_per_op_2025-03-28_main_d310a27.xlsx'  (ID: 2840264136) to nightly-tests_artifacts_20250328_run_id_14121749737/models_op_per_op_2025-03-28_main_d310a27.xlsx.zip
```

Fixed it by this part of my change, squashed into same commit to specify we always want 8 character or more git commit hashes.

```
     - Use git rev-parse --short=8 HEAD when generating artifact names to get
       consistent 8 character hashes (was occasionally changing between
       reruns of same job on CI).
```

jameszianxuTT (approved):
Looks great, thanks for making this script!

Issue #513, opened 2025-03-27
reduce_graph error with inplace ops


ddilbazTT:
Whilst adding gliner #424 I discovered a problem with `reduce_graph` in `pass_pipeline`

`reduce_graph` is removing inline ops and resulting in a shape mismatch between expected inputs.

Before reduce graph:

```
opcode         name                  target                              args                               kwargs
-------------  --------------------  ----------------------------------  ---------------------------------  --------
placeholder    args_0                args_0                              ()                                 {}
placeholder    args_1                args_1                              ()                                 {}
placeholder    args_2                args_2                              ()                                 {}
placeholder    args_3                args_3                              ()                                 {}
placeholder    args_4                args_4                              ()                                 {}
placeholder    args_5                args_5                              ()                                 {}
placeholder    args_6                args_6                              ()                                 {}
call_function  lt                    aten.lt.Tensor                      (args_1, args_2)                   {}
call_function  convert_element_type  prims.convert_element_type.default  (lt, torch.int64)                  {}
call_function  index                 aten.index.Tensor                   (args_3, [args_4, args_5])         {}
call_function  index_put             aten.index_put.default              (args_0, [args_4, args_6], index)  {}
call_function  copy__default         aten.copy_.default                  (args_0, index_put)                {}
output         output                output                              ((convert_element_type,),)         {}
```
After reduce graph:

```
opcode         name                  target                              args                        kwargs
-------------  --------------------  ----------------------------------  --------------------------  --------
placeholder    args_1                args_1                              ()                          {}
placeholder    args_2                args_2                              ()                          {}
call_function  lt                    aten.lt.Tensor                      (args_1, args_2)            {}
call_function  convert_element_type  prims.convert_element_type.default  (lt, torch.int64)           {}
output         output                output                              ((convert_element_type,),)  {}
```
```
Shapes of example_inputs:
  Input 0: torch.Size([1, 5, 768])
  Input 1: torch.Size([1, 5])
  Input 2: torch.Size([1, 1])
  Input 3: torch.Size([1, 250, 768])
  Input 4: torch.Size([5])
  Input 5: torch.Size([5])
  Input 6: torch.Size([5])
```

Instead of comparing input1 and input 2 with less than (<), it compares input 0 and input 1, resulting in the following error:

```
[...]

Issue #424, opened 2025-03-11
Gliner


AleksKnezevic:
Potential source: https://huggingface.co/urchade/gliner_large-v2

ddilbazTT:
Whilst adding this model, I discovered a problem with `reduce_graph` in `pass_pipeline`

`reduce_graph` is removing inline ops and resulting in a shape mismatch between expected inputs.

Before reduce graph:

```
opcode         name                  target                              args                               kwargs
-------------  --------------------  ----------------------------------  ---------------------------------  --------
placeholder    args_0                args_0                              ()                                 {}
placeholder    args_1                args_1                              ()                                 {}
placeholder    args_2                args_2                              ()                                 {}
placeholder    args_3                args_3                              ()                                 {}
placeholder    args_4                args_4                              ()                                 {}
placeholder    args_5                args_5                              ()                                 {}
placeholder    args_6                args_6                              ()                                 {}
call_function  lt                    aten.lt.Tensor                      (args_1, args_2)                   {}
call_function  convert_element_type  prims.convert_element_type.default  (lt, torch.int64)                  {}
call_function  index                 aten.index.Tensor                   (args_3, [args_4, args_5])         {}
call_function  index_put             aten.index_put.default              (args_0, [args_4, args_6], index)  {}
call_function  copy__default         aten.copy_.default                  (args_0, index_put)                {}
output         output                output                              ((convert_element_type,),)         {}
```

After reduce graph:
```
opcode         name                  target                              args                        kwargs
-------------  --------------------  ----------------------------------  --------------------------  --------
placeholder    args_1                args_1                              ()                          {}
placeholder    args_2                args_2                              ()                          {}
call_function  lt                    aten.lt.Tensor                      (args_1, args_2)            {}
call_function  convert_element_type  prims.convert_element_type.default  (lt, torch.int64)           {}
output         output                output                              ((convert_element_type,),)  {}
```

```
Shapes of example_inputs:
  Input 0: torch.Size([1, 5, 768])
  Input 1: torch.Size([1, 5])
  Input 2: torch.Size([1, 1])
  Input 3: torch.Size([1, 250, 768])
  Input 4: torch.Size([5])
  Input 5: torch.Size([5])
  Input 6: torch.Size([5])

```
Instead of comparing input1 and input 2 with less than (<), it compares input 0 and input 1, resulting in the following error:

```
[...]

PR #536, opened 2025-04-01
Add urchade/gliner_large-v2 #424

ddilbazTT:
Added to op-by-op-flow.

This model unearthed an issue with `reduce_graph` #513

### Ticket
#424 

### Problem description
Added Gliner.
- Discovered an unexpected behaviour from `reduce_graph` when dealing with inplace ops, commented out `reduce_graph`. Issue is tracked --> #513 
- Realized an aten.gather op with float dim values which brought the whole program down. Hard-coded a fix in `run_gm_op_by_op`.
- Edited parsing/ helping functions which iterate over output. Output contains string/ int/ etc so adopted accordingly. 
- Alphabetically ordered `requirements.txt`


ddilbazTT:
@kmabeeTT how do we decide whether to add to nightly or weekly tests?

kmabeeTT:
> @kmabeeTT how do we decide whether to add to nightly or weekly tests?

Here is some info:

1. run-op-by-op-model-tests-nightly.yml was the right call, start off in there.  
2. Can also add to run-e2e-tests.yml (e2e compile) if you happened to try it and it's passing (it just uses env-var `TT_TORCH_COMPILE_DEPTH=TTNN_IR` alongside full-eval).
3. Once it is passing full-model-execute (ie. full-eval without env-var to limit compile depth) in the future, can then move it from nightly (wip/debug) to weekly (stable) op-by-op lists.

tt_torch/dynamo/torch_backend.py
```diff
@@ -355,6 +370,10 @@ def run_gm_op_by_op(self, *inputs):
                         )
                         tensor = node.target(*args, **node.kwargs)
                 else:
+                    # TODO: This is a temporary fix to handle gather op with non-int dimension values.
```
AleksKnezevic, line 373: This is a little too hacky for me. Can we either reduce the scope of this test (i.e. call a certain module) to bypass this? Alternatively, does it work in the SHLO op by op flow.

ddilbazTT: I agree with your sentiment, and I'll look into it.

ddilbazTT: Stablehlo also has a problem with inplace copy op, `verify_ir` fails. Which made me realize we don't need to do  
```
        if name == "aten::copy_":
            raise ValueError(f"inline ops are not supported: {name}")
            return None, None
```
We could simply do:

```
        if not verify_ir(module):
            raise ValueError(f"Failed to verify IR for {name}")
            return None, None
```
I will change this, and I am also debugging the gather op failure. Another idea I had was a decomposition to cast any non-int dimension to int.

AleksKnezevic: The decomp could work if you can add a dtype specific overload.

ddilbazTT: Aleks found the issue to be related to self.typecase_inputs offline. Basically, a bool dimension is passed into the torch op as a float.

ddilbazTT: Also, verify_ir doesn't work when using torch op-by-op flow. So I will not be switching to that.

tt_torch/dynamo/shlo_backend.py
```diff
@@ -353,7 +353,7 @@ def shlo_op_by_op(self):
                     op.compilation_status = OpCompilationStatus.EXECUTED
                 except Exception as e:
                     print(f"Failed to execute {idx}/{num_ops}: {op.op_name}: {e}")
-        self.binary = binary
+        # self.binary = binary
```
AleksKnezevic, line 356: error?

ddilbazTT: I think it is partially an error, partially intended. I will open an issue to add onnx model proto to stablehlo executor, like we add a program when the input is a pytorch graph. If binary isn't generated of the last op, this can result in failure. I encountered that when I was trying to run the model with stablehlo op by op.

tt_torch/dynamo/passes.py
```diff
@@ -96,7 +96,7 @@ def pass_pipeline(gm: torch.fx.GraphModule, example_inputs, compiler_config):
 
     gm = bypass_redundant_getitem(gm)
 
-    reduce_graph(gm)
+    # reduce_graph(gm)
```
AleksKnezevic, line 99: please link to issue why it's disabled.

tt_torch/dynamo/shlo_backend.py
ddilbazTT: [https://github.com/tenstorrent/tt-torch/issues/543](https://github.com/tenstorrent/tt-torch/issues/543)

AleksKnezevic (approved):
Small nit below, once fixed, please feel free to merge.

ddilbazTT:
@kmabeeTT cannot run the model end to end unfortunately, so I'm keeping it in nightly op-by-op test

kmabeeTT:
> @kmabeeTT cannot run the model end to end unfortunately, so I'm keeping it in nightly op-by-op test

No problem. Curious to see if it comes end to end though, did you check? We can always promote it later if you didn't check.

ddilbazTT:
@kmabeeTT I checked, didn't run end-to-end :/ By the way, are we using the "model_group" tag? For this design, would it be "red"?

PR #542, opened 2025-04-01
Split red/generality tests into different test modules in the same file

jameszianxuTT:
### Ticket
#547 

### Problem description
- Database design issue where per-parameterization `group` values cannot be stored and cause subtle and intermittent data anomalies if present.
- No CI self-verification for renamed or deleted test cases in Pytest scripts

### What's changed
- Split red/generality models in timm and torchvision image detection tests into separate test modules in the same file, refactor out a common test runner to avoid duplicated logic.
- Add CI verification job to run-tests yaml to check for duplicated and nonexistent tests attempting to be called

### Checklist


jameszianxuTT:
This is a necessary workaround but should be reverted once the database schema is fixed.

jameszianxuTT:
There currently isn't a good way to check if a renamed pytest got renamed in all the test yamls it's referenced in. This is a common error (which I made in the first iteration of this PR) that will lead to tests failing.

Created a script to verify that there are no nonexistent tests or duplicated tests within a yaml file

kmabeeTT (approved):
Thanks James! This is great (solving ND reporting issue) and unfortunate at the same time that "model_group" (used to be internal to python file) now needs to spill into our testlist yamls, which brings risk of mistakes, hence the new verification script/job.  Makes sense, we talked offline. I understand this is temporary workaround and can be ~reverted once DB schema is updated later.  

Approved - thanks for working through them and figuring them out!

Issue #496, opened 2025-03-25
Add helpful error


AleksKnezevic:
Add helpful user-facing error messages with a link to file an issue for various failure modes.

jazpurTT:
@AleksKnezevic I pushed a commit with my understanding of the task, it looks like the points of entry for the tt-torch compilers pipeline are backend() and compile_onnx() so I added a wrapper to add an error message if an exception is raise at any point during those functions.

PR #549, opened 2025-04-02
Add user-facing error message for tt-torch compilation failures #496

jazpurTT:
### Ticket
#496 

### Problem description
Add user-facing error messages with a link to file an issue for tt-torch compilation failures

### What's changed
- added function to append tt-torch compilation error message to stderr with instructions on where/how to open tt-torch issue when there is compilation failures
- added functions as a wrapper for backend() and compiler_onnx()

### Checklist

### DIFF
--- a/tt_torch/dynamo/backend.py
+++ b/tt_torch/dynamo/backend.py
@@ -30,6 +30,7 @@ from tt_torch.tools.utils import (
     OpByOpBackend,
     CompilerConfig,
     CompileDepth,
+    tt_torch_error_message,
 )
 
 
@@ -135,6 +136,7 @@ def _base_backend(gm, example_inputs, compiler_config):
     return executor
 
 
+@tt_torch_error_message
 def backend(gm, example_inputs, options=None):
     assert isinstance(gm, torch.fx.GraphModule), "Backend only supports torch graphs"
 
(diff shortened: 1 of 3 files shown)


ddilbazTT (commented):
What if the error happens in the ModelTester level? Are we only concerned about errors in the backend?

LPanosTT:
> What if the error happens in the ModelTester level? Are we only concerned about errors in the backend?

If something fails in the model tester that would more likely be an infra issue. Plus users will probably not be using the ModelTester infra as they'll want to just `torch.compile` their models. If we do want to cover this case perhaps its best to print a different error message.

jazpurTT:
> > What if the error happens in the ModelTester level? Are we only concerned about errors in the backend?
> 
> If something fails in the model tester that would more likely be an infra issue. Plus users will probably not be using the ModelTester infra as they'll want to just `torch.compile` their models. If we do want to cover this case perhaps its best to print a different error message.

That was my thought behind limiting the message to backend and compile_onnx, and yeah for the tester we could add a separate error log if we think it's necessary

ddilbazTT (approved):

LPanosTT (approved):

AleksKnezevic (approved):

PR #555, opened 2025-04-02
Base yolov10 inference for future reference. 

ddilbazTT:
src is cloned from https://github.com/THU-MIG/yolov10, edited to address issue https://github.com/THU-MIG/yolov10/issues/523

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,7 @@ pillow
 pre-commit
 protobuf
 pybind11
+py-cpuinfo
 pytest
 pytest-cov
 pytorchcv
@@ -29,6 +30,7 @@ sentencepiece
 seaborn
 setuptools
 soundfile
+supervision
 tabulate
 tiktoken
 timm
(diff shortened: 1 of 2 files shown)


AleksKnezevic (commented):
Weren't you going to just commit the source of the repo to our repo as opposed to submoduling it in?

AleksKnezevic:
Then you could just fix in place.

requirements.txt
```diff
@@ -20,6 +20,7 @@ pillow
 pre-commit
 protobuf
 pybind11
+py-cpuinfo
```
AleksKnezevic, line 23: What's this?

ddilbazTT:
> Weren't you going to just commit the source of the repo to our repo as opposed to submoduling it in?

That's what I did... I'll try in another branch again, I don't know what went wrong there. I even removed and added the folder again.

requirements.txt
ddilbazTT: It was one of the dependencies that gave an error when missing. I can try to edit the code to remove it if we don't want it installed.

PR #562, opened 2025-04-03
Isolate Full Model Execute Pytest Processes and Quarantine Bert

jameszianxuTT:
### Ticket
None

### Problem description
Full model pytests are run in a single invocation of pytest and therefore generate a single test report and run in a single process. Process-killing failures are rare, but will cause the report to not be generated.

### What's changed
- Isolate full execute pytest processes so each generates individual report
- Restore op-by-op and compile jobs temporarily removed earlier today

### Checklist
- [x] CollectData job picks up all pytest results even though they are written to unique files. Post-processed report JSONs look well formed on inspection

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -24,9 +24,15 @@ jobs:
     secrets: inherit
     with:
       docker-image: ${{ needs.docker-build.outputs.docker-image }}
-  test_full_model_nightly:
+  test_op_by_op:
     needs: [docker-build, build]
-    uses: ./.github/workflows/run-full-model-execution-tests-nightly.yml
+    uses: ./.github/workflows/run-op-by-op-model-tests-nightly.yml
+    secrets: inherit
+    with:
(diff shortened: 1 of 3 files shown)


jameszianxuTT:
Validation:
- Ran a [test with isolated pytests](https://github.com/tenstorrent/tt-torch/actions/runs/14251552714) with a few of the full model execute jobs
- Verified that reports generated as a result of this test are not malformed by [Collect Data](https://github.com/tenstorrent/tt-torch/actions/runs/14253098602) job suggesting the parser is able to handle such a modification.
- I am using a similar process isolation technique for benchmark collection. The parser is able to parse those reports, and I can query it in the database, which provides precedent for this technique

jameszianxuTT:
I just checked - the data from my isolated pytest run also made it to the database and looks well formed - PCC, model name etc. are reported successfully

kmabeeTT (approved):
Looks good! I added a bullet point to description about restoring previously removed jobs.  Thanks for making changes, this will definitely help.

PR #564, opened 2025-04-04
Apr 4 Benchmark promotion + Benchmark Autopromotion Foundation

jameszianxuTT:
### Ticket
None

### Problem description
The benchmark promotion process when new pytests are added or pytests are renamed is manual and relies on committing a benchmark_durations file to tree.

### What's changed
- Create an artifact-based benchmark matrix auto promotion + quarantining + load balancing workflow (currently not used)
- Manually promote this week's benchmarks due to serdes issues
- Update download_artifacts tool to accept a variable output directory, which allows it to be more easily programmatically called by other tools 

## Todo
- Deal with JSON / YML serdes issues and Github actions syntactical issues

### Checklist
- [x] A dummy benchmark job passed


.github/workflows/generate-benchmark-matrix-dynamic.yml
```diff
@@ -0,0 +1,89 @@
...
+      id: generate_matrix
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }} # needed to download artifacts
+      working-directory: ${{ steps.strings.outputs.work-dir }}
+      run: |
+        source env/activate
+
+        # Using the actual built docker image seems to cause git corruption
+        apt install -y libgl1 libglx-mesa0
```
jameszianxuTT, line 83: I will amend this if I can figure out the git corruption due to incorrect string passing between jobs, but it is a workaround for now. Ideally we can use the CI docker but that requires some more testing

jameszianxuTT:
This autopromotion + load balancer + quarantiner is neat because it doesn't require us to commit the table of results to tree. 

Instead, we use @kmabeeTT 's artifact downloader tool to pull the output report artifact from last week's benchmark (run on main) and parse it to extract timings and a testlist. 

From there, we can find the set of tests we ran last week, and therefore the set of tests this week that should be quarantined (because they were not run last week). We can also load balance the tests because we know their test durations.

Note: This fails at the very last step, because of some complexity in how to pass a dynamically generated test matrix between jobs. I cannot pass a file to specify the test matrix, I must make a properly formatted matrix as a JSON string, and properly escape/unescape everything per Github Actions spec. 

Github actions also has issues passing around a multiline string between jobs, and I'm sure I'll run into more quirks as I try to fix this

.github/workflows/depth-benchmarks.yml
```diff
@@ -24,12 +24,17 @@ jobs:
...
+    uses: ./.github/workflows/generate-benchmark-matrix-dynamic.yml
+    secrets: inherit
   run-benchmark:
-    needs: [docker-build, build]
+    needs: [docker-build, build, generate-matrix-dynamic]
     uses: ./.github/workflows/run-depth-benchmark-tests.yml
     secrets: inherit
     with:
       docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      matrix-json: ${{ needs.generate-matrix-dynamic.outputs.matrix-json }} # Pass the dynamically generated matrix to the benchmark job
```
jameszianxuTT, line 37: Note - this output is not used in the run-benchmark job, so the preceding job currently has no effect.

.github/workflows/run-depth-benchmark-tests.yml
```diff
@@ -8,35 +8,42 @@ on:
...
+          {runs-on: 'wormhole_b0', name: 'bmk_13_qtn_test_mistral_7b[full-eval]', tests: ['tests/models/mistral/test_mistral_7b.py::test_mistral_7b[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_14_qtn_test_musicgen_small[full-eval]', tests: ['tests/models/musicgen_small/test_musicgen_small.py::test_musicgen_small[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_15_qtn_test_openpose[full-eval]', tests: ['tests/models/openpose/test_openpose.py::test_openpose[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_16_qtn_test_segment_anything[full-eval]', tests: ['tests/models/segment_anything/test_segment_anything.py::test_segment_anything[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_17_qtn_test_speecht5_tts[full-eval]', tests: ['tests/models/speecht5_tts/test_speecht5_tts.py::test_speecht5_tts[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_18_qtn_test_vit[full-eval]', tests: ['tests/models/vit/test_vit.py::test_vit[full-eval]']},
+          {runs-on: 'wormhole_b0', name: 'bmk_19_qtn_test_yolov5[full-eval]', tests: ['tests/models/yolov5/test_yolov5.py::test_yolov5[full-eval]']},
+        ]
+
```
jameszianxuTT, line 46: The next step here is to replace all these testlists between `build [ * ]` with a single `${{ inputs.matrix-json}}` but there are a bunch of github-actions-specific issues with passing it between the previous job and this one. I have just copy-pasted the result of the previous job here, but this should be fully automated by next week.

kmabeeTT (approved):

PR #569, opened 2025-04-06
Capture if op hangs/exceeds timeout and report on it in .json/xls

kmabeeTT:
### Ticket
None

### Problem description
- It's hard/impossible to tell what "Error message cannot be extracted" signature means, lately there is an uptick of them in a couple models under debug.

### What's changed
- Added explicit message "Timeout exceeded for op after xx seconds." that can be displayed/captured/reported on more easily to user knows they are hangs on silicon
- And add conditional highlighting for unknown-errors and timeouts as well as sums for their columns to calculate totals across models

### Checklist

### DIFF
--- a/results/parse_op_by_op_results.py
+++ b/results/parse_op_by_op_results.py
@@ -190,6 +190,10 @@ def parse_runtime_output(output):
 
         return error_message
 
+    timeout_message = re.search(r"Timeout exceeded for op.*", output)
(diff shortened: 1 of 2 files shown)


kmabeeTT:
Example of how new columns look:

Screenshot 2025-04-05 at 9 04 34 PM

jameszianxuTT (approved):

Issue #441, opened 2025-03-16
Allow user to run a model on a specific device


tapspatel:
If a user machine has multiple cards on it, we want to allow the user to compile for a specific device and run the model on that device. This requires some API that needs to be exposed to the user.

AleksKnezevic:
@hshahTT, please disregard. We assumed that some lower-level support needed for this to work was there, but unfortunately it isn't yet.

PR #572, opened 2025-04-07
Add a multi-device support API

hshahTT:
### Ticket
[441](https://github.com/tenstorrent/tt-torch/issues/441)

### Problem description
Allow a user to acquire multiple devices (if using a multi-device board), and select what chip to use for a specific model run.

### What's changed
- Created a singleton class called DeviceManager that provides functions to enable creation of parent and sub-meshes. It also provides some basic validation to make sure that the sub-meshes being created fit into the capacity of the parent mesh.
    - Note: currently we just validate that the area defined by the sub-mesh shape is valid, not the actual configuration (such as by taking mesh_offset into account). This will be added in a future PR.
  
- Removed device management from `CompilerConfig`. This includes:
   - Removing `runtime_device`, `mesh_device_shape`, and `mesh_device_options` member variables
   - Removing the `initialize_device` and `cleanup_device` functions

- Changed the backend logic:
   - Each backend function now takes an additional argument for `device` (with the default being `None`)
   - If `device` remains None, then a single device will automatically be used for execution and the user will not have to do any manual device management.
   - Instead of the user now just providing the backend with only `compiler_config`, the user can optionally choose to provide a device as well via the `torch.compile` call.
   - In order to support this, the convention of passing options to `torch.compile` has changed.
      - Previously, the `options` sent to the backend was a `CompilerConfig` object:
         ```
         cc = CompilerConfig()
         tt_model = torch.compile(model, backend=backend, options=cc)
         ```
      - Now, `options` is a map with `"compiler_config"` and `"device"` being the optional keys:
         ```
         options = {}
         options["compiler_config"] = CompilerConfig()
         parent_mesh, devices = DeviceManager.acquire_available_devices()
         options["device"] = devices[0]
         tt_model = torch.compile(model, backend=backend, options=options)
         ...
         DeviceManager.release_parent_device(parent, cleanup_sub_devices=True)
         ```

### Checklist

### DIFF
--- a/demos/resnet/resnet50_demo.py
+++ b/demos/resnet/resnet50_demo.py
@@ -10,6 +10,7 @@ from torchvision import transforms
 import torchvision.models as models
 import tabulate
 import requests
+from tt_torch.tools.device_manager import DeviceManager
 
 
 def main():
@@ -22,10 +23,12 @@ def main():
     cc.enable_consteval = True
     cc.consteval_parameters = True
 
-    # devices = tt_torch.query_devices()
-    tt_model = torch.compile(model, backend=backend, dynamic=False, options=cc)
+    devices = DeviceManager.get_available_devices()
+    options = {}
+    options["compiler_config"] = cc
+    options["device"] = devices[0]
+    tt_model = torch.compile(model, backend=backend, dynamic=False, options=options)
 
-    # tt_model(*inputs, devices[0])
     headers = ["Top 5 Predictions"]
     topk = 5
     prompt = 'Enter the path of the image (type "stop" to exit or hit enter to use a default image): '
@@ -55,6 +58,7 @@ def main():
         print()
 
         img_path = input(prompt)
+    DeviceManager.release_devices()
 
 
 if __name__ == "__main__":
--- a/demos/resnet/resnet50_demo.py
+++ b/demos/resnet/resnet50_demo.py
@@ -23,10 +23,8 @@ def main():
     cc.enable_consteval = True
     cc.consteval_parameters = True
 
-    devices = DeviceManager.get_available_devices()
     options = {}
     options["compiler_config"] = cc
-    options["device"] = devices[0]
     tt_model = torch.compile(model, backend=backend, dynamic=False, options=options)
 
     headers = ["Top 5 Predictions"]
@@ -58,7 +56,6 @@ def main():
         print()
 
         img_path = input(prompt)
-    DeviceManager.release_devices()
 
 
 if __name__ == "__main__":
--- a/demos/resnet/resnet50_demo.py
+++ b/demos/resnet/resnet50_demo.py
@@ -22,8 +22,10 @@ def main():
     cc.enable_consteval = True
     cc.consteval_parameters = True
 
+    # devices = tt_torch.query_devices()
     tt_model = torch.compile(model, backend=backend, dynamic=False, options=cc)
 
+    # tt_model(*inputs, devices[0])
     headers = ["Top 5 Predictions"]
     topk = 5
     prompt = 'Enter the path of the image (type "stop" to exit or hit enter to use a default image): '
--- a/tests/torch/test_basic.py
+++ b/tests/torch/test_basic.py
@@ -54,13 +54,10 @@ def test_add_multidevice():
             return torch.add(x, y)
 
     num_devices = DeviceManager.get_num_available_devices()
-    print("Number of available devices: ", num_devices)
     device_list = DeviceManager.get_available_devices()
     assert (
         len(device_list) == num_devices
     ), "Number of devices is not equal to expected."
-    print("Submeshes secured: ", device_list)
-    print("Parent mesh: ", DeviceManager._parent)
     threads = []
     compiler_configs = [CompilerConfig() for _ in range(num_devices)]
     for i in range(num_devices):
--- a/tests/torch/test_basic.py
+++ b/tests/torch/test_basic.py
@@ -5,10 +5,18 @@ import torch
 from torch import nn
 import pytest
 import math
+import threading
 
 import tt_torch
 from tt_torch.tools.verify import verify_module
 from tt_torch.tools.utils import CompilerConfig
+from tt_mlir import (
+    open_mesh_device,
+    close_mesh_device,
+    MeshDeviceOptions,
+    create_sub_mesh_device,
+    release_sub_mesh_device,
+)
 
 
 def test_abs():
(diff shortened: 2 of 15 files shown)


demos/resnet/resnet50_demo.py
```diff
@@ -22,7 +23,11 @@ def main():
     cc.enable_consteval = True
     cc.consteval_parameters = True
 
-    tt_model = torch.compile(model, backend=backend, dynamic=False, options=cc)
+    devices = DeviceManager.get_available_devices()
```
AleksKnezevic, line 26: Is this needed by default? If not, let's not introduce it into the demo. I'd rather we have a `resnet50_data_parallel_demo.py`

tt_torch/dynamo/backend.py
```diff
@@ -50,22 +51,30 @@ def dump_module(module, name, compiler_config):
 
 
 def _shlo_backend(
-    shlo, example_inputs, compiler_config, program=None, graph_constants=None
+    shlo,
+    example_inputs,
+    compiler_config,
+    program=None,
+    graph_constants=None,
+    device=None,
```
AleksKnezevic, line 59: Since the device object lives in `compiler_config` is there a reason to pass it around here as well?

AleksKnezevic: Never mind, reading comprehension fail on my part.

AleksKnezevic:
This looks awesome @hshahTT! Please add some N300 tests to CI that exercise this.

demos/resnet/resnet50_demo.py
hshahTT: Removed the DeviceManager stuff from this file and added a new `resnet50_data_parallel_demo.py` as requested. Confirmed that both demo files work on an n300.

tests/torch/test_basic.py
```diff
@@ -33,6 +35,51 @@ def forward(self, x, y):
...
+            target=verify_module_helper, args=(mod, input_shapes, cc, device)
+        )
+        threads.append(thread)
+
+    for thread in threads:
+        thread.start()
+    for thread in threads:
+        thread.join()
+
+    DeviceManager.release_devices()
```
jameszianxuTT, line 78: Q. Does this test work on a machine with no devices? 

We ask in our documentation that people run all this tests in test_basic to verify their installation.

tests/torch/test_basic.py
```diff
@@ -33,6 +35,51 @@ def forward(self, x, y):
     verify_module(Basic(), input_shapes=[(256, 256)] * 2)
 
 
+def verify_module_helper(module, input_shapes, compiler_config, device):
```
brataTT, line 38: this seems to directly call `verify_module`
does `threading` not let us use `verify_module` directly somehow?

tt_torch/tools/device_manager.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        # Secure a parent mesh device
+        self._parent = tt_mlir.open_mesh_device(
+            mesh_shape=mesh_shape,
+            options=options,
+        )
+        num_devices = mesh_shape[0] * mesh_shape[1]
+        for i in range(num_devices):
+            sub_device = tt_mlir.create_sub_mesh_device(self._parent, (1, 1), (0, i))
```
jazpurTT, line 90: can we make these mesh_shape=(1,1), mesh_offset=(0, i) for better readability?

hshahTT: Done

tt_torch/tools/device_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            if device in devices_copy:
+                tt_mlir.release_sub_mesh_device(device)
+                self._devices.remove(device)
+            else:
+                warnings.warn(
+                    f"Device {device} not found in the list of managed devices. Ignoring it."
+                )
+
+        if len(self._devices) == 0:
```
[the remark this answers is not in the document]
hshahTT: This is a fair point. My goal for the first pass of this API was to keep the internals abstracted away from the user so that if they just wanted to utilize all devices on the board they'd have a simple way of doing so.

I could definitely provide two "versions" of this API: one that keeps things simple for the user, and another more advanced one that allows the user full control of all internals (including parent mesh management).

tt_torch/tools/device_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # Secure a parent mesh device
+        self._parent = tt_mlir.open_mesh_device(
+            mesh_shape=mesh_shape,
+            options=options,
+        )
+        num_devices = mesh_shape[0] * mesh_shape[1]
+        for i in range(num_devices):
+            sub_device = tt_mlir.create_sub_mesh_device(
+                self._parent, mesh_shape=(1, 1), mesh_offset=(0, i)
```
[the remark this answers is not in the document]
hshahTT: Good catch on the offset. I'll fix that.

Re: multiple parents, I think we as team will need to agree on the level of abstraction we want to keep from the user. For now I kept it simple with high abstraction, but a better way might be to allow advanced users more control while keeping the existing abstractions for users who just want a quick way to access multiple devices.

tests/torch/test_basic.py
```diff
@@ -33,6 +35,48 @@ def forward(self, x, y):
...
+        )
+        threads.append(thread)
+
+    for thread in threads:
+        thread.start()
+    for thread in threads:
+        thread.join()
+
+    DeviceManager.release_devices()
+    assert DeviceManager._parent is None, "Parent device is not released."
```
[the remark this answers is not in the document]
hshahTT: Users can call get_available_devices() if they want to see the existing acquired devices, so that acts as a getter for `_devices` already.

I didn't add a getter for _parent because in the current API setup I intentionally want to hide the parent mesh away from the end user. 

The way the device management API is currently set up, users should never need to directly know what the parent mesh is (since it'll be managed automatically by the DeviceManager class) and will have no need to call a parent getter in normal use cases.

The reason for these assertions here is just to help verify that the user's dev environment is set up correctly (i.e., there's nothing funky going on with the chips that could cause issues when the user uses DeviceManager). 

This is all predicated on whether we decide to keep the API simple for now, or if we allow users access to parent meshes as well. If we go for the latter then I definitely agree that a parent getter will be needed.

tt_torch/tools/device_manager.py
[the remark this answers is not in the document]
jnie-TT: Agreed with the simple/advanced interfaces. I guess multiple parents is a bit of a stretch, but yeah it would be nice to be able to have more control and flexibility when desired.

tests/torch/test_basic.py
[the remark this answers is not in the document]
hshahTT: Added some changes that should address this.

hshahTT:
Refactored the DeviceManager class to provide more functions that directly expose parent and sub mesh creation to the user.

AleksKnezevic (approved):

jnie-TT (approved):

PR #573, opened 2025-04-07
Replace PCC/ATOL invalid numeric values with -1

jameszianxuTT:
### Ticket
None

### Problem description
YoloV5 results are not being ingested by the database because the tags it generates are not parsable. This is because it generates a numeric infinite ATOL. 

Both the JSON and postgres spec "officially" do not support NaN or infinity values and those should be stripped from the reports.

### What's changed
Replace infinity and NaN values with -1 in PCC and ATOL. 

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -318,6 +318,22 @@ class ModelTester:
         else:
             raise ValueError(f"Current mode is not supported: {self.mode}")
 
+    def filter_nan_inf_for_record(self, metric_key):
+        metric_list = self.record_tag_cache.get(metric_key, [])
+        if metric_list:
+            metric_list = [
+                -1
+                if (isinstance(x, float) and x != x)
+                else -1  # NaN case
+                if (isinstance(x, float) and x == float("inf"))
+                else -1
+                if (isinstance(x, float) and x == -float("inf"))
+                else x
+                for x in metric_list
+                if isinstance(x, (int, float)) or isinstance(x, str)
+            ]
+        self.record_tag_cache[metric_key] = metric_list
+
     def record_aggregate_model_metric(self, metric_key, default_value=-1):
         # read a metric from the tag cache and write out the average and min values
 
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
From the JSON RFC:

> [2.4](https://www.rfc-editor.org/rfc/rfc4627#section-2.4).  Numbers

>  The representation of numbers is similar to that used in most
   programming languages.  A number contains an integer component that
   may be prefixed with an optional minus sign, which may be followed by
   a fraction part and/or an exponent part.

 > Octal and hex forms are not allowed.  Leading zeros are not allowed.

 > A fraction part is a decimal point followed by one or more digits.

>  An exponent part begins with the letter E in upper or lowercase,
   which may be followed by a plus or minus sign.  The E and optional
   sign are followed by one or more digits.

 > Numeric values that cannot be represented as sequences of digits
   (such as Infinity and NaN) are not permitted.


And from the [postgres spec for the numeric type](https://www.postgresql.org/docs/current/datatype-json.html):
> number | numeric | NaN and infinity values are disallowed

jameszianxuTT:
Also, I know that -1 is technically a valid PCC but not one I think we have a usecase for. 

I think we also do not have a usecase to store NaN and +/-Infinity values. If this is actually the case, we will have to make changes to the parser and all queries that handle PCC and ATOL as these cases need special handling. It is possible in theory but I think not necessary.

jameszianxuTT:
Verification: Check if the data collection for YoloV5 in this test run gets reported: https://github.com/tenstorrent/tt-torch/actions/runs/14316054513 and collected in the database.

AleksKnezevic (approved):

jameszianxuTT:
image

As final verification for this change, we see that an entry for YOLOv5 is now recorded in the full model execute nightlies with 0.016 PCC and depth EXECUTE.

PR #579, opened 2025-04-08
Add benchmark result upload path

jameszianxuTT:
### Ticket
None

### Problem description
The previous idea to self-consume the previous benchmark report artifact requires more stability in CI than we currently have, and adds fragility to the pipeline if the benchmark job fails for any unrecoverable reason in a week. 

### What's changed
Create a workflow that directly uploads an artifact from any branch, effectively using Github Artifacts as a remote filestore for a test durations file

### Checklist


results/download_artifacts.py
```diff
@@ -131,12 +131,17 @@ def process_zip_file(zip_path, folder_name):
         print(f"Failed to process zip file '{zip_path}': {e}")
 
 
-def get_run_details(args, headers, workflow_name):
+def get_run_details(args, headers, workflow_name, run_index_n=0):
     """
     Retrieves run details either from a specified run-id or by fetching the latest run.
     Returns a tuple of (run_id, date_str, folder_name).
+
+    lookback
```
kmabeeTT, line 139: is this supposed to be bigger comment/description? I was confused at first until I looked further below.

kmabeeTT (approved):

Issue #335, opened 2025-02-20
yolo v10


AleksKnezevic:
Add yolov10 model from ultralytics. Test should run in op-by-op nightly pipeline.

ddilbazTT:
This should work: [https://github.com/THU-MIG/yolov10](https://github.com/THU-MIG/yolov10 )

ddilbazTT:
There is also an onnx version of the model available in the official repository: https://github.com/THU-MIG/yolov10/releases/tag/v1.1

I'll still continue with the pytorch codebase. One issue I have had is that, there is something off with the way their model is created. It is already an issue that is raised: https://github.com/ultralytics/ultralytics/issues/9112

Basically, model.eval() starts training, and it's because the model inherits from nn.Module object. So the following code snippet causes an error:

```
    def get_framework_model(self):
        model = (
            self.framework_model.eval()
            if hasattr(self.framework_model, "eval")
            else self.framework_model
        )
        return model
```

For now, I'm just running the model with `self.framework_model`. 
I was also able to remove py-cpuinfo. So we don't need to install that anymore.

kmabeeTT:
Great job merging yolov10 from ultralytics to op by op flow, okay close this now?

ddilbazTT:
Yep it's running! So closing it now

PR #588, opened 2025-04-09
Add YOLOv10 #335

ddilbazTT:
### Ticket
#335 

### Problem description
Added YOLOv10 using Ultralytics library and *.pt file from [https://github.com/THU-MIG/yolov10](https://github.com/THU-MIG/yolov10)

### Checklist
- [X] Added `tests/models/yolov10/test_yolov10.py`
- [X] Added yolov10 to nightly test as end-to-end test is not passing
- [X] Added `ultralytics` dependency in `requirements.txt`

### DIFF
--- a/tests/models/yolov10/test_yolov10.py
+++ b/tests/models/yolov10/test_yolov10.py
@@ -1,0 +1,95 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import cv2
+import numpy as np
+import torch
+import pytest
+import os
+from pathlib import Path
+import requests
+from tests.utils import ModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+from ultralytics import YOLO
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        # Reference: https://github.com/THU-MIG/yolov10
+        url = "https://github.com/THU-MIG/yolov10/releases/download/v1.1/yolov10n.pt"
+
+        if (
+            "DOCKER_CACHE_ROOT" in os.environ
+            and Path(os.environ["DOCKER_CACHE_ROOT"]).exists()
+        ):
+            download_dir = Path(os.environ["DOCKER_CACHE_ROOT"]) / "yolov10_weights"
+        else:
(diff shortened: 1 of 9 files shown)


tests/models/yolov10/test_yolov10.py
```diff
@@ -0,0 +1,81 @@
...
+from tests.utils import ModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+from ultralytics import YOLO
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        url = "https://github.com/THU-MIG/yolov10/releases/download/v1.1/yolov10n.pt"
+
+        download_dir = Path.home() / ".cache/yolov10_weights"
```
AleksKnezevic, line 20: Can you please use a ENV_VAR for this, which should be mapped to `/mnt/dockercache/` in CI. See https://github.com/tenstorrent/tt-torch/blob/2544dd69a327980d994993ae71259a0790de7aa9/.github/workflows/run-op-by-op-model-tests-nightly.yml#L231

ddilbazTT: I assume I need to:
- create new env variable that is used in the workflow files. i.e. DOCKER_CACHE_ROOT=/mnt/dockercache
- reference the env variable in which ever tests that use Path.home() (I noticed yolov3 also does this so I will change that too.)

ddilbazTT: This is how I intend to change it:
```
 if "DOCKER_CACHE_ROOT" in os.environ and Path(os.environ["DOCKER_CACHE_ROOT"]).exists(): 
    download_dir = Path(os.environ["DOCKER_CACHE_ROOT"]) / "yolov10_weights" 
else: 
    download_dir = Path.home() / ".cache/yolov10_weights"
```
Should I also check if I have permissions to write there?

ddilbazTT:
Pushed a commit addressing PR comments, would appreciate your reviews again @AleksKnezevic @jazpurTT

AleksKnezevic (approved):

tests/models/yolov10/test_yolov10.py
AleksKnezevic: no need to check permission. btw, when you mark a conversation as resolved, I can't see it when I just go to the PR, so you usually won't get a response 😄

jazpurTT (approved):

PR #600, opened 2025-04-10
Fix incorrect reporting of Full Execute isolated pytest success

jameszianxuTT:
### Ticket
None

### Problem description
Full model execution tests were converted to run in isolated pytest processes in #562. They have not been correctly reporting job failure status due to "failure" variable subshell variable scope.

### What's changed
Avoid running bash loop in a subshell so that the "failure" variable is set in the global context.

### Checklist
- [x] Ran tests with simulated failures, with and without this change, to verify it fixes the issue

### DIFF
--- a/.github/workflows/run-full-model-execution-tests-nightly.yml
+++ b/.github/workflows/run-full-model-execution-tests-nightly.yml
@@ -184,7 +184,7 @@ jobs:
         failure=0
 
         # Split the test string into individual tests and run them one by one
-        echo "${{ matrix.build.tests }}" | tr ' ' '\n' | while read -r test; do
+        while read -r test; do
           if [[ -n "$test" ]]; then
             echo "Running test: $test"
             report_file="${{ steps.strings.outputs.test_report_dir }}/$(echo $test | tr '/' '_' | tr ':' '_')__${{ steps.strings.outputs.test_report_path_models }}"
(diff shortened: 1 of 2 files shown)


kmabeeTT (approved):
Good catch! Do we think there were any other failures hiding recently here (fingers crossed no).

jameszianxuTT:
> Good catch! Do we think there were any other failures hiding recently here (fingers crossed no).

We would have seen them in the dashboard, I've been monitoring and we only saw YoloV5 fail silently due to a transient dependency shift.

In general this only affects full model execute / full model execute nightlies, which have been historically stable anyways. If there do turn out to be missed regressions, then we'll see them in today's nightlies (and possibly in this OnPR check).

jameszianxuTT:
Going to trig nightly full model execute as a test to make sure there aren't any other issues: https://github.com/tenstorrent/tt-torch/actions/runs/14390472337

kmabeeTT:
> Going to trig nightly full model execute as a test to make sure there aren't any other issues: https://github.com/tenstorrent/tt-torch/actions/runs/14390472337

Previous comment makes sense, thx I agree.  This run was good, just one fail with not-a-real-issue:

> 
> FAILED tests/models/detr/test_detr.py::test_detr[full-eval] - urllib.error.HTTPError: HTTP Error 403: rate limit exceeded
> ============================== 1 failed in 2.98s ===============================

jameszianxuTT:
> Previous comment makes sense, thx I agree. This run was good, just one fail with not-a-real-issue:

Nice! The fact that the job shows as failing when this one test fails is good evidence that this change works - we would've seen all success before :)

jameszianxuTT:
> This run was good, just one fail with not-a-real-issue:
Also opened an issue for these 403 `torch.hub.load` fails #607 and a quick PR #608

PR #603, opened 2025-04-11
Downgrade to Python 3.10

jameszianxuTT:
### Ticket
None

### Problem description
We are using python 3.11 and that does not align with other tt-forge frontends.

### What's changed
Downgrade project to python 3.10.

### Checklist
- [x] Swept on PR and subset of nightlies to ensure no obvious breaks or regressions

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -11,8 +11,8 @@ RUN apt-get update && apt-get install -y \
     python3-dev \
     python3-venv \
     python3-pip \
-    python3.11-dev \
-    python3.11-venv \
+    python3.10-dev \
+    python3.10-venv \
     git \
     git-lfs \
     libhwloc-dev \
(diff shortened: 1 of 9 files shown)


jameszianxuTT:
At this point, we have 3 things that _could_ break or see perf regressions:
1. Nightlies (run at midnight)
2. Weekly op by op (run after nightlies, Saturday)
3. Benchmark (run after nightlies, Sunday)

If onPR passes, we'll just merge this in and assess the state of (1) and (2) around midday Saturday. Should those fail, we'll revert and assess to avoid breaking benchmarks.

AleksKnezevic:
@jazpurTT, will the torchvision wheel in the docker be built automatically, or do we need to somehow kick it off prior to merging this?

jameszianxuTT:
I believe it is already rebuilt - you can see this [docker build](https://github.com/tenstorrent/tt-torch/actions/runs/14406363298/job/40403788667?pr=603#step:8:17191) did rebuild the entire docker image and it should now be cached (with cache key being a hash of the dockerfile).

jazpurTT:
By updating the DockerFile it should automatically trigger the docker to build the new image with the new torchvision wheel build and cache it

AleksKnezevic:
Awesome, that makes sense. Thanks!

jazpurTT (approved):
let people know that this is getting merged to make sure there is no conflict with the torchvision wheel that is currently built under their local dist/ dir

mmanzoorTT (approved):

PR #604, opened 2025-04-11
Improve Benchmark CI (Auto load balancer, out of band matrix autogeneration, test runner isolation)

jameszianxuTT:
### Ticket
None

### Problem description
Benchmarks are unreliable due to:
- The possibility of timeout, and the impossibility of setting a timeout duration because we cannot know the a priori test runtime. Using previous runtime will produce false negatives because runtime may rightly drift week over week, and we'd have to requeue jobs into a new workflow to avoid a global timeout to disambiguate timing out tests vs. elongated tests. 
- The possibility of device hang causing all tests beyond hanging point in a group to fail even if we had a timeout-terminator without a device reset mechanism
- The possibility of multiple timeout/hang causing a runner bottleneck for remaining tests

### What's changed
- Isolate benchmark tests in their own runner job, so there is a clean state before running each test. This will lead to problems with job count, as we have a hard limit of 250, but haven't hit it yet
- Strip already EXECUTing tests from benchmarks as they are fully redundant and are unpromotable anyways
- Add extra testname sanitation level to avoid artifact upload failure as seen in previous runs
- Always run report generation even if previous steps fail for an unforseen reason

### Checklist
- [x] Ran dummy benchmarks to check functionality


.github/workflows/run-depth-benchmark-tests.yml
```diff
@@ -19,35 +19,17 @@ on:
...
-          {runs-on: 'wormhole_b0', name: 'bmk_14_qtn_test_musicgen_small[full-eval]', tests: ['tests/models/musicgen_small/test_musicgen_small.py::test_musicgen_small[full-eval]']},
-          {runs-on: 'wormhole_b0', name: 'bmk_15_qtn_test_openpose[full-eval]', tests: ['tests/models/openpose/test_openpose.py::test_openpose[full-eval]']},
-          {runs-on: 'wormhole_b0', name: 'bmk_16_qtn_test_segment_anything[full-eval]', tests: ['tests/models/segment_anything/test_segment_anything.py::test_segment_anything[full-eval]']},
-          {runs-on: 'wormhole_b0', name: 'bmk_17_qtn_test_speecht5_tts[full-eval]', tests: ['tests/models/speecht5_tts/test_speecht5_tts.py::test_speecht5_tts[full-eval]']},
-          {runs-on: 'wormhole_b0', name: 'bmk_18_qtn_test_vit[full-eval]', tests: ['tests/models/vit/test_vit.py::test_vit[full-eval]']},
-          {runs-on: 'wormhole_b0', name: 'bmk_19_qtn_test_yolov5[full-eval]', tests: ['tests/models/yolov5/test_yolov5.py::test_yolov5[full-eval]']},
-        ]
+        group: ${{ fromJson(inputs.matrix-json-splits) }}
```
jameszianxuTT, line 26: Replacing the above huge test matrix with this generated matrix was the goal for benchmark autopromotion and it seems to work.

tt_torch/tools/benchmark_promotion.py
```diff
@@ -190,59 +212,75 @@ def parse_benchmark_results_xlsx(file_path):
...
 
 
 def generate_dynamic_benchmark_test_matrix():
 
     output_file = "benchmark_test_matrix.json"  # hardcoded into CI
+    output_file_splits = "benchmark_test_matrix_splits.json"  # hardcoded into CI
     report_dir = "benchmark_report"
-
-    # download previous report from run of depth-benchmarks workflow on main
+    report_branch = "benchmark_report_fs"
```
jameszianxuTT, line 266: Created a parallel branch in tt-torch that will be used as the source of truth for "what ran last time". We are using github artifacts from a separate branch as a filestore, and this way we don't need to make PRs to update our known test durations.

jameszianxuTT:
Benchmark partial run generates a valid xlsx report at the end which confirms this doesn't break, immediately - https://github.com/tenstorrent/tt-torch/actions/runs/14405353977/job/40402720845

tt_torch/tools/benchmark_promotion.py
```diff
@@ -190,59 +212,75 @@ def parse_benchmark_results_xlsx(file_path):
...
-    #     formatted_matrix += f"  {{runs-on: '{job['runs-on']}', name: '{job['name']}', tests: {job['tests']}}},"
-    # formatted_matrix += "]"
-    build = [matrix]
+def get_existing_execute_tests():
+    # We don't care about tests that already run in execute.
+    # Benchmarks tell us:
+    #   if a model regressed (eg. expect to run in execute, but now run in compile)
+    #   if a model is promotable (eg. expect to run in compile, runs in execute)
+    # Regression testing here is redundant to full model exec tests in nightly
+    # Models cannot be promoted beyond execute so we don't care about promotability either
```
kmabeeTT, line 247: Nice, agreed.

tt_torch/tools/benchmark_promotion.py
kmabeeTT: Nice!

kmabeeTT (approved):
Lots of cool changes here, overall looks good, no real concerns - let's see how it works!

Issue #607, opened 2025-04-11
Nondeterministic Rate Limiting HTTP 403 error when using torch.hub.load


jameszianxuTT:
This has been observed 3 times in the past day, for example [breaking UNET op by op nightly tests](https://github.com/tenstorrent/tt-torch/actions/runs/14395348326/job/40411874840). 

It seems nondeterministic and has the following error signature:
```
tests/models/unet/test_unet.py:67: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/utils.py:42: in __init__
    self.framework_model = self._load_model()
tests/models/unet/test_unet.py:18: in _load_model
    model = torch.hub.load(
env/venv/lib/python3.11/site-packages/torch/hub.py:638: in load
    repo_or_dir = _get_cache_or_reload(
env/venv/lib/python3.11/site-packages/torch/hub.py:258: in _get_cache_or_reload
    _validate_not_a_forked_repo(repo_owner, repo_name, ref)
env/venv/lib/python3.11/site-packages/torch/hub.py:203: in _validate_not_a_forked_repo
    response = json.loads(_read_url(Request(url, headers=headers)))
env/venv/lib/python3.11/site-packages/torch/hub.py:185: in _read_url
    with urlopen(url) as r:
/usr/lib/python3.11/urllib/request.py:216: in urlopen
    return opener.open(url, data, timeout)
/usr/lib/python3.11/urllib/request.py:525: in open
    response = meth(req, response)
/usr/lib/python3.11/urllib/request.py:634: in http_response
    response = self.parent.error(
/usr/lib/python3.11/urllib/request.py:563: in error
    return self._call_chain(*args)
/usr/lib/python3.11/urllib/request.py:496: in _call_chain
    result = func(*args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <urllib.request.HTTPDefaultErrorHandler object at 0x7ff7eefff410>
req = <urllib.request.Request object at 0x7ff7e3fd6290>
fp = <http.client.HTTPResponse object at 0x7ff7da8e4130>, code = 403
msg = 'rate limit exceeded'
[...]

Issue #595, opened 2025-04-10
Setup cached access to S3 LFS


AleksKnezevic:
We'll be setting up an S3 bucket to hold all of our onnx models and other large files. We don't want to be pulling the models from there on every CI run as this is slow and expensive, so we want to cache them locally.

AleksKnezevic:
For the caching, we'll want to do something similar to [this](https://github.com/tenstorrent/tt-torch/blob/7780f0c252c196e5d32cb0ec94d847766bb3e515/tests/models/yolov10/test_yolov10.py#L19). Please extract into a separate function so that the code is reusable. We can use a github secret to store the login credentials

PR #608, opened 2025-04-11
Prevent torch.hub.load ND rate limiting errors

jameszianxuTT:
### Ticket
#607 

### Problem description
Random torchhub 403 rate limiting errors have been showing up in CI for the past few days.

### What's changed
Add as seen in yolov5 test, add torch.hub._validate_not_a_forked_repo = lambda a, b, c: True to overload a check that's consuming the rate limit.

### Checklist

### DIFF
--- a/tests/models/detr/test_detr.py
+++ b/tests/models/detr/test_detr.py
@@ -17,6 +17,7 @@ class ThisTester(ModelTester):
         The model is from https://github.com/facebookresearch/detr
         """
         # Model
+        torch.hub._validate_not_a_forked_repo = lambda a, b, c: True
         model = torch.hub.load(
             "facebookresearch/detr:main", "detr_resnet50", pretrained=True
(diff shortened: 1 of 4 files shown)


jameszianxuTT:
Assessing in a [test run of the 4 modified tests](https://github.com/tenstorrent/tt-torch/actions/runs/14409759108) to ensure I didn't break something and that they all pass constructor.

jameszianxuTT:
> Can we download these to our cache, and see if they're there first, before downloading from torch.hub? 

This is introducing ND crashes in our test coverage now, and I don't think the changes in #595 will be ready today. The logic here should be updated to use the model cache once we have that set up, but can we merge this as a temporary fix in the meantime?

AleksKnezevic:
Sure, please add todo to remove once #611 is in.

AleksKnezevic (approved):

PR #615, opened 2025-04-14
Added missing tests for atan2, reverse, bitwise and/or/not/xor ops

kmitrovicTT:
All of them were successfully added some time ago in tt-mlir, some were blocked by issues, now that they all work I am adding torch tests.


ddilbazTT (approved):
You need to fix the pre-commit issues but otherwise LGTM

Issue #260, opened 2025-02-03
Push model PCC issues


AleksKnezevic:
Many of the models run in push pipelines have bad PCC, some have good PCC but checking isn't enabled. Please go through them and enable PCC checking, and debug where the failures are.

AleksKnezevic:
In order to do this, we'll need to integrate intermediates into tt-mlir runtime. @LPanosTT knows what needs to be done here, so please sync with him.

jameszianxuTT:
I have implemented a prototype version of this, and I have swept the tests we run in full model execute. The sweep provides these results:
```
Test Summary:
Can generate runtime intermediates: 31
Fail to generate runtime intermediates: 47
Test segfaults: 1 (stable diffusion unet)
```

Of the models that fail to produce runtime intermediate results, we expose a number of quirks that result from deviation between our stack and Pytorch CPU.

Most of these failures are due to some sort of mismatch between the `golden` tensor and the runtime tensor. For example:
1. op `view` leads to a shape mismatch `E           AssertionError: Shape mismatch on output 0: torch.Size([16, 16]) vs torch.Size([256, 16, 16])`
2. op `unsqueeze` leads to a type mismatch `E       RuntimeError: Promotion for uint16, uint32, uint64 types is not supported, attempted to promote UInt32 and Int` between `tensor = tensor([[[1, 1, 1, 1, 1, 1, 1, 1, 1]]], dtype=torch.int32)
golden_tensor = tensor([[[1, 1, 1, 1, 1, 1, 1, 1, 1]]], dtype=torch.uint32)`
3. op `max_pool2d_with_indices` has several issues - the golden is a tuple of two tensors, but we have no way of retrieving more than one tensor per op from the runtime. In this case, we can discard the second tensor, as it is unused auxiliary data, but this is not going to be true in general.  It also meets a shape mismatch `E           AssertionError: Shape mismatch on output 0: torch.Size([1, 64, 56, 56])   (golden) vs torch.Size([1, 1, 3136, 64]) (actual)`
4. op `convert_element_type` has a shape mismatch `E           AssertionError: Shape mismatch on output 0: torch.Size([1, 10]) vs torch.Size([1, 10, 1, 1])`

There are probably more. I haven't done a through survey of all the failure modes I've seen, and the code early exits so there may be new failures beyond these points.

---

[...]

PR #622, opened 2025-04-15
Add runtime intermediate capture to debug executing model PCC issues

jameszianxuTT:
### Ticket
#260 

### Problem description
We don't currently have a way to check intermediate outputs from nodes within the torchfx graph, from a model running in full execute. As a result, for models that execute e2e, we don't have a way to validate at which op, in runtime, things are going wrong. 

### What's changed
- Add runtime intermediate cache to store a map of (torchFxNodeName, runtimeIntermediateTensor, CPUGoldenTensor)
- Add report generation to stdout, dumping per-node PCC/ATOL values in a table. 
  - Due to TTNN reshaping and other slight discrepancy between runtime intermediate shape and ground truth shape, also show PCC/ATOL values between flattened items
- Add bindings to collect tensors from hardware using tt-mlir runtime post-op callback and convert to torch Tensor
- Add function to process the graphmodule right before it's passed to runtime, and trace the graph to generate a golden cache from CPU
- Create XLSX report generator and CI infra improvements so this can be run in CI with testlist automation, useful to sweep all models in EXECUTE and summarize runtime intermediate results by parsing stdout and adding to XLSX with color highlights

### Checklist
- Models we expect to have good PCC are shown to have a final intermediate output with good PCC
- Sweep of all e2e EXECUTE tests shows a few errors and discrepancies for large models, but almost all produce well formatted and plausible reports


jameszianxuTT:
Example of the [nightly pipeline with split build + tools tests](https://github.com/tenstorrent/tt-torch/actions/runs/14495772694)

image

jameszianxuTT:
TODO before this goes in:
- QOL changes requested by Kyle
  - Adjust PCC highlighting thresholds
  - Fixed or autosize column widths in excel report
  - Print final output 
  - Print first failing point w/ adjustable "failing PCC threshold"
+Future QOL change: Compile and extract ttir fragment from the node to facilitate debugging

tt_torch/dynamo/executor.py
```diff
@@ -222,6 +223,61 @@ def _cleanup_resources(self, preprocessed_activations, device):
...
+                if isinstance(golden, torch.Tensor) and golden.dim() == 0:
+                    print(f"Unsqueezing golden {golden} to {golden.unsqueeze(0)}")
+                    golden = golden.unsqueeze(0)
+
+                # some ops return a tuple of tensors as output (e.g. max_pool_2d_with_indices)
+                # we expect to only use the first, though this may be changed in the future
+                elif isinstance(golden, (tuple, list)) and len(golden) > 1:
+                    golden = golden[0]
+                    print(
+                        f"[WARNING] {node.name} has {len(golden)} outputs, but we can only get one from runtime."
```
AleksKnezevic, line 273: Can we make this orange so it's easier to spot?

tt_torch/tools/verify.py
```diff
@@ -141,25 +142,18 @@ def verify_against_golden(
...
             assert False, err_msg
     elif assert_pcc and not assert_atol:
         print("Ignoring ATOL check\n")
-        if passed_pcc:
-            print(msg)
-        else:
+        if not passed_pcc:
             assert False, err_msg
-    else:
+    if not disable_print:
```
AleksKnezevic, line 155: Just so I understand the code change, when verifying intermediates we will print pcc/atol of each intermediate result?

AleksKnezevic: That is the behaviour that makes sense to me, at least.

AleksKnezevic (approved):
Left a couple of questions/nits, but otherwise looks good.

tt_torch/tools/verify.py
jameszianxuTT: No, the purpose of this change is to suppress the printouts that are normally generated in this function. Those exist only as a convenience side effect feature from this function.

I just use this function to perform the PCC/ATOL comparison, and cache the numerical result. I then cleanly print out the results in a table with a format of my choosing

AleksKnezevic: ok, perfect, as long as the results are clearly visible at time of running without needing to open other docs, I'm happy.

jameszianxuTT:
First failing op is annotated inline in the xlsx now with a "comment" as well as bold format.

image

PR #625, opened 2025-04-17
Update parse_op_by_op_results.py compile failure handling and stablehlo-to-ttir-pipeline options

kmabeeTT:
### Ticket
None

### What's changed
 - Remove deprecated --stablehlo-to-ttir-pipeline=enable-remove-dead-values=true argument falsely being reported as an error in compile-depth-4 fails, replace with correct newer args.
 - In the case where parse_op_by_op_results.py run of ttmlir-opt (to extract failing signature) passes, record a message rather than nothing at all so user knows it passed on rerun.  
 - Improve stderr parsing to better catch errors and crashes, instead of common warning that appears on first line

### Checklist
- [x] Tested on nightly regression .json files and locally generated .json files, works .

### DIFF
--- a/results/parse_op_by_op_results.py
+++ b/results/parse_op_by_op_results.py
@@ -545,6 +545,28 @@ def generate_all_ops_worksheet(worksheet, bold, all_ops, not_executing_only=Fals
     # worksheet.autofit()
 
 
+# Parse error output from stderr
+def parse_error_output(stderr):
+    stderr_lines = stderr.strip().splitlines()
+
+    # Find the first real error line (MLIR format)
+    error = next((line for line in stderr_lines if ": error:" in line), None)
+
+    # If no error line found, look for crash indicators
+    if not error:
+        crash_keywords = ["Assertion `", "PLEASE submit a bug report", "Stack dump:"]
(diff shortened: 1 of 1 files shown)


results/parse_op_by_op_results.py
```diff
@@ -734,6 +734,10 @@ def generate_op_reports_xlsx():
                     if result.returncode != 0:
                         error = result.stderr.split("\n")[0]
                         trace_dump = result.stderr
+                    else:
```
AleksKnezevic, line 737: do you want this here to clutter the spreadsheet? Completely up to you, as you're the one consuming it :)

AleksKnezevic (approved):

results/parse_op_by_op_results.py
kmabeeTT: Yeah, I like it for now - I am often printing all error messages and grepping through them, and it's useful to see something here instead of nothing, in these cases where something flaky happened (probably not a legit failure), so I can see how many there were and exclude them.

kmabeeTT:
screenshot of better handling/capture of errrors.
image

Issue #580, opened 2025-04-08
Create system desc each time it's compiled


brataTT:
# Current state
Currently, system desc is created once every time device is opened. And the contents are copied to a temporary file before being passed into the compiler.

# Problem
Since the function to system desc opens a device for itself, it cannot be used while a device is open. Once passed in, the contents of the file are changed, so it needs to be re-written every time.

# Solution
The function to create system desc in tt-mlir needs to be updated to accept a device being passed in.
That way, a new system desc can be created every time one needs to be passed into the compiler.
Creating a temporary file can also be avoided if a new desc is created each time.

PR #628, opened 2025-04-17
Get system descriptor for opened device

brataTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/580

### Problem description
We couldn't create a system descriptor from tt-mlir while having the device open before.
But now that the function `getCurrentSystemDesc` allows opened device to be passed in, we can use it.
This will be useful for multidevice to make sure we pass system descriptor of the device being used for that workload.

### What's changed
- Create a new system desc for every compilation.
- Pass in device being used to get system desc for that device.

### Checklist

### DIFF
--- a/tt_torch/csrc/bindings.cpp
+++ b/tt_torch/csrc/bindings.cpp
@@ -260,18 +260,6 @@ std::vector<at::Tensor> run_end_to_end(std::vector<at::Tensor> &inputs,
   return outputs;
 }
 
-tt::runtime::Device
-open_mesh_device(const std::vector<uint32_t> &mesh_shape,
-                 const tt::runtime::MeshDeviceOptions &options) {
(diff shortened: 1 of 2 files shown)


ddilbazTT (approved):

ddilbazTT:
nit: is the checklist things that are done or will be done in the future?

brataTT:
> nit: is the checklist things that are done or will be done in the future?

Thanks for the review.

Skipping the first task since there's nothing to compare against.
Doing a schema check seems unnecessary.

Second task is done 👌 The current implementation creates the system_desc when it is used.
Instead of leaving a delay between usage and creation.
So, this segfault will not occur again.
To get a proper error message instead of segfault, we'll need to add an assert from metal side to make sure an opened device is not opened again. 
So, I'm not adding any extra logic here.

PR #630, opened 2025-04-17
Restore Model Caches at HF_HOME and TORCH_HOME

jameszianxuTT:
### Ticket
None

### Problem description
Fix bug causing HF_HOME and TORCH_HOME to resolve to paths inside the container due to incorrect reference of environment variable. This caused all model caches to be misdirected to an empty in-container path at the start of job execution since April 9, explaining the increased occurrence of CI failures due to network failures and rate limits (which would have been avoided if a cache hit was possible). 

### What's changed
Move DOCKER_CACHE_ROOT env variable to workflow-global scope. It is still accessible as an env variable from the any shell, run in a step or job within the workflow.

### Checklist

### DIFF
--- a/.github/workflows/run-depth-benchmark-tests.yml
+++ b/.github/workflows/run-depth-benchmark-tests.yml
@@ -16,6 +16,9 @@ on:
     workflows: [Build] # backref to run-build as dependency
     types: [completed]
 
+env:
+  DOCKER_CACHE_ROOT: /mnt/dockercache
+
 jobs:
   tests:
     timeout-minutes: 500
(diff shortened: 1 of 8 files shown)


jameszianxuTT:
Verification:
1. [Control experiment](https://github.com/tenstorrent/tt-torch/actions/runs/14521816275/job/40745161151#step:10:66) shows that TORCH_HOME is resolving to `/torch` which is incorrect and a container-scope path
2. [Fix experiment](https://github.com/tenstorrent/tt-torch/actions/runs/14522810331/job/40748947360#step:10:66) shows that TORCH_HOME is resolving to /mnt/dockercache/torch which is mounted outside the container on host
3. [Reference experiment](https://github.com/tenstorrent/tt-torch/actions/runs/14523598683/job/40750019577#step:4:7) shows that workflow-scoped `env` variables are accessible from within lower scoped contexts like job and step.  Docs [corroborate this fact](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow)

AleksKnezevic (approved):
Great catch!

PR #638, opened 2025-04-21
Const cache bugfix

vwellsTT:
### Problem description
I found some issues with the test_distillbert_multiloop test when preparing to merge my `tt-mlir` const-eval changes.  I think these are responsible.

### What's changed
Constants come before activations in the test executor, but when constants were cached, inputs would be assigned layouts as if they were first.  Now we pass in a start index which offsets any cached constants as needed.

### Checklist


tt_torch/csrc/bindings.cpp
```diff
@@ -204,24 +205,22 @@ preprocess_inputs(tt::runtime::Device device, std::vector<at::Tensor> &inputs,
...
-                   tt::runtime::Layout layout = tt::runtime::getLayout(
-                       binary, program_idx, rt_inputs_with_layout.size());
 
-                   tt::runtime::Tensor tensor =
-                       tt::runtime::toLayout(t, device, layout);
-
-                   tt::runtime::setTensorRetain(tensor, /*retain=*/true);
-                   return tensor;
-                 });
+  // Use a for loop instead of std::transform to properly handle indices
```
jnie-TT, line 209: nit: I think we can remove this comment.

jnie-TT (approved):
Awesome catch, looks great!

Issue #637, opened 2025-04-21
Tracking ticket for various runtime/execute asserts in tt-torch models


kmabeeTT:
Taken from Sunday April 20th Nightly Regression : https://github.com/tenstorrent/tt-torch/actions/runs/14555841211

**Quick Pulse**

In general 98.8% of unique ops (12475 of 12621) can pass EXECUTE (7) depth in op by op reports, so we're in pretty good shape, but remaining signatures will unlock few more models (we can currently run 95 of ~133 models we're testing).

**Summary:**

- Ignoring 8 Timeout and 3x "Error message not extracted" for now
- 39 unique ops hitting asserts (ignoring 8x timeouts, 3x "Error message cannot be extracted" for now)
- 7 buckets total (2 bigger ones, 5 smaller ones)

### Bucket 1 - 25 ops - Convolutions 

opened tt-metal ticket while back for these and earlier conv2d signatures and ported to TTNN Conv2D sweep infra : https://github.com/tenstorrent/tt-metal/issues/19762

```
Model Name: Stable_Diffusion NumOps: 184683.0 CompileDepth: 6 Torch Name: aten::convolution Error: Error: TT_THROW @ /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/program/program.cpp:791: tt::exception info: Statically allocated circular buffers in program <num> clash with L<num> buffers on core range [(x=<num>,y=<num>) - (x=<num>,y=<num>)]. L<num> buffer allocated at <num> and static circular buffer region ends at <num>
Model Name: Stable_Diffusion NumOps: 184683.0 CompileDepth: 6 Torch Name: aten::convolution Error: Error: TT_THROW @ /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/allocator/bank_manager.cpp:140: tt::exception info: Out of Memory: Not enough space to allocate <num> B L<num> buffer across <num> banks, where each bank needs to store <num> B
[...]

Issue #329, opened 2025-02-20
flux.1


AleksKnezevic:
Add test for [flux.1 dev](https://huggingface.co/black-forest-labs/FLUX.1-dev) and [flux.1 schnell](https://huggingface.co/black-forest-labs/FLUX.1-schnell) Tests should run in nightly op-by-op pipeline.

kmabeeTT:
Can we add flux models back yet @jazpurTT / @jameszianxuTT , they were previously pulled because of disk space issues

jameszianxuTT:
Yes, need to do some more testing but this should be possible now since #630 was merged. I'll get on it today.

kmabeeTT:
Marking closed, this got merged back - thanks James!

PR #639, opened 2025-04-21
Disable golden calculation/verification when evaluating in op_by_op mode

jazpurTT:
### Ticket
Follow up on #329 and #637

### Problem description
When working on models from tickets like #329 and #637 on op_by_op mode the compilation/debugging time is extended because we run the full torch model to get the golden first. The golden is not needed for the op_by_op flow so we want to disable it. This should also speed up the op_by_op CI workflows.

### What's changed
Separated test_model_eval() into two modes, if the compile config depth is set to op_by_op we just compile and run the model using the op_by_op executors, and we skip the golden/output, if not, we use the already existing logic to eval compilation and verify golden

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -305,8 +305,16 @@ class ModelTester:
         )
         return model
 
-    @torch.inference_mode()
     def test_model_eval(self, on_device=True, assert_eval_token_mismatch=True):
+        if (
+            self.compiler_config.compile_depth == CompileDepth.COMPILE_OP_BY_OP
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):
Agree that full-model-goldens aren't really relevant for op by op analysis so there's no need to precompile the whole model. 

Will monitor to see how this affects nightly times.

kmabeeTT (approved):
Awesome - thank you!

PR #641, opened 2025-04-22
Cache golden values for unique ops, and re-use them next time

AleksKnezevic:
### Problem description
Op-by-op tests were taking too long. This was especially true in diffusion/generative tests where models needed to be run in a loop, and total ops >> unique ops. 

### What's changed
Instead of calculating the golden values of ops each time (allowing us to verify e2e) we cache golden values of each unique op and propagate it. I believe this is safer than the empty tensor change @jazpurTT proposed since we should see fewer runtime errors than with just processing unbounded random tensors, but let's discuss. 

I did some additional testing, it's only calculating the goldens in bfp16 that's problematic. For a single iteration of stable diffusion UNET, calculating goldens in bf16 takes ~20 minutes and calculating them in fp32 takes 3 seconds!  Updated PR to calculate goldens in fp32. 

Change RMBG, and StableDiffusion tests to use bfloat16 to reduce memory footprint especially for conv2d now that they run fast.

### Checklist

### DIFF
--- a/tt_torch/dynamo/torch_backend.py
+++ b/tt_torch/dynamo/torch_backend.py
@@ -7,6 +7,7 @@ import tt_mlir
 import torch_mlir
 import os
 import re
+import time
 import ml_dtypes
 import numpy as np
 from torch_mlir.dialects import torch as torch_dialect
@@ -158,6 +159,27 @@ def lower_to_stable_hlo(module, op=None, enable_ir_printing=False):
 ##################################################################
 # TorchExecutor covers all CompileDepth options except for EXECUTE
 ##################################################################
+
+
+def cast_ios_and_run(node, args, kwargs):
+    try:
+        out_df = node.meta["tensor_meta"].dtype
+        out_df_known = True
+    except Exception:
+        out_df_known = False
+
+    if out_df_known:
+        cast_args = [
+            arg.to(torch.float32)
+            if isinstance(arg, torch.Tensor) and torch.is_floating_point(arg)
+            else arg
+            for arg in args
+        ]
+        golden = node.target(*cast_args, **kwargs)
+        golden = golden.to(out_df)
+    else:
+        golden = node.target(*args, **kwargs)
+    return golden
 
 
 class TorchExecutor(OpByOpExecutor):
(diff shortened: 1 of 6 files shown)


jazpurTT (approved):

tt_torch/dynamo/torch_backend.py
```diff
@@ -162,6 +162,23 @@ def lower_to_stable_hlo(module, op=None, enable_ir_printing=False):
 
 import time
 
+
+def cast_ios(node, args, kwargs):
```
jazpurTT, line 166: why do we need to cast the inputs? won't it slow down the execution?

AleksKnezevic: Doing the math in bf16 is MUCH slower than in FP32. For one iteration of SD unit calculating intermediates in bf16 took 20 minutes, and casting to fp32, doing the math in fp32 and casting back took less than 3 seconds. So it's much faster.

tt_torch/dynamo/torch_backend.py
```diff
@@ -411,6 +448,9 @@ def run_gm_op_by_op(self, *inputs):
         if self.stderror_redirected:
             os.unlink(self.file_stderr.name)
             self.stderror_redirected = False
+        print(
```
kmabeeTT, line 451: Mixed feelings on keeping all the elapsed time profiling code above but it's helpful so I think it's okay. The Compiling/Running, etc print markers now occur after the action instead of before, which is side effect to keep in mind.  Not a big deal.

kmabeeTT (approved):
Awesome - looks good. Interesting change, learned something new here.

kmabeeTT:
From branch CI [run](https://github.com/tenstorrent/tt-torch/actions/runs/14604189725) (thx for launching), number of failing conv2d cases dropped in half from 22 to 11 across RMBG and SD tests in op-by-op flow (nice!) though somehow 2 aten::view (ttnn.reshape) ops remain f32 and still assert at runtime in StableDiffusion as before (darn).  I updated description above to mention the BFP16 change in these tests so squashed message is more complete, and also pushed a change to RMBG post-processing to conver a tensor back to float32 to hopefully solve this:

```
        if isinstance(pic, torch.Tensor):
            if pic.ndim == 3:
                pic = pic.permute((1, 2, 0))
>           pic = pic.numpy(force=True)
E           TypeError: Got unsupported ScalarType BFloat16
```

new CI running [here](https://github.com/tenstorrent/tt-torch/actions/runs/14606298598) should pass.

AleksKnezevic:
@jazpurTT, with this change in, can you please see whether you can run the full flux test?

jazpurTT:
> @jazpurTT, with this change in, can you please see whether you can run the full flux test?

@AleksKnezevic Just tested them, the second inference sequence uses the cached golden from the first sequence making the run much shorter

PR #643, opened 2025-04-22
Fix data parallel demo by creating a new instance of the `options` 

hshahTT:
### Problem description
In the initial multi-device [PR](https://github.com/tenstorrent/tt-torch/pull/572) that added the data parallel demo, there was a bug that caused the models to only compile for the last device in the acquired `devices` list. This was due to the options map being passed by reference to each torch.compile call. This is fixed now by creating a fresh map for each iteration.

### Checklist

### DIFF
--- a/demos/resnet/README.md
+++ b/demos/resnet/README.md
@@ -47,12 +47,15 @@ from tt_torch.tools.device_manager import DeviceManager
 ...
 def main():
     ...
-    options = {}
-    options["compiler_config"] = cc
     # Acquires all available devices and returns them in a list
     parent, devices = DeviceManager.acquire_available_devices()
     tt_models = []
     for device in devices:
+        # It is important that each `options` map being passed to `torch.compile`
+        # is a new object.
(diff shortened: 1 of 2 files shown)


demos/resnet/README.md
```diff
@@ -47,12 +47,15 @@ from tt_torch.tools.device_manager import DeviceManager
...
     # Acquires all available devices and returns them in a list
     parent, devices = DeviceManager.acquire_available_devices()
     tt_models = []
     for device in devices:
+        # It is important that each `options` map being passed to `torch.compile`
+        # is a new object.
+        # Otherwise the "device" key will get overwritten for all previous devices.
+        options = {}
+        options["compiler_config"] = cc
         options["device"] = device # Explicitly compile the model for a specific device
```
jnie-TT, line 59: Probably outside the scope of this PR, but if options starts to get large it might be worthwhile to define a class like `CompileOptions` that contains everything we need. Or else it's not immediately clear what we need to set/populate in a vanilla python dictionary.

jnie-TT (approved):

PR #644, opened 2025-04-22
Add llmbox

jmcgrathTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,15 +24,17 @@ jobs:
       fail-fast: false
       matrix:
         build: [
-          {runs-on: wormhole_b0, name: "run"},
+          {runs-on: wormhole_b0, name: "run", container-options: "--device /dev/tenstorrent/0"},
+          {runs-on: llmbox, name: "run", container-options: "--device /dev/tenstorrent/0 --device /dev/tenstorrent/1 --device /dev/tenstorrent/2 --device /dev/tenstorrent/3"},
         ]
 
+    name: tests (${{ matrix.build.runs-on }}, ${{ matrix.build.name }})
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):

jmcgrathTT:
@jameszianxuTT That is correct. Currently, we only have one llmbox per repo, but I can look into adding more. 

But if you want, we can revert and move it to nightly.

jameszianxuTT:
@jmcgrathTT I'll prepare a new pipeline for a smaller llmbox smoketest that can run in OnPR. 

Discussed this with tt-torch team and we don't need to run all the tests in `run-tests.yml` on llmbox

jameszianxuTT:
Out of curiosity:
image

This queuing lag isn't horrible; and in no 1hr time window did we spend more than 47 minutes split between all active OnPR runs waiting for llmbox availability.

In theory if we got hammered by many, many PRs within the same 1hr window, we would start seeing more of an queuing lag, but we're not at that rate yet (even though we hit a record # of PRs yesterday)

jameszianxuTT:
On a closer look, jobs on average waited no more than ~25 minutes for LLMbox availability. This is not so great because it effectively adds ~50-90% to the actual runtime of the onPR job.


image

jmcgrathTT:
@jameszianxuTT I Agree. I think the best course of action would be to comment out that runner for now and then move it too a lighter job that you mentioned earlier. There are a few things I want to keep in the original PR. 

Thoughts?

jmcgrathTT:
> Added you as reviewer to https://github.com/tenstorrent/tt-torch/pull/687

Thanks @jameszianxuTT I approved

PR #647, opened 2025-04-23
A couple parse_op_by_op_results.py bugfixes and torch_backend print tweaks

kmabeeTT:
### Ticket
None

### What's changed
 - Fix simple bug in parse_op_by_op_results.py where timeout/unknown errors parsed from wrong column since recent change
 - Fix sneaky bug in parse_op_by_op_results.py where long cells exceeding xlswriter limit will make rendering corrupted, making cells appear empty in same row. Some ops hitting this with huge IRs, noticed when going through compile errors.
 - A couple minor changes to op_by_op flow debug prints to add date/time when saving unique_ops.json and limit compiling/running/golden time to 2 decimal places. Another ticket exists to reduce verbosity of prints may change this.

### Checklist
- [x] Tested script changes on last night's nightly regression .json files to generate new xlsx

### DIFF
--- a/results/parse_op_by_op_results.py
+++ b/results/parse_op_by_op_results.py
@@ -399,8 +399,9 @@ def create_summary_worksheet(workbook, models_info):
         worksheet.write(row, 17, executing_formula)
         worksheet.set_column(17, 17, 15)
 
-        # Determine how many ops for model hit unknown error (column M is "Compile error")
-        unknown_errors_formula = f'=COUNTIF(INDIRECT("\'" & "{model_name}" & "\'!N:N"), "Error message not extracted.")'
+        # Determine how many ops for model hit unknown error.
+        col_id = "N"  # Compile Error column
+        unknown_errors_formula = f'=COUNTIF(INDIRECT("\'" & "{model_name}" & "\'!{col_id}:{col_id}"), "Error message not extracted.")'
         worksheet.set_column(18, 18, 2)
         worksheet.set_column(19, 19, 13)
         worksheet.write(2, 19, "Unknown Errors")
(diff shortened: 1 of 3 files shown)


kmabeeTT:
For the XLS file dropping/corrupting fields due to huge text in cells exceeding 32K characters, here is example Before:

image

After (nice, can see all Errors now):

image

tt_torch/dynamo/torch_backend.py
```diff
@@ -449,7 +449,7 @@ def run_gm_op_by_op(self, *inputs):
             os.unlink(self.file_stderr.name)
             self.stderror_redirected = False
         print(
-            f"Compiling time: {OpByOpExecutor.compiling_time}, Running time: {OpByOpExecutor.running_time}, Golden time: {OpByOpExecutor.golden_time}"
+            f"Total Time. Compiling: {OpByOpExecutor.compiling_time:.2f} s, Running: {OpByOpExecutor.running_time:.2f} s, Golden: {OpByOpExecutor.golden_time:.2f} s"
```
ddilbazTT, line 452: very minor nit: Should it be `Total Time:` instead of `.` ?

results/parse_op_by_op_results.py
```diff
@@ -397,7 +400,7 @@ def create_summary_worksheet(workbook, models_info):
         worksheet.set_column(17, 17, 15)
 
         # Determine how many ops for model hit unknown error (column M is "Compile error")
```
ddilbazTT, line 402: nit: perhaps instead of using M / N literally, we could have `column_id` and set that to M/ N?

ddilbazTT (approved):

tt_torch/dynamo/torch_backend.py
kmabeeTT: I generally only use colon before printing a value, so I switched to dash here instead (Total Time - Compiling: 123 s, ...) as compromise :)

results/parse_op_by_op_results.py
```diff
@@ -28,6 +28,9 @@
     "EXECUTED",
 ]
 
+# xlswriter limit, neighbor cells corrupted if exceeded.
+MAX_CELL_LEN = 32767
```
jameszianxuTT, line 32: good to know!

kmabeeTT: Yeahhhhhh.... this took me on a wild trip.

Issue #596, opened 2025-04-10
Add EfficientNet model to onnx op by op flow


kmabeeTT:
We eventually want this model in ONNX for a customer. Please add EfficientNet to onnx op-by-op flow if possible (otherwise torch is fine). It's model_group=red model.

Source: https://github.com/lukemelas/EfficientNet-PyTorch

Looks like in addition to efficientnet-b0 they have several other flavors (b1-b7) we can consider adding them all.

ddilbazTT:
I would like to try this model. There's a chance the codebase might be quite outdated since it was last updated 4 years ago.

PR #648, opened 2025-04-23
Add EfficientNet #596

ddilbazTT:
### Ticket
#596 

### Problem description
Add EfficientNet PyTorch and ONNX models

### What's changed
- Added EfficientNet test files 
- Added `torch_input_to_onnx` function which can now handle tuple, list, tensor inputs more comprehensively 
- Fixed `run_model` bug; made it comprehensive for all input types


kmabeeTT:
Nice! Really cool that this is passing full model execute at the same time it's being added - thx for adding to torch_op_by_op weekly list. 

For full-model-execute, I think we don't want to get into the habit of duplicating onnx and torch versions though for cases like this (same model run twice, counted twice in red list), just the onnx version in full model execute is fine, and eventually (once storage is worked out) we could consider removing the onnx.export() and just keeping the onnx.load(), at least that was the original idea.

jameszianxuTT (approved):

tests/models/EfficientNet/test_EfficientNet.py
```diff
@@ -0,0 +1,83 @@
...
+from efficientnet_pytorch import EfficientNet
+
+import pytest
+from tests.utils import ModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        model = EfficientNet.from_pretrained("efficientnet-b0")
```
kmabeeTT, line 19: Did you try the other variants b1-b7 mentioned in ticket too? Free coverage if we can add the ones that also pass for us and the runtime is reasonable. Forge is running b0 and b4 it looks like, with others commented out.

image

kmabeeTT (approved):
Thanks! Would be good to try the other variants, and add them to op-by-op flow if runtime is reasonable.

tests/models/EfficientNet/test_EfficientNet.py
kmabeeTT: We don't have to add them all, but b0 and b4 like forge is good idea.

brataTT (approved):

tests/models/EfficientNet/test_EfficientNet.py
ddilbazTT: Sounds good! I'll give it a try.

ddilbazTT: Hi Kyle, here are the runs so you could take a look at the runtimes:
full model execute (running all variants on onnx): [https://github.com/tenstorrent/tt-torch/actions/runs/14646198836/job/41102109063](https://github.com/tenstorrent/tt-torch/actions/runs/14646198836/job/41102109063)
op-by-op (running all variants on torch op-by-op): [https://github.com/tenstorrent/tt-torch/actions/runs/14645790450?pr=648](https://github.com/tenstorrent/tt-torch/actions/runs/14645790450?pr=648)

kmabeeTT: Amazing, this is great- let's keep all variants full model execute, but just b0 variant in op-by-op weekly to keep things light.  op-by-op weekly importance will decrease over time I suspect anyways. And let's set it to merge to be included in nightly tonight!

kmabeeTT: Try to rebase on latest and resolve the yml file conflict too either locally or in gui here.

tt_torch/tools/utils.py
```diff
@@ -775,6 +775,41 @@ def wrapper(*args, **kwargs):
...
+        return tensor.detach().cpu().numpy()
+
+    # single tensor
+    if isinstance(torch_input, torch.Tensor):
+        if len(input_names) == 1:
+            return {input_names[0]: tensor_to_numpy(torch_input)}
+
+    # dictionary input
+    if isinstance(torch_input, dict):
+        if len(input_names) == len(torch_input):
```
AleksKnezevic, line 794: should there be an else here?

.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -27,70 +27,84 @@ jobs:
       fail-fast: false
       matrix:
         build: [
+          # {
```
AleksKnezevic, line 30: You can leave it on this one, but it would really help legibility it future PR's if you made this change (to strip out all the nightly tests) in a separate branch. It would also ensure it doesn't accidentally get merged into main.

tt_torch/tools/utils.py
ddilbazTT: That would be captured by `ValueError` which comes at the end

AleksKnezevic: right!

PR #656, opened 2025-04-24
Introduce tt-forge-models repo and use it (loaders, src) for 5 models to start

kmabeeTT:
### Issue

https://github.com/tenstorrent/tt-torch/issues/657

### Summary

This PR starts to use a new shared model repository structure ([tt-forge-models](https://github.com/tenstorrent/tt-forge-models)) to centralize model implementations used across TT-Forge frontend repositories. This eliminates code duplication and improves maintainability by providing a single source of truth for model implementations.

### Changes - TT-Forge-Models
 - Migrated model implementations for YOLOv4, YOLOv3, OFT, VGG19_UNET from test directories to tt-forge-models
 - Implemented a consistent API across all models with load_model() and load_inputs() methods
 - README.md doc with usage details and how to add models, also convert BERT as HF example.
 - Used relative imports for package portability
 
### Changes - TT-Torch (this PR)
 - Use tt-forge-models project via ExternalProjectAdd in third_party/tt_forge_models/
 - Updated tests to use the new loader classes, deleted local model implementations, loader code
 - Include tt-forge-models in wheel generated for CI to have access to files

### Benefits
 - Standardization: Consistent model loading API across all models
 - Maintainability: Single source of truth for model implementations
 - Portability: Can be used as a git submodule or dependency in other projects
 - Simplification: Reduced test code complexity by hiding implementation details, removed from release wheel


third_party/tt_forge_models/yolov3/loader.py
```diff
@@ -0,0 +1,108 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
AleksKnezevic, line 1: To me it makes more sense to have model src in a separate folder, so the base `loader.py` could be in the root and everything else in a `src` folder. What do you guys think?

AleksKnezevic:
This looks really good @kmabeeTT. When you're ready, we can migrate to a separate repo and have the forge-fe tt-xla folks review.

AleksKnezevic:
Also, I think it makes sense having all models here. Even the ones that we just straight up load from HF. Then they can all be reusable across FEs.

third_party/tt_forge_models/yolov3/loader.py
kmabeeTT: Is this what you're thinking? I like it.

```
third_party/tt_forge_models/
├── __init__.py
├── base.py
├── yolov4/
│   ├── __init__.py
│   ├── loader.py        # Public interface
│   └── src/             # Implementation details
│       ├── yolov4.py
│       ├── downsample1.py
│       ├── downsample2.py
│       └── ...
├── yolov3/
│   ├── __init__.py
│   ├── loader.py        # Public interface
│   └── src/             # Implementation details
│       └── yolov3.py
└── oft/
    ├── __init__.py
    ├── loader.py        # Public interface
    └── src/             # Implementation details
        ├── oftnet.py
        ├── oft.py
        └── utils.py
```

jameszianxuTT:
Looks like my [CI test autodetection fails to import your loaders](https://github.com/tenstorrent/tt-torch/actions/runs/14654693673/job/41127861440?pr=656#step:10:265) but the paths look okay ... am I missing something?

kmabeeTT:
> Looks like my [CI test autodetection fails to import your loaders](https://github.com/tenstorrent/tt-torch/actions/runs/14654693673/job/41127861440?pr=656#step:10:265) but the paths look okay ... am I missing something?

Whoops, I was missing a small change in my previous push, fixed it today and force pushed, passing now.

kmabeeTT:
https://github.com/tenstorrent/tt-forge-models finally got switched to public few minutes ago, so I will be pulling this in soon and switching this PR to be non-draft.

kmabeeTT:
Time permitting, I plan to convert a couple more models (vgg19_unet which has a src file, and maybe bert as a HF model example) into tt-forge-models before I merge this later today - but want to get this out for review + approval in meantime.

See the README and docs in https://github.com/tenstorrent/tt-forge-models for reference.

tests/models/oft/test_oft.py
```diff
@@ -22,17 +21,7 @@ def _load_model(self):
...
-        dummy_image = torch.randn(batch_size, 3, 224, 224)
-        dummy_calib = torch.randn(batch_size, 3, 4)
-        grid_size = (80.0, 80.0)  # width, depth
-        grid_depth = 8
-        grid_depth = int(grid_size[1] / self.grid_res)
-        grid_width = int(grid_size[0] / self.grid_res)
-        dummy_grid = torch.randn(batch_size, grid_depth, grid_width, 3)
-        input = (dummy_image, dummy_calib, dummy_grid)
-        return input
+        return OFTLoader.load_inputs(grid_res=self.grid_res, grid_size=(80.0, 80.0))
```
AleksKnezevic, line 24: Would it make sense for this to have a generic name, instead of OFTLoader? Then we could eventually have a programatic test that would go into all files within some `models` dir and load the model, load the inputs without needing to customize each test

third_party/CMakeLists.txt
```diff
@@ -118,3 +119,15 @@ else()
     install(DIRECTORY ${TORCH_MLIR_INSTALL_PREFIX}/python_packages/torch_mlir/torch_mlir DESTINATION "${CMAKE_INSTALL_PREFIX}" USE_SOURCE_PERMISSIONS)
 
 endif()
+
+# tt-forge-models - Python-only project, no need to build.
+ExternalProject_Add(
+    tt-forge-models
+    SOURCE_DIR ${TTTORCH_SOURCE_DIR}/third_party/tt-forge-models
```
kmabeeTT, line 126: I'll have to change this, cannot import python package from folders with dashes in the name, so will have to make the directory we checkout tt-forge-models into called tt_forge_models

tests/models/oft/test_oft.py
kmabeeTT: Sure, I can make that change.. rename them all to be `ModelLoader` for each model instead of model-specific names like `OFTLoader`, `YOLOv3Loader` etc.  We also talked offline about wanting to make the interface consistent, ie remove flexibility from test... currently have this:

```
        model = OFTLoader.load_model(
            num_classes=1,
            frontend="resnet18",
            topdown_layers=8,
            grid_res=self.grid_res,
            grid_height=4.0,
        )
```
and this
```
        model = YOLOv3Loader.load_model(dtype=torch.bfloat16)
```     

Both would be changed to this pattern for load_model() and load_inputs() taking no arguments typically:

```
from third_party.tt_forge_models.yolov3 import ModelLoader
model = ModelLoader.load_model()
inputs = ModelLoader.load_inputs()
```

kmabeeTT:
Converting this back to draft, need to polish it more since turns out CI does things different and uses wheels and so third_party/tt_forge_models/ is not available without some updates..somewhere...maybe setup.py to include the dir wholesale.

kmabeeTT:
Follow up to Aleks' idea for tests to clone tt-forge-models for CI.  This is a possible change fc6e3aee58ecd6e90c3fa33234c15c5b9ad570fa if we really want to exclude models from the wheel (tested on CI [here](https://github.com/tenstorrent/tt-torch/actions/runs/14847175044)), but it gets applied to 12 yaml files (all 12 places where tt-torch wheel is installed) so it's a bit large, and annoying.

```
diff --git a/.github/workflows/run-full-model-execution-tests.yml b/.github/workflows/run-full-model-execution-tests.yml
index c6cbb92..1148018 100644
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -158,6 +158,21 @@ jobs:
         source env/activate
         pip install ${{ steps.strings.outputs.dist-dir }}/*.whl

+    - name: Get tt-forge-models version from CMakeLists.txt
+      id: get-forge-models-version
+      shell: bash
+      run: |
+        VERSION=$(grep "^set(TT_FORGE_MODELS_VERSION" third_party/CMakeLists.txt | cut -d'"' -f2)
+        echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+        echo "Detected tt-forge-models version: $VERSION"
+
+    - name: Clone tt-forge-models
+      uses: actions/checkout@v4
+      with:
+        repository: tenstorrent/tt-forge-models
+        path: third_party/tt_forge_models
+        ref: ${{ steps.get-forge-models-version.outputs.version }}
+
     - name: Run Full Model Execution Tests
       env:
         HF_HOME: ${{ env.DOCKER_CACHE_ROOT }}/huggingface
```

@jameszianxuTT had an idea that I thought about too - that we should just keep the `third_party/tt_forge_models` bundled into the wheels, but only for CI, enabled with an optional argument... one line change to pass it, and one line change to wrap the code in setup.py - @AleksKnezevic what do you think?

```
diff --git a/.github/workflows/run-build.yml b/.github/workflows/run-build.yml
index 38f288a..217d67f 100644
--- a/.github/workflows/run-build.yml
+++ b/.github/workflows/run-build.yml
@@ -76,11 +76,11 @@ jobs:
     # Build project
     - name: Build
       shell: bash
       run: |
         source env/activate
-        python setup.py bdist_wheel --code_coverage ${{ inputs.setup-args }}
+        python setup.py bdist_wheel --code_coverage --include-models ${{ inputs.setup-args }}

     - name: Verify tt-mlir SHA override
       if: ${{ inputs.mlir_override }}
       continue-on-error: true
       shell: bash
```

AleksKnezevic:
@jameszianxuTT's idea of optionally including in the wheel seems like a good compromise. Are there any other OSS projects doing this?

third_party/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(TT_MLIR_VERSION "2e81bbc04af6c86899eb4ff247bcf65cc9a17579")
 # torch-mlir is using branch tt_torch/main
 # see issue https://github.com/tenstorrent/tt-torch/issues/671
 set(TORCH_MLIR_VERSION "7c86f6d21de6d548c41cc56ebf0d799e1eadc626")
+set(TT_FORGE_MODELS_VERSION "3ef1f8ea483316cfa530eee199151971fafe237d")
```
kmabeeTT, line 10: I thought about this previously, but I am going to propose to move these versions closer to where they are used in file rather than at top of file because having them so close together will cause conflicts for folks, which then requires CI to be rerun which is completely not neccessary. Happened today:

```
<<<<<<< kmabee/tt_forge_models_poc
set(TORCH_MLIR_VERSION "7c86f6d21de6d548c41cc56ebf0d799e1eadc626")
set(TT_FORGE_MODELS_VERSION "3ef1f8ea483316cfa530eee199151971fafe237d")
=======
set(TORCH_MLIR_VERSION "fe292ef77433c9da6ff539780f2b09d8fab146f4")
>>>>>>> main
```

ddilbazTT (approved):

.github/workflows/run-build.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       shell: bash
       run: |
         source env/activate
-        python setup.py bdist_wheel --code_coverage ${{ inputs.setup-args }}
+        python setup.py bdist_wheel --code_coverage --include-models ${{ inputs.setup-args }}
```
jameszianxuTT, line 81: I meant for inputs.setup-args to explicitly all arguments passed to the builder job. It would be larger change, but I think clearer to pass "--include-models" through that mechanism from the caller, rather than to embed it here. 

This makes the arguments passed to the build more explicit, and gives us a path to optionally excluding tt-forge-models in some jobs that won't need it in the future (eg. small smoketests, bleeding edge release builds etc.) 

Example from nightlies:
```
  build-debug:
    needs: [pre-commit, spdx, docker-build]
    uses: ./.github/workflows/run-build.yml
    secrets: inherit
    with:
      docker-image: ${{ needs.docker-build.outputs.docker-image }}
      setup-args: "--build_runtime_debug"
      artifact-key: "-debug" # avoid name collision for generated artifact
```

jameszianxuTT (approved):
Other than one nit~ish comment, looks good.

.github/workflows/run-build.yml
kmabeeTT: Oh nice, yes I briefly considered that - will adjust for that change, thx for pointing it out!

kmabeeTT: Made that change, there were 6 places using `run-build.yml`, now they all pass `--include-models` via `setup-args`, example:

```
diff --git a/.github/workflows/run-build.yml b/.github/workflows/run-build.yml
index 217d67fb..38f288a2 100644
--- a/.github/workflows/run-build.yml
+++ b/.github/workflows/run-build.yml
@@ -78,7 +78,7 @@ jobs:
       shell: bash
       run: |
         source env/activate
-        python setup.py bdist_wheel --code_coverage --include-models ${{ inputs.setup-args }}
+        python setup.py bdist_wheel --code_coverage ${{ inputs.setup-args }}

     - name: Verify tt-mlir SHA override
       if: ${{ inputs.mlir_override }}
diff --git a/.github/workflows/weekly-tests.yml b/.github/workflows/weekly-tests.yml
index 60b48985..7b0a1970 100644
--- a/.github/workflows/weekly-tests.yml
+++ b/.github/workflows/weekly-tests.yml
@@ -23,6 +23,7 @@ jobs:
     secrets: inherit
     with:
       docker-image: ${{ needs.docker-build.outputs.docker-image }}
+      setup-args: "--include-models"

```

kmabeeTT:
Going to merge this PR shortly now that comments addressed, and bunch of tt-forge-models repo AI's are done including:

- is open sourced, locked down (PR's only), requires approval before merge PR, has CI setup on it, has a bunch of feedback items implemented on it.

PR #659, opened 2025-04-25
Added test logs needed for nightly fail inspector

nsumrakTT:
### Ticket

### Problem description
To implement bisect algorithm test logs are needed.

### What's changed
Added test logging and upload as artifacts.
Existing logging in run-op-by-op-model-testing-nightly is changed to fit specific file name format.

### DIFF
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -175,6 +175,7 @@ jobs:
         source env/activate
 
         mkdir -p ${{ steps.strings.outputs.test_report_dir }}
+        rm -f pytest.log
 
         # prevent script termination on failure
         set +e
@@ -192,7 +193,8 @@ jobs:
 
             pytest --durations=0 -v "$test" \
               --junit-xml="$report_file" \
-              --cov=tt_torch --cov-report term --cov-report xml:coverage.xml --cov-append
+              --cov=tt_torch --cov-report term --cov-report xml:coverage.xml --cov-append \
+              2>&1 | tee -a pytest.log
 
 
             # Check the return code of pytest
@@ -214,6 +216,13 @@ jobs:
           echo "All tests passed."
           exit 0
         fi
+
+    - name: Upload Test Log
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        name: test-log-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
+        path: pytest.log
 
     - name: Upload Test Report Models
       uses: actions/upload-artifact@v4
--- a/.github/workflows/run-op-by-op-model-tests-nightly.yml
+++ b/.github/workflows/run-op-by-op-model-tests-nightly.yml
@@ -252,6 +252,7 @@ jobs:
           fi
         done
 
+        cp pytest.log full_job_output.log
 
         # If any test failed, exit nonzero to mark the job as failed
         if [ $failures -ne 0 ]; then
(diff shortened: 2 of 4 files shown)


.github/workflows/run-op-by-op-model-tests-nightly.yml
```diff
@@ -287,5 +284,5 @@ jobs:
       if: success() || failure()
       uses: actions/upload-artifact@v4
       with:
-        name: full-logs-nightly-${{ matrix.build.name }}
-        path: full_job_output.log
```
jameszianxuTT, line 291: There is one hardcoded reference to this logfile name that will probably fail to work. 
https://github.com/tenstorrent/tt-torch/blob/ad0b5b8ee5f95fd3a8eb114ed4f49c551a325c2f/results/download_artifacts.py#L105-L131

I don't really think this is too critical but it would make it harder to have that convenient testname prepending for op by op tests @kmabeeTT. It could be worked around, but we'd need to find another way to determine that the artifact is an op by op test without relying on the side effect of the hardcoded artifact name.

.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -215,6 +217,13 @@ jobs:
           exit 0
         fi
 
+    - name: Upload Test Log
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        name: test-log-${{ matrix.build.runs-on }}-${{ matrix.build.name }}-${{ steps.fetch-job-id.outputs.job_id }}
+        path: pytest.log
+
```
jameszianxuTT, line 226: Just a note that this particular workflow `.github/workflows/run-full-model-execution-tests.yml` is run onPR so there is no need for auto-bisection on these particular tests. But uploading the logs may still be helpful anyways.

nsumrakTT: The same workflow is run on nightly so I'll need report. For PRs, it doesn't hurt if we have it :-)

.github/workflows/run-op-by-op-model-tests-nightly.yml
nsumrakTT: Thanks for letting me know, I search whole repo but didn't find it. Since my scripts also rely on artifact name, I'll just duplicate artifact upload.

nsumrakTT:
@jameszianxuTT That is good point. But it makes script more complicated because downloading and parsing predefined name can easily match specific logs, eg. type of hardware/machine it was run on. With logs it is possible, just more complicated imho.

.github/workflows/run-op-by-op-model-tests-nightly.yml
jameszianxuTT: Oh, duplicating the artifact upload is an easier solution, yeah.

jameszianxuTT (approved):

kmabeeTT:
@nsumrakTT / @jameszianxuTT With this change, our stdout in op-by-op-nightly went from nearly [nothing](https://github.com/tenstorrent/tt-torch/actions/runs/14656701731/job/41170286991), (just pass/fail results) of tests ...

```
[ 1 / 3 ] tests/models/t5/test_t5.py::test_t5 PASSED
[ 2 / 3 ] tests/models/flan_t5/test_flan_t5.py::test_flan_t5 PASSED
[ 3 / 3 ] tests/models/speecht5_tts/test_speecht5_tts.py::test_speecht5_tts PASSED
```

.. to [massive](https://github.com/tenstorrent/tt-torch/actions/runs/14676575646/job/41194090286) (exceeding limits) and not possible for human to make sense of output and see which tests are failing.  It's due to high verbosity of op-by-op flow that used to just write to logfile, but now is shown in github stdout.

Do we really needs output shown to stdout for bisect flow to work, or can that be changed just to write to new log files?

nsumrakTT:
@kmabeeTT That is cool, I just needed a log file not an actual console output. Its my mistake, sorry for the inconvenience.

PR #660, opened 2025-04-25
Create dynamic timeout logic for benchmarks

jameszianxuTT:
### Ticket
None

### Problem description
Individual pytests are run in isolated runners in weekly depth benchmarks. Due to runner count limitations, enough hanging tests can create a runner bottleneck, and statistical simulation suggests a small probability of benchmarks running for >10h. Most recent benchmark tests have run for 9h30m or so which comes close.

### What's changed
Setup a 3 stage timeout and infra to dynamically set that timeout based on known test timings.
1.  If there is no known duration for the test, set the timeout to a reasonable default of 2 hours
2.  If a test is expected to take less than 30 minutes, set the timeout to 1 hour.
3.  If the expected duration exceeds 30 minutes, set it to the MAX_TIMEOUT of 500m 

### Checklist
- [x] Partial benchmark tests have been run to check that the timeout is calculated, passed and processed correctly, and that the timeout killer works

### DIFF
--- a/tt_torch/tools/benchmark_promotion.py
+++ b/tt_torch/tools/benchmark_promotion.py
@@ -13,6 +13,7 @@ import pprint
 import json
 
 MAXIMUM_JOB_TIMEOUT_MINUTES = 500  # 500 minutes maximum per-job timeout
+DEFAULT_JOB_TIMEOUT_MINUTES = 120
 
 
 def enumerate_all_tests(filter_full_eval=True):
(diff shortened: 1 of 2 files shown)


tt_torch/tools/benchmark_promotion.py
```diff
@@ -232,9 +256,11 @@ def generate_formatted_test_matrix_from_partitions(
                 "runs-on": runs_on,
                 "name": job_name,
                 "group-id": i,  # zero indexed
+                "t-o": actual_timeout,  # timeout in minutes, for timeout-minutes field
```
jameszianxuTT, line 259: Using short name to save on characters (so we don't go past the JSON string limit in Github Actions)

kmabeeTT:
This is interesting, I saw this but got side tracked, will take closer look before Sunday.

kmabeeTT (approved):
Nice - looks good!

Issue #651, opened 2025-04-23
Extend Tracy support to multiprocessed op by op tests


jameszianxuTT:
### Is your feature request related to a problem? Please describe

Currently #257 is only implemented for the full execute path. Tracy is a single process profiler and all the underlying infrastructure expects the load under investigation to be executed as a single process.

Because we wrap op by op tests in a multiprocessing workflow (I believe this is to avoid dealing with signals that would kill the parent python orchestrator process), this will be complicated to synchronize. 

### Describe the solution you'd like

- An extension to the pytest tracy server controller to respawn and synchronize against multiple serially spawned child target processes 
- Pytests to validate multiprocess trace generation & analysis workflows
 -Documentation for the profiling workflow for tt-torch

PR #661, opened 2025-04-25
Add tt-torch profile.py documentation, port and fix an obscure profiling deadlock

jameszianxuTT:
### Ticket
#651 

### Problem description
- There is no documentation on perf tracing in tt-torch
- A rare capture server deadlock scenario can occur due to recursive imports and pybind mechanics causing something in tt-metal or tt-mlir to report that the orchestrator process is a tracy client
- Port number cannot be specified for tracy client/server pair

### What's changed
- Add documentation file describing prerequisites, usage, limitations, debugging tips
- Move filemanager to its own file, to avoid anything in the import path from importing the tt_mlir bindings which creates a potential deadlock
- Add port specification option -p to profile.py
- Add location stripping to get torchfx node name in LOC column and preserve line numbered locations in the last column

### Checklist
- [x] (Todo) - Check for full op coverage in llama3 test

### DIFF
--- a/docs/src/profiling.md
+++ b/docs/src/profiling.md
@@ -8,6 +8,7 @@ tt-torch uses the tt-metal Tracy fork to manage a Tracy server. Tracy is a singl
 
 - In the tt-torch building step ([Building](./build.md)), it is required to configure your cmake build with the additional cmake directive `TT_RUNTIME_ENABLE_PERF_TRACE=ON`.
     - i.e. run: `cmake -G Ninja -B build -DTT_RUNTIME_ENABLE_PERF_TRACE=ON`
+- Paths in this document are given relative to the repo root.
 
 ## Usage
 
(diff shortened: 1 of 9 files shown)


jameszianxuTT:
Also: 
- TT Metal 65K trace limit fixed in https://github.com/tenstorrent/tt-metal/issues/20801
- TT MLIR trace autodump every 1K ops due to DRAM space limitations fixed in https://github.com/tenstorrent/tt-mlir/pull/2832

jameszianxuTT:
Verification - llama perf trace vs. ttir module defined ops set difference shows that:
- All ops in the perf trace are defined in ttir 
- All ops in the ttir show up in the perf trace, except some `{permute, view, unsqueeze}` which is expected

tt_torch/tools/profile.py
```diff
@@ -10,14 +9,32 @@
...
+    #   This could even be a header import from way down in the stack.
+
+    # The TRACY_PORT env variable is set in setup_tracy_server.
+
+    env_vars = dict(os.environ)
+    env_vars["TRACY_PORT"] = tracy_port
+    env_vars["TT_METAL_DEVICE_PROFILER"] = "1"
+    env_vars["TT_METAL_CLEAR_L1"] = "1"
+    env_vars["TT_METAL_DEVICE_PROFILER_DISPATCH"] = "0"
+    # env_vars["TT_METAL_PROFILER_SYNC"] = "1"  # fix desync between host and device calls - disabled due to CI issue
```
jameszianxuTT, line 34: Made a devex ticket for this breaking only in CI.

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
+# Profiling
```
AleksKnezevic, line 1: I believe you need a link to the doc in SUMMARY. @ddilbazTT to confirm.

AleksKnezevic (approved):
Other than link in SUMMARY if needed, looks good to me.

docs/src/profiling.md
ddilbazTT: you can use `docs/generate_summary.py` to generate the summary. I suggest using the version in this pr: https://github.com/tenstorrent/tt-torch/pull/674

jameszianxuTT: I see that in the [mdbook documentation](https://rust-lang.github.io/mdBook/format/summary.html) and will add this in. Probably should have a crosscheck in CI for this as well

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Introduction
+
+tt-torch uses the tt-metal Tracy fork to manage a Tracy server. Tracy is a single process profiler, and uses a client-server model to perform performance tracing on both host calls and on-device operations. tt-torch implements a wrapper called `profile.py` with custom orchestration logic to handle the spawning of the Tracy capture server and the client workload to be profiled, as well as report generation and data postprocessing functionality.
+
+## Prerequisites
+
+- In the tt-torch building step ([Building](./build.md)), it is required to configure your cmake build with the additional cmake directive `TT_RUNTIME_ENABLE_PERF_TRACE=ON`.
+    - i.e. run: `cmake -G Ninja -B build -DTT_RUNTIME_ENABLE_PERF_TRACE=ON`
+- Paths in this document are given relative to the repo root.
```
einnesTT, line 11: This isn't a prerequisite. It's important to describe as an assumption, but it should probably be a note. I would consider placing it after the introductory paragraph as a note. People are likely to see it standing out, and apply that information to the rest of the content.

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Usage
+
+The `profile.py` tool is the recommended entrypoint for profiling workloads in tt-torch.
+
```
einnesTT, line 16: Entrypoint is a docker reference. If you used docker to create your content, that would be useful to highlight. Or say something like 'If you are using Docker, XYZ information.'

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
...
+```
+profile.py [-h] [-o OUTPUT_PATH] [-p PORT] "test_command"
+```
+**Note: The `test_command` must be quoted!**
+
```
einnesTT, line 21: I think github lets you do an actual note if you use this formatting - [!NOTE] Note content here.

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
...
+
+As a minimal example, this will run and profile the MNIST test:
+```
```
einnesTT, line 24: example, the following command will run...

docs/src/profiling.md
jameszianxuTT: Our docs are published through mdbook, which I don't think supports such callout natively

docs/src/profiling.md
```diff
@@ -0,0 +1,43 @@
...
+python tt_torch/tools/profile.py "pytest -svv tests/models/mnist/test_mnist.py::test_mnist_train[full-eval]"
+```
+
+It generates the following report: `results/perf/device_ops_perf_trace.csv`. This report displays a table of operations executed on device and rich timing and configuration data associated with them.
+
+## Limitations
+
```
einnesTT, line 31: I would consider listing this as a section above Prerequisites. Frequently, people go through a document bit by bit, following the instructions. It would be easy for someone to start work on the instructions, then make this discovery, whereas if it's listed before they start, they can find something else for their project if these steps won't work for them.

ddilbazTT (approved):
Nit: you don't need to make any more changes, but sometimes you add new lines, probably by accident. I also do the same but it's probably better not to add those new lines to keep commits cleaner.

docs/src/profiling.md
ddilbazTT: I don't know what type of crosscheck we could have in CI, but one can run `docs` workflow on their branch to see how the website would look like and if their content is showing up: [https://github.com/tenstorrent/tt-torch/actions/workflows/pages.yml](https://github.com/tenstorrent/tt-torch/actions/workflows/pages.yml) Deployment would always fail since we only deploy from main, but you can download the artifacts and open `index.html` to see the website and whether the content you added is updated.

jameszianxuTT: Oh thanks, yeah, I didn't think of that.

PR #662, opened 2025-04-25
Replace EfficientNet onnx e2e tests with pytorch tests

ddilbazTT:
Remove MobileNetV2 onnx end-to end test since it runs until stablehlo. @kmabeeTT is working on an op-by-op-stablehlo workflow, where MobileNetV2 and EfficientNet will go to.


kmabeeTT (approved):
Thank you!

PR #669, opened 2025-04-28
Change op-by-op nightly back to not printing on screen, too verbose

kmabeeTT:
### Ticket
None

### Problem description
- Recent change f28104c introduced pytest.log but made op-by-op-flow also write to stdout in CI which is too much text and makes it impossible to read output, and exceeds CI output log length often

### What's changed
- Small change to prevent output from being printed to screen again, similar to previous behavior, but while still keeping it in pytest.log

### Checklist
- [x] Tested on branch https://github.com/tenstorrent/tt-torch/actions/runs/14697916675

### DIFF
--- a/.github/workflows/run-op-by-op-model-tests-nightly.yml
+++ b/.github/workflows/run-op-by-op-model-tests-nightly.yml
@@ -226,7 +226,7 @@ jobs:
           counter=$((counter + 1))
 
           echo "====== BEGIN LOG: $test_case ======" >> pytest.log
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):

jameszianxuTT (commented):
TIL that &>> is syntactic sugar for >>file &2>1

["3.6.5 Appending Standard Output and Standard Error"](https://www.gnu.org/software/bash/manual/html_node/Redirections.html)

kmabeeTT:
Back to quiet output w/ this change.

image

PR #674, opened 2025-04-28
TTNN Op Files and Website Updates

ddilbazTT:
### Problem description
TTNN Op md files do not have context to what's being reported. Failing ops have changed since last ttnn op trace update. 

### What's changed
- Regenerated ttnn md files with appropriate summary at the beginning
- Changed how SUMMARY.md file is generated and regenerated it.


ddilbazTT:
> Can you also add a `about` or `summary` page for the ops with a description of what's being shown here, i.e. something like:
> 
> ```
> The following pages have traces of operations that are currently not being compiled correctly. They can be updated by (INSTRUCTIONS).  
> ```

@AleksKnezevic I just wanted to confirm - we only show pages for unsupported ops? Not all ops?

kmabeeTT:
Just to add some context, I added the `--failures_only`a while back mainly since I got impression these nightly-regression-generated .json/.md files weren't being used by anyone, and so I could use the "failures only" to generate TTNN sweep/repro cases as requested by conv2d folks.  Not against removing the flag, maybe I can filter by something else if I need to do it again.

ddilbazTT:
@AleksKnezevic @kmabeeTT can I get a review?

kmabeeTT (approved):
Looks good to me now, thx for making small corrections.

PR #679, opened 2025-04-29
Add llama 7b pipeline parallel test

### LINKED ISSUE #591: Runtime Pipeline Parallel
This issue tracks initial bringup of pipeline parallel + simple test / model.  Ideally we use something like llama 3 8b which doesn't fit on a single chip.

jnie-TT:
### Ticket
closes #591 

### Problem description
We want a POC of a large model that doesn't fit on single device to run in runtime pipeline parallel fashion on multiple chips.

### What's changed
Added llama 7b that is split in half and runs pipeline parallel on 2 chips.

Since this doesn't fit into our `ModelTester` flow and also since LLama3b is getting bad pcc, adding this test currently with a separate flow. One future task is to integrate this into our `ModelTester` flow so we can properly gather results/reports like we do for other models FYI @jameszianxuTT .

### Checklist

### DIFF
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -92,6 +92,11 @@ jobs:
                 tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification_generality[full-eval-wide_resnet101_2]
             "
           },
+          {
+            runs-on: n300, name: "eval_6", tests: "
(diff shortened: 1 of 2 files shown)


nsmithtt (approved):
🔥

jameszianxuTT:
Looks like it [passes in ci](https://github.com/tenstorrent/tt-torch/actions/runs/14740119245/job/41376323173) and takes 5.5m. Nice!

PR #682, opened 2025-04-29
Fix PCC calculation logic and argument order for verification

sshonTT:
### Ticket
None

### Problem description
- The previous 'calculate_pcc' implementation, in edge cases where both tensors had constant but unequal values (e.g., all 0s vs. all 1s), the correlation result could still yield `1.0`, resulting in a false positive.
- (minor) The order of arguments in `calculate_pcc` and `calculate_atol` in `verify_against_golden()` was inconsistent and potentially confusing.

### What's changed
- Refactored `calculate_pcc` to:
  - Fallback to `np.allclose()` comparison if correlation fails or is NaN.
  - This avoids false positives when tensors have constant but non-equal values.
- Fixed argument order in `verify_against_golden()`

### Reference
- NumPy `corrcoef` documentation: https://numpy.org/doc/stable/reference/generated/numpy.corrcoef.html

### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


jameszianxuTT:
I think it would be nice to add a testcase for this in CI! This seems like a sneaky bug and it would be good to ensure this edgecase is always caught.

If you add a test function to the end of [test_basic.py](https://github.com/tenstorrent/tt-torch/blob/58ca9327bb92a7b63c12ead02a25722a84b4b4de/tests/torch/test_basic.py) to check this edgecase or any other file in `tests/torch` it would run in the "tests" check ([ref](https://github.com/tenstorrent/tt-torch/blob/58ca9327bb92a7b63c12ead02a25722a84b4b4de/.github/workflows/run-tests.yml#L104))

sshonTT:
> I think it would be nice to add a testcase for this in CI! This seems like a sneaky bug and it would be good to ensure this edgecase is always caught.
> 
> If you add a test function to the end of [test_basic.py](https://github.com/tenstorrent/tt-torch/blob/58ca9327bb92a7b63c12ead02a25722a84b4b4de/tests/torch/test_basic.py) to check this edgecase or any other file in `tests/torch` it would run in the "tests" check executed on every PR ([ref](https://github.com/tenstorrent/tt-torch/blob/58ca9327bb92a7b63c12ead02a25722a84b4b4de/.github/workflows/run-tests.yml#L104))

Sure, will do it.

Issue #670, opened 2025-04-28
device opened for every op after systemdesc changes slows down op-by-op flow


kmabeeTT:
### Describe the bug

I noticed op-by-op nightly is much slower over the weekend. Wasn't hard to look at CI logs and trace to changes from https://github.com/tenstorrent/tt-torch/pull/628 where the slowdown comes from the fact that device is re-opened for every op run instead of just once and re-used for every unique op run.  

Links to before/after on-Push CI for that commit merged Friday, take a look and you'll see this way more often:

```
2025-04-25 11:45:39.687 | INFO     | SiliconDriver   - Opened PCI device 0; KMD version: 1.27.1, IOMMU: disabled
2025-04-25 11:45:39.691 | INFO     | SiliconDriver   - Harvesting mask for chip 0 is 0x300 (physical layout: 0x5, logical: 0x300, simulated harvesting mask: 0x0).
```

Links, notice how time (gets bundled in compile time here) more than doubles:

Before:
https://github.com/tenstorrent/tt-torch/actions/runs/14663193173/job/41152398347

```
25/25 ops executed
Verified all ops ran in Mnist
Total Time - Compiling: 333.77 s, Running: 181.57 s, Golden: 0.31 s
PASSED
```

After:
https://github.com/tenstorrent/tt-torch/actions/runs/14663193173/job/41152398347

```
25/25 ops executed
Verified all ops ran in Mnist
Total Time - Compiling: 761.02 s, Running: 161.31 s, Golden: 0.29 s
PASSED
```

Please take a look and find way to get back to previous behavior where device is re-used.

FYI @jameszianxuTT 

### Steps to reproduce the issue

pytest -svv tests/models/mnist/test_mnist.py --op_by_op_torch

### Expected behavior

single device open instead of one per op

brataTT:
Closing, since op by op in nightlies seem to be 2x faster now.
e.g. RMGB: [before](https://github.com/tenstorrent/tt-torch/actions/runs/14746472263/job/41394969109) 1h56m -> [after](https://github.com/tenstorrent/tt-torch/actions/runs/14769598028/job/41480107154) 56m

PR #685, opened 2025-04-30
#670 Create system descriptor less often

brataTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/670

### Problem description
Creating system descriptor at every compilation slows down op by op flow.

### What's changed
- Create system descriptor before each test session
- Only create new one if a recent valid one does not already exist

### Checklist
- [x] Test against op by op on CI
    - [Run 3 op by op tests on main](https://github.com/tenstorrent/tt-torch/actions/runs/14756236486)
    - [Run 3 op by op tests with change](https://github.com/tenstorrent/tt-torch/actions/runs/14752764458)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@ import socket
 import os
 import json
 import shutil
+import tt_mlir
 
 global junitxml_path
 junitxml_path = None
@@ -35,6 +36,21 @@ def manage_dependencies(request):
     subprocess.check_call(
         [sys.executable, "-m", "pip", "uninstall", "-y"] + dependencies
     )
+
+
+@pytest.fixture(scope="session", autouse=True)
+def create_system_descriptor(request):
+    try:
+        tt_mlir.create_system_desc()
+    except Exception as e:
+        print(f"Failed to create system descriptor: {e}")
+
+    yield
+
+    try:
+        os.remove(os.getenv("SYSTEM_DESC_PATH"))
+    except OSError:
+        pass
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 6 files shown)


kmabeeTT:
Summarizing some results of this change... do we know how this compares to current main, but with ad0b5b8ee5f95fd3a8eb114ed4f49c551a325c2f reverted?  Can you kick off same experiment on CI with that change reverted to fill in below?

Model | Before| Today | ThisFix
-- | -- | -- | -- 
test_falcon.py |  | 626 s | 538 s
test_mnist.py |  | 144 s | 96 s
test_llama_7b.py |  | 722 s | 420 s

brataTT:
Thanks for arranging the runtimes into a summary @kmabeeTT 
Running CI for [ad0b5b8 reverted](https://github.com/tenstorrent/tt-torch/actions/runs/14759094313) and [not creating system desc](https://github.com/tenstorrent/tt-torch/actions/runs/14759119390) as references
I'll add runtimes from these to your comment

kmabeeTT:
Data on CI above is a bit wonky especially the falcon datapoint.  We talked offline.  I ran mnist op-by-op and falcon op-by-op in a loop 5x each taking average, with problematic change ad0b5b8  reverted (ie. Before) and with this change (ThisFix) and results are basically the same, so this looks good to me.

image

We also talked about how create/close device frequency in run_end_to_end() is unrelated to all these system desc changes, so we dont' need to change that here.

kmabeeTT (approved):
Back to previous speed, thanks!

PR #686, opened 2025-04-30
Revert remove flux models from nightly CI runs #629

jameszianxuTT:
### Ticket
#329 

### Problem description
Flux models caused runner diskspace issues due to broken caching resolved in #630. 

FYI @kmabeeTT that we're adding back a 2h test to nightly op by op coverage

### What's changed
This reverts commit b730e1d3ca55463f2aa6d7c8b2fe021f67f48b02.

### Checklist


jameszianxuTT:
As validation, I run the [Flux tests](https://github.com/tenstorrent/tt-torch/actions/runs/14760311544) in a single job on nightly and target runner `torch-n150-2` which should have ~30GB of free disk on host.

This should download 62GB of weights, to the model cache which is supposedly on a network drive. If those 60GB of weights somehow end up on host for some reason, then we'll know because this run dies and all of CI starts breaking.

jameszianxuTT:
Edit. Targeting a different low disk runner [torch-n300-5 instead](https://github.com/tenstorrent/tt-torch/actions/runs/14760474304) as torch-n150-2 will be in use for the next while.

jameszianxuTT:
Okay, this works! 
- Test time, 1h58
- Pre-run space: 45GB
- Post-run space: 44GB (drop is unimportant)
- No crashing or any errant behaviour
  - This suggests model caching is working without any unexpected side effects or significant leaks into host memory. 1GB drop might just be the husk of the container lying around in memory 
- The previous out-of-disk failures would've dropped this from 45 to -15GB free space and obliterated the runner. The fact that this test completed suggests that we should be ok to merge this into main and add back flux tests to our coverage

kmabeeTT (approved):
Thanks for testing! Yep.. Longer nightly is okay for now...

kmabeeTT:
There was a conflict in yml file, just resolved it.

jameszianxuTT:
@kmabeeTT Thanks for resolving it, just got home and will merge now

PR #690, opened 2025-04-30
Fix false positive assert pcc / atol reporting

jameszianxuTT:
### Ticket
None

### Problem description
When inverting the logic of verify_against_golden I messed up and now this common entry point to all our test verification is throwing false positives in some cases.

### What's changed
Correctly invert logic for not(A AND B). Thanks @jazpurTT for catching this!

### Checklist
- [-] New/Existing tests provide coverage for changes

### DIFF
--- a/tt_torch/tools/verify.py
+++ b/tt_torch/tools/verify.py
@@ -142,7 +142,7 @@ def verify_against_golden(
 
     # Now that all tensors are checked, print the final message
     if assert_pcc and assert_atol:
-        if not passed_pcc and passed_atol:
+        if not passed_pcc or not passed_atol:
             assert False, err_msg
     elif not assert_pcc and assert_atol:
         print("Ignoring PCC check\n")


jameszianxuTT:
I have prepared some tests to this verify_against_goldens to verify its correctness in the sixteen (!) permutations of:

(true pcc = low/high, true atol = low/high, asserting pcc = true/false, asserting atol = true/false)

I am unfortunately overcommitted and cannot push these tests in today; but I'd like to get this logic fix in ASAP

jameszianxuTT:
Thank you again @jazpurTT for raising this absolutely treacherous bug. I hope nothing is actually failing in our CI in a way we have not caught

kmabeeTT:
Oh man, many failures... does it make sense? Was the bug introduced recently or a long time ago? If recently, I am surprised by the failures...

jameszianxuTT:
@kmabeeTT This change was introduced with #622 on April 21, so 9 days. Not all the failures we're seeing are necessarily regressions, some may be mis-promoted or mis-asserting based on the false positive results. 

Will take some time to track down and root cause everything, luckily the window is not that large.

AleksKnezevic (approved):
Nice catch!

jameszianxuTT:
Sweeping nightly full model execute tests: https://github.com/tenstorrent/tt-torch/actions/runs/14767402705

Cursory check of all failing tests in onPR suggests two causes:
- I think most if not all failures are us **promoting things to assert on ATOL** that we didn't assert on before within the last few days. There are no apparent PCC failures in onPush, suggesting all failures there are ATOL-assert related (and come from asserts we just turned on since this error was introduced)

1. [[red/generality split](https://github.com/tenstorrent/tt-torch/pull/542/files#diff-99d8b031f58906f0401be0ec5700e49bb312feccff28c99133ec7ff469c8d996)](https://github.com/tenstorrent/tt-torch/pull/542/files#diff-99d8b031f58906f0401be0ec5700e49bb312feccff28c99133ec7ff469c8d996) + [[ATOL check added](https://github.com/tenstorrent/tt-torch/commit/efa96cd4850eb0353df8c502ef6ca22ce901adcb)](https://github.com/tenstorrent/tt-torch/commit/efa96cd4850eb0353df8c502ef6ca22ce901adcb) implicitly caused all models in this section to newly start asserting ATOL before. (5 days ago)
2. I also [[added ATOL checks to 5 models](https://github.com/tenstorrent/tt-torch/commit/e7ec6076c8012dec7becb9a14c26d630c97e2d11)](https://github.com/tenstorrent/tt-torch/commit/e7ec6076c8012dec7becb9a14c26d630c97e2d11) which weren't checking ATOL before (7 days ago)

jameszianxuTT:
As a resolution, I will disable all ATOL checking that was added within the past few days. 

I believe we actually had no ATOL checking on model tests before this bug was introduced, and the places we do `assert_atol` were accidentally added in the past 9 days.

kmabeeTT (approved):
Very nice, thanks for investigation and jumping on this - masterful work!

jazpurTT (approved):
good job tracking the failed models

jameszianxuTT:
Merging now. Summary of checks + fixes:
- Models that were explicitly marked with both `assert_pcc=True` and `assert_atol=True` since this bug was pushed ->  These should have failed immediately when run, but didn't due to this bug
- Models that were not marked with either asserts, resulting in constructor **default marking both to true**, then promoted to a context where they were tested for pcc/atol (eg. from op by op -> execute). Closer examination of GPT2 is needed, as it barely fails both PCC/ATOL

Issue #597, opened 2025-04-10
Add Falcon3 model variants to torch op by op flow


kmabeeTT:
Please add Falcon3 (4 variants) to nightly torch op-by-op flow. It is model_group=red model.

> Falcon3 1B
> Falcon3 3B
> Falcon3 7B
> Falcon3 10B

Source: https://huggingface.co/tiiuae/Falcon3-3B-Base

ddilbazTT:
```
    cc.enable_consteval = True
    cc.consteval_parameters = True
```
This causes a leaked semaphore. Upon monitoring htop, I noticed right before the program crash, memory rises to 10% and 103GB used. So, we cannot `consteval_parameters` without a crash. Therefore, we will only `enable_consteval`.

On an unrelated note though, seems like our SWP is at 100%. I am not sure if that's expected. Will talk to it and relay.

kmabeeTT:
Awesome job - and all the variants pass all ops through op-by-op flow so they are good candidates for promotion to compile-only list and full-model-execute list for the smaller models that fit on device... 

![Image](https://github.com/user-attachments/assets/3872054a-57e5-42fb-b930-dccd422fe7bb)

running a CI to try that now: https://github.com/tenstorrent/tt-torch/actions/runs/14777061079

ddilbazTT:
Yeah they will fail, I tested that

ddilbazTT:
@kmabeeTT Seems like this should have been mentioned in the PR. Let's try to move towards good PR process and including everything before merging; than rushing things to get in, as that's better practice. It would also prevent us from working after work hours.

kmabeeTT:
@ddilbazTT  - Please accept my apologies, didn't mean to promote after hours work - I agree with improved PR process, but I didn't mention it in PR since I assumed it was not ready, and adding to full-model-execute at same time is nice/bonus but not required as part of adding models (typically done as follow up PR once model is ready).

Issue #646, opened 2025-04-23
Wrapping debug messages in a debug mode


ddilbazTT:
### Is your feature request related to a problem? Please describe

Based on the changes proposed in PR #634 , I think it would be great to wrap debug messages in a debug mode. As Aleks suggested, 

> We already have a TT_TORCH_IR_LOG_LEVEL env var. We can either tie into that or use an additional LOGGER_LEVEL which would enable these.

### Describe the solution you'd like

The user would specify the `TT_TORCH_IR_LOG_LEVEL` variable to enable/ disable debug messages. Debug messages would not be printed by default.

kmabeeTT:
If `TT_TORCH_IR_LOG_LEVEL` is going to also dump IR and/or do other things, I don't think we want to use that to guard the ~useful op-by-op flow progress/debug prints, and a `LOGGER_LEVEL-like` env-var seems more appropriate to me.

PR #692, opened 2025-04-30
Add Falcon 3 1B, 3B, 7B, 10B to torch op-by-op #597

ddilbazTT:
### Ticket
#597 

### Problem description
Add Falcon3 1B, 3B, 7B, 10B Base

### What's changed
- Added Falcon3 1B, 3B, 7B, 10B Base models to torch op-by-op flow. Instruct mode is also available under model names, can run torch op-by-op but not added to nightly runs.
Note: `consteval_parameters` is disabled because it results in a memory related crash

### Checklist
- [X] Create `tests/models/falcon/test_falcon3.py`
- [X] Add `test_falcon3` to op-by-op nightly tests

### DIFF
--- a/tests/models/falcon/test_falcon3.py
+++ b/tests/models/falcon/test_falcon3.py
@@ -1,0 +1,77 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from tests.utils import ModelTester
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            self.model_name, padding_side="left", torch_dtype=torch.bfloat16
+        )
+        model = AutoModelForCausalLM.from_pretrained(
+            self.model_name, torch_dtype=torch.bfloat16
+        )
+        return model.generate
+
+    def _load_inputs(self):
+        self.prompt = "Hey, are you conscious? Can you talk to me?"
+        inputs = self.tokenizer(
+            self.prompt, return_tensors="pt", return_token_type_ids=False
+        )
+        return inputs
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "model_name",
+    [
+        "tiiuae/Falcon3-1B-Base",
+        "tiiuae/Falcon3-3B-Base",
+        "tiiuae/Falcon3-7B-Base",
+        "tiiuae/Falcon3-10B-Base",
+        "tiiuae/Falcon3-1B-Instruct",
+        "tiiuae/Falcon3-3B-Instruct",
+        "tiiuae/Falcon3-7B-Instruct",
+        "tiiuae/Falcon3-10B-Instruct",
+    ],
+)
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
+def test_falcon(record_property, model_name, mode, op_by_op):
+    cc = CompilerConfig()
+    cc.enable_consteval = True
+    # consteval_parameters is disabled because it results in a memory related crash
+    if op_by_op:
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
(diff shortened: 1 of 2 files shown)


kmabeeTT (approved):
Beautiful! It is missing model_group=red which I will quickly push to your branch now to fix it.  Would love to merge this before midnight so it's included in nightly - any objections?

Edit: I also pushed change to make TT_TORCH_DUMP_DEBUG=1 an env-var used in CompilerConfig's apply_environment_overrides() used to set CompilerConfig's dump_debug=True

tests/models/falcon/test_falcon3.py
```diff
@@ -0,0 +1,75 @@
...
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        model_group="red",
```
jameszianxuTT, line 66: N.B for myself 

I think the default behaviour of modeltester is to assert_atol=True if unspecified which is going to start breaking stuff on promotion

 I won't change it now since it doesn't affect this test as it doesn't run in EXECUTE and it will require a ~40m rerun of push CI; but I added it to my list of things to fix tomorrow

ddilbazTT: What's wrong with `assert_atol=True`? Could you provide some context? I would appreciate it, thanks :)

kmabeeTT: Some context.. see https://github.com/tenstorrent/tt-torch/pull/690 - basically, assert_atol=True was partially ignored past 9 days due to a typo/bug that just got fixed 5 mins ago in main, bunch of tests had it set to True and were now failing with ATOL mismatches after the fix, so James had to set them back to assert_atol=False for now.

ddilbazTT:
> Edit: I also pushed change to make TT_TORCH_DUMP_DEBUG=1 an env-var used in CompilerConfig's apply_environment_overrides() used to set CompilerConfig's dump_debug=True
I just looked at your addition. Seems like we set self.dumb_debug twice:

```
        if dump_intermediates:
            self.dump_debug = dump_intermediates == "DEBUG"
            self.dump_info = self.dump_debug or dump_intermediates == "INFO"
        dump_debug = os.environ.get("TT_TORCH_DUMP_DEBUG")
        if dump_debug and int(dump_debug):
            self.dump_debug = True
```
Is that what you intended? I think in #646 the idea was to use the existing environment variables. I don't prefer this addition, so I'll hold off adding this PR right now. I would appreciate if you could explain what was missing in dump_debug, maybe there's a better way to handle this.

kmabeeTT:
> > Edit: I also pushed change to make TT_TORCH_DUMP_DEBUG=1 an env-var used in CompilerConfig's apply_environment_overrides() used to set CompilerConfig's dump_debug=True
> 
> I just looked at your addition. Seems like we set self.dumb_debug twice:
> 
> ```
>         if dump_intermediates:
>             self.dump_debug = dump_intermediates == "DEBUG"
>             self.dump_info = self.dump_debug or dump_intermediates == "INFO"
>         dump_debug = os.environ.get("TT_TORCH_DUMP_DEBUG")
>         if dump_debug and int(dump_debug):
>             self.dump_debug = True
> ```
> 
> Is that what you intended? I think in #646 the idea was to use the existing environment variables. I don't prefer this addition, so I'll hold off adding this PR right now. I would appreciate if you could explain what was missing in dump_debug, maybe there's a better way to handle this.

Yes this was my intention, I did not see a way today with your change to enable `self.dump_debug` from cmd line using env-var and didn't want folks to have to enable "dump intermediates", so I needed to add something quick to make these messages still available without editing test source.  Let's merge this, to get these new tests in regression, can change/refine/improve logging and env-vars later if #646 needs more work.

ddilbazTT:
I think I'll just remove the parts addressing #646 from this PR, and enable automerge. We have two things we can use: `dump_info` and `dump_debug`. We just need to figure out how to use them smart. I don't want to merge unfinished code.

kmabeeTT:
> I'll still go ahead with removing the #646 fixes for now.

Got it, thanks for explanation, and sorry for trouble - sounds good to me.

tests/models/falcon/test_falcon3.py
ddilbazTT: I added `assert_atol=False` and `assert_pcc=False`.

ddilbazTT: Also, out of curiousity, what was the error? Was it happening on the CI side of things? Or on the superset side of things? I did run these models with nightly tests flow without an issue: https://github.com/tenstorrent/tt-torch/actions/runs/14759478890

kmabeeTT: The bug (accidentally not checking ATOL despite assert_atol=True) was visually caught, and when the bug was fixed the error is ATOL failure for models that are failing ATOL when running full-model-execute.   Example from [CI](https://github.com/tenstorrent/tt-torch/actions/runs/14767402705/job/41461695533#step:10:1502):

```
FAILED tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification_generality[full-eval-vgg13_bn] - AssertionError: ATOL of output 0: 1.3125, threshold: 0.01 ❌
```

These new Falcon3 models are not affected though becuase they are not yet running full-model-execute.

Issue #590, opened 2025-04-09
Bring up 5 models data parallel


nsmithtt:
This issue covers bringing up 5 (or more :) models in tt-torch that exercise runtime data parallel via subdevish mesh.  I think so far we've brought up resnet, which we should get into main and after that we should push for at least 4 more models.

There's also been talk of being able to make this part of the test harness itself so it could be toggled on for any model, that would be even more ideal.

hshahTT:
Closing this since https://github.com/tenstorrent/tt-torch/pull/846 added 31 RT data parallel models to the nightly CI.

PR #696, opened 2025-05-01
Add a method to get on-device runtime tensors as needed instead of on-host torch tensors

hshahTT:
### Ticket
[590](https://github.com/tenstorrent/tt-torch/issues/590)

### Problem description
Currently the `run()` function in bindings.cpp used to run the models on the device first gets the runtime tensors on the device, and immediately transfers them to host to be used as torch tensors. This prevents us from being able to run models asynchronously on multiple devices. This PR introduces a way to directly get the runtime tensors so they don't have to immediately be transferred back to host.

### What's changed

- **bindings.cpp:** Added 2 new functions
   - `run_async()`: Submits the inputs and model binary to device for execution and returns the runtime tensors (this method will return immediately, assuming all ops within the model are supported on-device)
   - `to_host()`: Takes a runtime tensor, blocks until it is populated, and converts it to a Torch tensor on-host.

- **backend.py, \*_backend.py, and executor.py:** Added a `async_mode` boolean flag within all backend functions to signal to the executor whether to return runtime tensors (via `run_async()`) or normal Torch tensors (via `run()`)

- **resnet50_data_parallel_async.py:** Added a demo to showcase how to use the new asynchronous mode of execution via runtime tensors and `to_host()`. Important Note: currently the ResNet model uses ops that aren't fully supported only on device, so the runtime tensors will automatically move to host without the explicit invocation of the `to_host()` function. So this example will not actually be faster than the normal demo and should be used for illustrative purposes only (at least until all ops are supported on device).

- **test_basic_async.py:** Added an async version of test_basic.py.

### Additional Notes
1. There are certain graphs that could be no-ops (i.e., the input and output tensors are the same). In such cases, `tt::runtime::submit()` will just return the input tensors as the output tensors. PyTorch then automatically deallocates the input tensor before `to_host()` is called, causing a segfault when calling `to_host()` with the now deallocated output tensor. In order to fix that, we added some special handling in `run_async()` that copies the output tensor to host as a new tensor.
2. Op by Op flows (i.e., when `CompilerConfig.compile_depth = COMPILE_OP_BY_OP or EXECUTE_OP_BY_OP`) are not supported in async mode.

#### Additional No-Op Refactor of passing options to the backend function
Previously we used a Python map to pass backend options to the backend function called in `torch.compile`. This PR also introduces a `BackendOptions` class as a replacement for this.

So code that was previously:
```
options = {}
options["compiler_config"] = cc
options["device"] = device

tt_model = torch.compile(model, backend=backend, dynamic=False, options=options)
```
Becomes:
```
options = BackendOptions()
options.compiler_config = cc
options.device = device

tt_model = torch.compile(model, backend=backend, dynamic=False, options=options)
```

### Checklist


tt_torch/csrc/bindings.cpp
```diff
@@ -116,12 +119,14 @@ std::vector<int64_t> as_vec_int64(std::vector<T> const &vec) {
   return result;
 }
 
-static torch::Tensor create_torch_tensor(const tt::runtime::Tensor &tensor,
-                                         const tt::runtime::TensorDesc &desc) {
```
jameszianxuTT, line 120: I think this function basically inlines the implementation of `tt::runtime::getTensorDesc`, though I don't think it really matters.

It is nice to strip the `tt::runtime::TensorDesc` as an input argument, and calculate it from within the `create_torch_tensor` function, thanks!

tt_torch/csrc/bindings.cpp
```diff
@@ -346,7 +364,13 @@ PYBIND11_MODULE(tt_mlir, m) {
   m.def("preprocess_inputs", &preprocess_inputs,
         "Preprocess inputs for execution");
   m.def("run", &run,
-        "Run the binary on pre-defined device and pre-processed inputs");
+        "Run the binary on pre-defined device and pre-processed inputs, "
+        "returning the final torch tensors on host");
+  m.def("get_runtime_tensors", &get_runtime_tensors,
```
AleksKnezevic, line 369: perhaps `run_async` is a more descriptive name than `get_runtime_tensors`

hshahTT: Updated the name

tests/torch/test_basic_async.py
```diff
@@ -0,0 +1,586 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
AleksKnezevic, line 1: should we just parameterize `test_basic` instead of duplicating it?

tt_torch/csrc/bindings.cpp
```diff
@@ -221,6 +226,52 @@ preprocess_inputs(tt::runtime::Device device, std::vector<at::Tensor> &inputs,
...
 }
 
+std::vector<tt::runtime::Tensor>
+run_async(tt::runtime::Device device, tt::runtime::Binary &binary,
+          uint32_t program_idx, std::vector<tt::runtime::Tensor> &rt_inputs) {
+  std::vector<tt::runtime::Tensor> rt_outputs =
+      tt::runtime::submit(device, binary, program_idx, rt_inputs);
+
+  bool is_nop_graph = true;
+  if (rt_inputs.size() == rt_outputs.size()) {
```
AleksKnezevic, line 236: Is this logic correct? If there are a different number of inputs/outputs it's not necessarily a nop graph. 

Maybe you could loop over the outputs and if you find the output in the inputs, then you run the code below to copy it, otherwise you just return the runtime outputs?

AleksKnezevic (changes requested):
This looks great! Please address the nop_graph comment and then I think we're good to go.

tt_torch/csrc/bindings.cpp
hshahTT: Good catch! It should've actually been:

```C++
// Assume that if the vectors' sizes don't match then the graph isn't a no-op
bool is_nop_graph = false;
if (rt_inputs.size() == rt_outputs.size()) {
  is_nop_graph = true;
  for (size_t i = 0; i < rt_inputs.size(); ++i) {
    if (rt_inputs[i].handle != rt_outputs[i].handle) {
      is_nop_graph = false;
      break;
    }
  }
}
if (is_nop_graph) {
  // Copy output tensors
}
```
Regarding your idea of looping over the outputs and finding the output in the input vector, I think it depends on how PyTorch behaves when some outputs are in the input vector, but not all.

For example:
if input_vec = [A, B, C] and output_vec = [A, D, E]

then will PyTorch deallocate A since it's present in both, or will it leave it alone since output_vec != input_vec? If it's the former then we'll have to go with your solution, otherwise my updated snippet above will work. I'll try to run some tests to see what the behaviour is.

I'd prefer not to have to iterate over the input vector for each output since that'll be a O(n^2) penalty to catch what should be a fairly rare case.

tests/torch/test_basic_async.py
hshahTT: I guess it depends on whether we consider async support to be part of the basic setup that users should test for after building tt-torch. I considered async mode to be a more "advanced" feature that doesn't necessarily signal anything wrong with the users' build or device setup, so I put it in a separate file. However I'm happy to move this to test_basic.py if you think it's needed there.

tt_torch/csrc/bindings.cpp
hshahTT: Update: I checked and PyTorch **will** deallocate an output tensor if it's the same as any input tensor (regardless of whether the other output tensors are different). I updated the `run_async` code to handle that case.

I also updated the `to_host` function to handle cases when the torch module returns a torch Tensor without applying any ops. I added two tests (`test_return_rt_tensor_and_torch_tensors` and `test_return_same_tensors`) for these cases.

AleksKnezevic (approved):

PR #697, opened 2025-05-01
Disable assert_atol for new tests and add test coverage for verify_against_golden

jameszianxuTT:
### Ticket
Related to #690 

### Problem description
- A couple items in Nightly failed due to accidental assert_atol.
- There is no test coverage for verify_against_golden

### What's changed
- Disable assert_atol for failing tests
- Add 16 test cases to cover all (actual PCC, actual ATOL, assert_pcc, assert_atol) cases in verify_against_golden

### Checklist

### DIFF
--- a/tests/torch/test_basic.py
+++ b/tests/torch/test_basic.py
@@ -639,14 +639,26 @@ def test_verify_against_golden_low_atol_high_pcc(assert_pcc, assert_atol):
 @pytest.mark.parametrize("assert_atol", [True, False])
 def test_verify_against_golden_high_atol_high_pcc(assert_pcc, assert_atol):
     # High ATOL and High PCC  (correlated, but numerically different tensors).
-    # Expected to fail ATOL if asserted, but pass if ATOL not asserted
+    # Expected to raise ATOL asssertion if asserting atol, but not raise if ATOL not asserted
 
     # True ATOL: 0.5
     # True PCC: 1
 
     golden_tensors = (torch.tensor([1.0, 2.0, 3.0]),)
     calculated_tensors = (torch.tensor([1.5, 2.5, 3.5]),)
-    try:
+
+    if assert_atol:
+        # Assert that an AssertionError is raised if we are asserting ATOL
+        # when comparing tensors with a true high ATOL
+        with pytest.raises(AssertionError) as e:
+            pccs, atols, passed_pcc, passed_atol = verify_against_golden(
+                golden_tensors,
+                calculated_tensors,
+                assert_pcc=assert_pcc,
+                assert_atol=assert_atol,
+                required_atol=0.1,
+            )
+    else:
         pccs, atols, passed_pcc, passed_atol = verify_against_golden(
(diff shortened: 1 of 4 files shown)


tests/torch/test_basic.py
```diff
@@ -611,3 +611,143 @@ def forward(self, x):
     verify_module(
         Basic(), input_shapes=[(32, 32)], input_range=(-2 * math.pi, 2 * math.pi)
     )
+
+
+@pytest.mark.parametrize("assert_pcc", [True, False])
```
jameszianxuTT, line 616: These 4 pytests can in theory be merged into a single one, but it would in my opinion make it much more confusing

AleksKnezevic (approved):

tests/torch/test_basic.py
jameszianxuTT: Said test results are now included in test_basic, with an example run in this workflow: https://github.com/tenstorrent/tt-torch/actions/runs/14775890580/job/41484337425?pr=697#step:12:450

ddilbazTT:
I was wondering if I could learn more about the issue. I was under the impression that designs failing due to bad pcc/ atol asserts was expected. 

I think it's somewhat related, but shouldn't we use `required_atol`/ `required_pcc`?  I was under the impression that `assert_atol`, `assert_pcc` should be turned off when atol/ pcc values are just too off. 

Again, I just don't think I understood the issue fully, so I would appreciate some more context, thanks!

kmabeeTT:
From talk to Aleks just now, we want to disable golden checking entirely when models are only running compile depth, like the failing pixtral test which is in e2e-compile list yet failing atol check.

jameszianxuTT:
> I was wondering if I could learn more about the issue. I was under the impression that designs failing due to bad pcc/ atol asserts was expected.

I think the flow for adding models should be like this:
1. Create model test with asserts initially turned OFF ->  `assert_pcc=False` and `assert_atol=False`
2. Add model test to appropriate workflows (most likely, op by op or compile e2e)
3. Once model gets promoted from initial workflow into EXECUTE, then:
4. Assertions will remain turned off to avoid failing the test
5. However, we automatically start monitoring EXECUTing model PCC/ATOL 
6. If monitoring from (5) shows initial good PCC/ATOL, we turn on those asserts 
7. If monitoring from (5) shows initial bad PCC/ATOL, keep the asserts off and create issues

---
Confusion currently results because we use EXECUTE tests to validate two things:
1. Were there any regressions in executability - i.e. new runtime exceptions, segfaults, dependency slips
2. Were there any regressions in accuracy (but only in the models we know could regress in accuracy; i.e. ones that have asserts on)

We have a well defined process to promote models to the EXECUTE tests; but a hazy one to promote them between subtests for (1 executability) and (2 accuracy). This implicit promotion\* also happens by changing the `assert_pcc/atol` in the constructor rather than moving tests around in the testlists.

\* "promotion" meaning, that we start testing for regressions in ways that we didn't before

PR #699, opened 2025-05-01
Add pcc/atol assert awareness to tag report

jameszianxuTT:
### Ticket
None

### Problem description
We don't capture if a model test does assert PCC / ATOL

### What's changed
Write out assertion states to model test report YAML in tags.

### Checklist

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -98,8 +98,12 @@ class ModelTester:
         print("[MODEL NAME]", model_name + model_name_suffix)
 
         self.record_tag_cache["required_pcc"] = self.required_pcc
-        self.record_tag_cache["required_atol"] = self.required_atol
-        self.record_tag_cache["relative_atol"] = self.relative_atol
+        # Avoid introducing conditional logic in DB to handle separate cases
+        self.record_tag_cache["required_atol"] = (
+            self.required_atol if self.required_atol is not None else self.relative_atol
+        )
+
+        # Postgres jsonb requires lowercased booleans. Pipeline transform from Python->XML->Python->JSON eventually lowercases these
         self.record_tag_cache["is_asserting_pcc"] = self.assert_pcc
         self.record_tag_cache["is_asserting_atol"] = self.assert_atol
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -97,6 +97,12 @@ class ModelTester:
 
         print("[MODEL NAME]", model_name + model_name_suffix)
 
+        self.record_tag_cache["required_pcc"] = self.required_pcc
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -97,6 +97,16 @@ def __init__(
 
         print("[MODEL NAME]", model_name + model_name_suffix)
 
+        self.record_tag_cache["required_pcc"] = self.required_pcc
+        # Avoid introducing conditional logic in DB to handle separate cases
+        self.record_tag_cache["required_atol"] = (
```
jameszianxuTT, line 102: From DB perspective it doesn't matter if this is relative or absolute. It's just a threshold to which we compare the atol calculated by verify_against_golden()

If a relative tolerance was specified, the RTOL is returned by verify_against_golden, stored as "tags->atol" and is compared by whatever is in the DB's "tags -> required_atol"

If an absolute tolerance was specified, the ATOL is returned by verify_against_golden and stored as "tags->atol"; and again compared with whatever is in the DB's "tags->required_atol"

This way the query stays the same and I don't think we care if tolerance is absolute or relative anyways, just that it doesn't exceed the specified required_*tol threshold

tests/utils.py
```diff
@@ -97,6 +97,16 @@ def __init__(
...
+            self.required_atol if self.required_atol is not None else self.relative_atol
+        )
+
+        # Postgres jsonb requires lowercased booleans. Pipeline transform from Python->XML->Python->JSON eventually lowercases these
```
jameszianxuTT, line 106: Verified this by inspecting [JSON output](https://github.com/tenstorrent/tt-torch/actions/runs/14780524235) from collect_data action triggered from a previous run of this workflow

```
"tags": {
                        "model_name": "Autoencoder (linear)",
                        "frontend": "tt-torch",
                        "required_pcc": 0.99,
                        "required_atol": 0.01,
                        "relative_atol": null,
                        "is_asserting_pcc": true,
                        "is_asserting_atol": true,
                        "pccs": [
                            0.9999970469895907
                        ],
                        "atols": [
                            0.0009765625
                        ],
                        "avg_pccs": 0.9999970469895907,
                        "min_pccs": 0.9999970469895907,
                        "max_pccs": 0.9999970469895907,
                        "avg_atols": 0.0009765625,
                        "min_atols": 0.0009765625,
                        "max_atols": 0.0009765625,
                        "pcc": 0.9999970469895907,
                        "atol": 0.0009765625,
                        "bringup_status": "PASSED"
                    }

kmabeeTT (approved):
Awesome, will be cool to see this in action for flagging tests that improved PCC

AleksKnezevic (approved):

PR #717, opened 2025-05-06
Nightly fail inspector

nsumrakTT:
### Ticket

### Problem description
Trying to bisect commits in order to find in which commit nightly tests started to fail.

### What's changed
Workflow will take test reports from previous nightly runs. It will select tests which fail in last nightly and it will run it on all commits happend between two nightly runs. After runs it will deduct in which commits which tests started to fail and attach report to the workflow run as well as send message to slack channel.

### Checklist

### DIFF
--- a/.github/workflows/fail-inspector.yml
+++ b/.github/workflows/fail-inspector.yml
@@ -1,0 +1,315 @@
+name: Failure Inspector
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: Nightly Tests
+    branches:
+      - main
+    types:
+      - completed
+
+
+jobs:
+  inspect:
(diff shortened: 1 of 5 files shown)


jameszianxuTT:
Is there a demo workflow somewhere to see the action of this feature?

jmcgrathTT (approved):

jameszianxuTT (approved):
Looks interesting, thanks for implementing this. Would be curious to see how this runs on nightlies, and there are a lot of directions we could iterate on this.

Doesn't look like anything that would break existing workflows.

nsumrakTT:
Unfortunately in tt-torch test logs were not well produced, so this PR should bring all that is required for action. So, it will need another few days until it can work properly.

PR #718, opened 2025-05-06
Enable IR,NN fusing option for tt-mlir

sshonTT:
### Ticket
NA

### Problem description
Since [patch in mlir](https://github.com/tenstorrent/tt-mlir/commit/f5c92c4c5069d2c53c875046b039b66246528fee), the IR, NN fusing option in tt-mlir is no longer enabled by default. This change led to a regression where two previously passing models began failing due to the missing optimization.

### What's changed
- Explicitly enabled IRFusing via `options.enableFusing = True` to restore expected behavior.
- Re-enabled PCC assertions for the two affected models(beit-large-patch16-224, opt) to ensure accuracy checks remain in place. These changes fix the regression and ensure future changes do not silently bypass PCC checks.


kmabeeTT (approved):
Awesome! Confirmed high PCC and tests passing, nice catch.

PR #722, opened 2025-05-07
Enable PCC assertions for models that already meet accuracy thresholds

sshonTT:
### Ticket
N/A

### Problem description
Some models are already achieving acceptable PCC metrics, but they were not asserting on PCC during test runs. This makes it hard to detect regressions if accuracy silently drops.

### What's changed
Enabled `assert_pcc=True` for a set of models confirmed to consistently meet PCC thresholds.

### Checklist
- [v ] Trigger nightly run [link](https://github.com/tenstorrent/tt-torch/actions/runs/14867312445)

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -26,8 +26,43 @@ jobs:
     secrets: inherit
     with:
       mlir_override: ${{ inputs.mlir_override }}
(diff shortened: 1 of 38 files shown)


kmabeeTT (approved):
Nicely done! Thanks!

PR #723, opened 2025-05-07
Jzx/efficientnet unmark

jameszianxuTT:
### Ticket
None

### Problem description
All efficientnet variants are marked as red, where only the b0 variant should be considered as such.

### What's changed
Split red/generality models into different modules.

### Checklist

### DIFF
--- a/tests/models/EfficientNet/test_EfficientNet.py
+++ b/tests/models/EfficientNet/test_EfficientNet.py
@@ -37,29 +37,7 @@ class ThisTester(ModelTester):
         return tfms(img).unsqueeze(0)
 
 
-@pytest.mark.parametrize(
-    "mode",
-    ["train", "eval"],
-)
-@pytest.mark.parametrize(
-    "model_name",
-    [
-        "efficientnet-b0",
-        "efficientnet-b1",
-        "efficientnet-b2",
-        "efficientnet-b3",
-        "efficientnet-b4",
-        "efficientnet-b5",
-        "efficientnet-b6",
-        "efficientnet-b7",
-    ],
-)
-@pytest.mark.parametrize(
-    "op_by_op",
-    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
-    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
-)
-def test_EfficientNet(record_property, model_name, mode, op_by_op):
+def run_EfficientNet_base(record_property, model_name, mode, op_by_op, _model_group):
     if mode == "train":
         pytest.skip()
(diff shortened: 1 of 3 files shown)


tests/models/EfficientNet/test_EfficientNet.py
```diff
@@ -77,7 +55,7 @@ def test_EfficientNet(record_property, model_name, mode, op_by_op):
         assert_atol=False,
         compiler_config=cc,
         record_property_handle=record_property,
-        model_group="red",
```
jameszianxuTT, line 80: @fzhengTT This change could have been an if/else on model_group, i.e. 
`model_group = 'red' if model_name == 'efficientnet-b0' else 'generality'`

However,  cicd_test_case has a unique constraint on `test_case_name`, `filepath`,  and the `group` field belongs to cicd_test_case. 

Therefore we cannot have variable group within a single `test_case_name` (i.e. pytest function) and therefore have to split one pytest function into two and rename all the tests.

jameszianxuTT: Also, attempting to have variable `group` in a single pytest function, based on the model name, led to self-overwriting anomalous behaviour where the `group` for all models tested in a single pytest function would appear to shift based on whichever was ingested last...that was a complex bug to root cause

kmabeeTT: Man, what an annoying change - thanks for making it.

jameszianxuTT: @kmabeeTT Per discussion with Frances, I think they can get in a database schema change quickly to remove the need for splitting pytest function to encode per-parameterization group assignment.

Created a DATA ticket and then I can revert this change + others with split pytest to just have if/else

kmabeeTT (approved):
from offline talk sounds like this might not be needed if data team change happens today. Approving for now anyways just in case since sounds like this will have short timetime.

jameszianxuTT:
Okay, doesn't look like data team change will happen immediately so I'll just push this now as a short term fix to get reporting correct, then revert + revert torchvision red/generality split tests once the schema is corrected.

fzhengTT:
@jameszianxuTT My PR is in review. Should be able to merge and deploy by EOD.

jameszianxuTT:
Okay, I stand corrected. Will wait to push then.

fzhengTT:
@jameszianxuTT My PR has been merged and changes are deployed.

jameszianxuTT:
Thanks to @fzhengTT for making and testing the schema change so quickly. Closing this PR in favor of #728

PR #728, opened 2025-05-07
Unsplit Red/Generality Model Tests and unmark efficientnet and swin variants as red

jameszianxuTT:
### Ticket
None

### Problem description
- Frances' change to add unique constraint to (pytest module, filetest, +group) in the cicd_test_case table allows us to mark varying model_group in a single pytest function. This caused some confusing data anomalies before.

### What's changed
- Unmark extraneous efficientnet and swin variants as red
- Recombine timm and torchvision red/generality split pytest modules into a single entry point, reverting #542 

### Checklist
- [x] Run CI verifier scan workflows locally to ensure renames were executed


jameszianxuTT:
I actually cannot verify this from the results of this PR alone... apparently in onPR we don't run any pair of testcases sharing the same pytest module but have different group markings. 

Launching a [separate test](https://github.com/tenstorrent/tt-torch/actions/runs/14895136400) + query afterwards to assess.

tests/models/torchvision/test_torchvision_image_classification.py
```diff
@@ -120,6 +119,8 @@ def run_torchvision_image_classification_test(
     assert_pcc = False if model_name in ["vit_h_14", "swin_b"] else True
     assert_atol = False
 
+    model_group = "red" if model_name == "swin_v2_s" else "generality"
```
kmabeeTT, line 122: Hmm... used to be just "swin" and now is more narrow "swin_v2_s"... is that right?

jameszianxuTT: Yes, I noticed in the spreadsheet that the specified variant was swin_v2_s; just wanted to get ahead of this.

kmabeeTT: Yeah, crap - I too noticed that right after I wrote this and did some digging.  Darn. Okay, good catch.

kmabeeTT (approved):
Looks good to me, thanks for (un)making changes.

jameszianxuTT:
Verification: Query on DB shows the following:
image

It was previously impossible due to database structure to have 1:N test_case_name:group relationship, only 1:1. 

Result of the query shows 1:N relationship is now possible, and validates @fzhengTT 's changes and this unsplit.

kmabeeTT:
Awesome - great work and turnaround, thanks @jameszianxuTT!

PR #740, opened 2025-05-08
Generate MLIR dumps and upload as nightly artifacts

ddilbazTT:
### Ticket
None

### Problem description
As we have more models compiling/running end-to-end, there's an increasing demand for MLIR dumps of these models at various stages (Stablehlo, TTIR, TTNN)

### What's changed
Added an environment variable that can be set to control at what stages will the MLIR be dumped. Note: this only works when models are run in full, not op-by-op. 

ex: `TT_TORCH_SAVE_MLIR=STABLEHLO pytest tests/models/resnet/test_resnet.py::test_resnet[full-eval]`
The `TT_TORCH_SAVE_MLIR` could be set to `STABLEHLO` `TTIR` `TTNN`. Or a list of them: `TT_TORCH_SAVE_MLIR=STABLEHLO,TTIR`

This environment variable is true by default, which means mlir files will be loaded as artifacts. It is set to false for on pr and on push flows, as that is unneeded overhead. Used merge-artifacts so that mlir dump artifacts are easy to navigate through.

### Checklist
- [X] Added `TT_TORCH_SAVE_MLIR` support to `backend.py` and `CompilerConfig`
- [X] Edited end-to end compile/ execute workflows to upload merged MLIR file artifacts on nightly run

### DIFF
--- a/.github/workflows/run-e2e-compile-tests.yml
+++ b/.github/workflows/run-e2e-compile-tests.yml
@@ -8,6 +8,11 @@ on:
         description: 'Docker image to use for build'
         required: true
         type: string
+      run-dump-mlir:
+        description: 'Dump MLIR files'
+        required: false
+        type: string
+        default: 'true'
   workflow_run:
     workflows: [Build]
     types: [completed]
@@ -89,6 +94,7 @@ jobs:
         echo "test_report_path_torch=report_torch_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
         echo "test_report_path_models=report_models_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
         echo "test_report_path_onnx=report_onnx_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
+        echo "mlir_dir=$(pwd)/model_mlir" >> "$GITHUB_OUTPUT" # Define the model_mlir directory
 
     - name: Git safe dir
       run: git config --global --add safe.directory ${{ steps.strings.outputs.work-dir }}
@@ -122,7 +128,7 @@ jobs:
       run: |
         source env/activate
 
-        TT_TORCH_COMPILE_DEPTH=TTNN_IR pytest --durations=50 -v -rf ${{matrix.build.tests}} \
+        TT_TORCH_SAVE_MLIR=STABLEHLO,TTIR,TTNN TT_TORCH_COMPILE_DEPTH=TTNN_IR pytest --durations=50 -v -rf ${{matrix.build.tests}} \
           --junit-xml=${{ steps.strings.outputs.test_report_path_models }} \
           2>&1 | tee pytest.log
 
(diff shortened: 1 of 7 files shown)


ddilbazTT:
I wanted to show what an example run looks like: [link](https://github.com/tenstorrent/tt-torch/actions/runs/14913738039)

There are in total 3 artifacts uploaded:
`model-mlir-compile` - generated from `.github/workflows/run-e2e-compile-tests.yml`
`model-mlir-execute-nightly` - generated from `.github/workflows/run-full-model-execution-tests-nightly.yml`
`model-mlir-execute-push` - generated from `.github/workflows/run-full-model-execution-tests.yml` 

compile vs execute denotes if the model can be run on device or can be lowered until TTNN at most. 

Thanks to @jameszianxuTT for suggesting `merge-artifact`. Otherwise, we would have uploaded dumps as an artifact per job, which includes several different models together. That would have been hard to navigate through.

I would appreciate while reviewing this PR, if you guys could let me know if you have any concerns with failing safe. 
@AleksKnezevic @kmabeeTT @jameszianxuTT @tapspatel

AleksKnezevic (approved):

.github/workflows/run-e2e-compile-tests.yml
```diff
@@ -139,3 +140,23 @@ jobs:
       with:
         name: test-reports-models-${{ matrix.build.runs-on }}-${{ matrix.build.name }}
         path: ${{ steps.strings.outputs.test_report_path_models }}
+
+    - name: Upload MLIR files
+      uses: actions/upload-artifact@v4
+      if: ${{ (success() || failure()) && inputs.run-dump-mlir == 'true' }}
```
kmabeeTT, line 146: Seems like `inputs.run-dump-mlir` is not defined or set by caller for this yml file, but maybe it should be if we want to have a clean one flag control to toggle it true/false (which looks like the intention by defining it as an input in other yml files).

.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -167,7 +173,7 @@ jobs:
       run: |
         source env/activate
 
-        pytest --durations=0 -v -rf ${{matrix.build.tests}} \
+        TT_TORCH_SAVE_MLIR=STABLEHLO,TTIR,TTNN pytest --durations=0 -v -rf ${{matrix.build.tests}} \
```
kmabeeTT, line 176: PR description says dumping only enabled in nightly which doesn't match this... I think you meant to set env-var based on the inputs.run-dump-mlir flag like I suggested earlier, here and in the other yaml files?  Seems mising.  Just put this a few lines higher next to the other envs is the right change I think:

```
TT_TORCH_SAVE_MLIR: ${{ inputs.dump-mlir == 'true' && 'STABLEHLO,TTIR,TTNN' || '' }}
```

kmabeeTT (changes requested):
Looks good - a couple setting/usage of inputs.run-dump-mlir seem off though?

.github/workflows/run-full-model-execution-tests.yml
ddilbazTT: I assumed it would be similar to how code-cov is right now

ddilbazTT: Yeah the description wasn't accurate, you're right

kmabeeTT (approved):

PR #744, opened 2025-05-09
Nightly fail inspector debug

nsumrakTT:
### Ticket

### Problem description
Fail inspector wrongly parsed test logs and could not find appropriate tests.

### What's changed
Better parsing is implemented.
Install artifacts are downloaded in install dir.
Wheel is installed without --force-reinstall

### Checklist


.github/workflows/fail-inspector.yml
```diff
@@ -220,13 +221,15 @@ jobs:
...
-            python runtest.py tests-${{matrix.build.runs-on}}.log 2>&1 | tee pytest.log
+            ls -lR tests
+            rt="tests/run_specific_tests.py"
+            echo "import pytest" >$rt
+            echo "import sys" >>$rt
+            echo "if __name__ == \"__main__\":" >>$rt
+            echo "    with open(sys.argv[1], \"r\") as fd:" >>$rt
+            echo "        test_list = fd.readlines()" >>$rt
+            echo "    sys.exit(pytest.main(test_list))" >>$rt
+            TT_TORCH_COMPILE_DEPTH=TTNN_IR python $rt tests-${{matrix.build.runs-on}}.log 2>&1 | tee pytest.log
```
jameszianxuTT, line 232: Sorry, just noticed this - TT_TORCH_COMPILE_DEPTH=TTNN_IR may change the nature of the test being run. 

I think at this point, tt-torch tests are designed such that this grep:
`grep '^FAILED ' "temp.log" | sed 's/^FAILED //; s/^\(.*\) - .*/\1/; s/^\(.*\]\).*/\1/' >>a-"$machine".log`

Which I think just reads the full test name of the failing pytest, does not fully determine the nature of the test being run. 

---

I'm not sure what a good solution to this would be though.

One approach is adding additional logic to determine, based on the specific job the failure is extracted from, which test to run, would be rather convoluted and fragile. 

A better approach with its own complexities is adding parameterizations to our pytests, such that the full name of a pytest in form:
`path/to/testfile::test_module[test-parameterization-list]` can fully determine the exact test being run. 

Would appreciate your thoughts @kmabeeTT  - should we add parameterizations to specify the compile depth rather than controlling through an env variable?

nsumrakTT: I think this would be very welcome change and necessary for nightly fail inspection. Alongside, I would also suggest:
- Use pytest mark to mark tests (eg. compile, eval ...)
- (optional) Instead of manual grouping (eg. compile_1, compile_2, eval_1, eval_2...) pytest is able to automatically group tests according to strategy. If strategy is duration, you can run test locally and check-in .test_durations which pytest can use for test grouping. In such way, you can select how many machines would execute whole test suite.

With these changes, we can streamline test into single test workflow. As an example you can check how it is done in tt-forge-fe where test-sub.yml is used to run tests for all scenarios (pr, push, nightly, weekly...) There are some pros to this approach:
- While building/debugging of code/tests developer can mark tests with his unique mark and run it independently of others in his own branch with versatile options (for example: https://github.com/tenstorrent/tt-forge-fe/actions/workflows/test.yml and click on "Run workflow").
- For each scenario you can select number of machines (groups) to run tests. Nightly can run on 10, nightly rerun can be set to run on 2, debugging on 1 etc.
- Single workflow is easier to maintain.

jameszianxuTT: I agree with the additional points, these are all CI tasks that should be done eventually:

1. Pytest marking would be a better way to specify meta test parameters than what we currently have, but would require some time perform and test this refactor.
2. We've considered pytest split and I implemented something similar for depth benchmarks. To avoid spam in the commit history, I have a separate branch used to store a `.test_durations-like file. I also wrote a custom load balancer similar to pytest split but with extra features like test quarantining and adjustable safety factor calculation; it was needed at the time for our benchmarking workflow 
  - This custom splitter or use of pytest split is an improvement that should be added to our CI at some point
3.  Thanks for the reference to forge fe's test.yml and test-sub.yml, that is interesting and I think our full execution and e2e compile tests **could be merged**. Op by op tests have a number of unique cleanup features so it may be harder to unify with other test yamls.
4. Adding parameterizations to specify test execution (eg. each compile depth could be a parameterization) would indeed be helpful, such that the executed test is fully characterized by its full pytest name.

---

That these are all features would improve QOL and maintainability, but I think not high priority for implementation at the moment as they aren't blocking anything. If we more clearly define the requirements from tt-torch side, can these tasks be offloaded to devex?

nsumrakTT: This is understandable. Surly, when you implement wanted changes we'll do the job of rearranging the workflows.

jameszianxuTT: >Adding parameterizations to specify test execution (eg. each compile depth could be a parameterization) would indeed be helpful, such that the executed test is fully characterized by its full pytest name.

This issue has been assigned in #755. 

For now, I think we could merge this PR just with the knowledge that fail inspector will give false negative results for compile-e2e tests, or alternatively hold off until #755 is fixed and merged.

nsumrakTT: Since there is a previous version of Nightly Fail Inspector already in the repo, I would prefer to merge this newer version, especially as it now uses shared TT actions (from tt-forge).

vmilosevic (approved):

PR #745, opened 2025-05-09
Use tt download-artifact

nsumrakTT:
### Ticket
slack

### Problem description
GitHub download-artifact is not stable.

### What's changed
Switch to using our custom implementation with retries and automatic untag

### Checklist

### DIFF
--- a/.github/workflows/generate-ttnn-md.yml
+++ b/.github/workflows/generate-ttnn-md.yml
@@ -47,20 +47,6 @@ jobs:
           name: ${{ inputs.spreadsheet_name }}
           path: ${{ steps.strings.outputs.work-dir }}/results
 
-      - name: Use build artifacts
-        uses: actions/download-artifact@v4
-        with:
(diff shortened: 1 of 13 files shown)


.github/workflows/generate-ttnn-md.yml
```diff
@@ -47,20 +47,6 @@ jobs:
           name: ${{ inputs.spreadsheet_name }}
           path: ${{ steps.strings.outputs.work-dir }}/results
 
-      - name: Use build artifacts
```
vmilosevic, line 50: You just removed download-artifact here, is that intended or you wanted to replace it with our action?

vmilosevic (approved):

.github/workflows/generate-ttnn-md.yml
nsumrakTT: Thanks, I forgot to add new download action.

Issue #746, opened 2025-05-09
op-by-op flow tests in nightly regression exit codes aren't captured properly reported as passing


kmabeeTT:
Seems like a bug not properly catching non-zero return status for op-by-op flow tests, so they are mistakenly marked as passing/green on CI. More details...

I noticed 3 models myteriously vanished from op-by-op XLS report on May 9th nightly, from visual tkdiff against May 8 nightly XLS, they are:

```
deepseek_ai_DeepSeek_V3
gliner_v2
mistralai_Ministral_8B_Instr
```

Just looking at gliner_v2 job we see:

```
[ 1 / 1 ] tests/models/gliner/test_gliner.py::test_gliner[op_by_op_torch-eval] PASSED
```

Yet it actually legit failed with ~known diskspace error (see artifact [full-logs-nightly-gliner](https://github.com/tenstorrent/tt-torch/actions/runs/14921201860/artifacts/3091575030)) and pytest process failed:

```
>           self._accessor.mkdir(self, mode)
E           OSError: [Errno 28] No space left on device: '/mnt/dockercache/huggingface/hub/models--urchade--gliner_largev2'

/usr/lib/python3.10/pathlib.py:1175: OSError
------------------------------ Captured log call -------------------------------
ERROR    huggingface_hub.file_download:file_download.py:1544 Could not cache non-existence of file. Will ignore error and continue. Error: [Errno 28] No space left on device: '/mnt/dockercache/huggingface/hub/models--urchade--gliner_largev2'
=============================== warnings summary ===============================
tt_torch/tools/utils.py:30
  /__w/tt-torch/tt-torch/tt_torch/tools/utils.py:30: DeprecationWarning: invalid escape sequence '\-'
    """

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/models/gliner/test_gliner.py::test_gliner[op_by_op_torch-eval] - OSError: [Errno 28] No space left on device: '/mnt/dockercache/huggingface/hub/models--urchade--gliner_largev2'
========================= 1 failed, 1 warning in 4.85s =========================
====== END LOG: tests/models/gliner/test_gliner.py::test_gliner[op_by_op_torch-eval] ========
```

Issue should be around here in `run-op-by-op-model-tests-nightly.yml`, maybe due to recent changes... looks like `exit_code=$?` is improperly placed (should happen right after pytest):

```
[...]

PR #747, opened 2025-05-09
Fix incorrect exit code sampling to avoid false positive op by op pass

jameszianxuTT:
### Ticket
#746 

### Problem description
Addition of fail logger caused the failure condition (sampling prev. command exit code) to start sampling the exit code of `sed` rather than `pytest`, which would always return 0 and show all op by op tests as passing regardless of exit code of the pytest.

### What's changed
Move exit code to be adjacent to pytest.

### Checklist


jameszianxuTT:
We expect to see the following workflow report a fail in ~20 minutes on the op by op tests, as I put an assert False in one op by op test to cause the pytest to exit !=0 

https://github.com/tenstorrent/tt-torch/actions/runs/14933997694

kmabeeTT (approved):
Looks great - thanks! I don't even need to see your side experiment to have high confidence in this fix!

Issue #332, opened 2025-02-20
stable diffusion 3.5


AleksKnezevic:
Add tests for stable diffusion 3.5 [medium](https://huggingface.co/stabilityai/stable-diffusion-3.5-medium) and [large](https://huggingface.co/stabilityai/stable-diffusion-3.5-large) to nightly op-by-op pipeline.

hshahTT:
Update: after talk with Legal it seems that we need to get a license from Stability AI before we can do anything with these models. I'll talk to Finance about procuring the license.

PR #749, opened 2025-05-09
Add Stable Diffusion 3.5 #332

ddilbazTT:
### Ticket
#332 

### Problem description
StableDiffusion (SD) 3.5 is a big model. The model uses SD3Transformer2DModel. I added two different tests - one that simply uses SD3Transformer2DModel and one that runs SD3.5. To get SD3.5 working, I made some model level optimizations and included memory cleanup in `pass_pipeline`. Even then, the
`tests/models/stable_diffusion/test_stable_diffusion_3_5.py` test can only run the medium variant op-by-op. It cannot run compile/ execute, and it cannot run the large variant.

Also added licensing material.

### What's changed
Added 2 test files for SD3.5 - one using the backbone transformer and one for the complete model. 
Edited workflow files to run SD3.5
Added some memory cleanup to backend

### DIFF
--- a/.github/workflows/run-full-model-execution-tests-nightly.yml
+++ b/.github/workflows/run-full-model-execution-tests-nightly.yml
@@ -115,6 +115,7 @@ jobs:
             # Approximately 40 minutes.
             runs-on: wormhole_b0, name: "eval_6", tests: "
                   tests/models/stable_diffusion/test_stable_diffusion_unet.py::test_stable_diffusion_unet[full-eval]
+                  tests/models/stable_diffusion/test_stable_diffusion_transformer.py::test_stable_diffusion_transformer[full-SD3.5-medium-transformer-eval]
             "
           },
           {
@@ -161,6 +162,9 @@ jobs:
                 tests/models/falcon/test_falcon3.py::test_falcon[full-tiiuae/Falcon3-7B-Base-eval]
                 tests/models/falcon/test_falcon3.py::test_falcon[full-tiiuae/Falcon3-10B-Base-eval]
                 tests/models/vovnet/test_vovnet_onnx.py::test_vovnet_onnx[full-eval]
+                tests/models/stable_diffusion/test_stable_diffusion_3_5.py::test_stable_diffusion_3_5[full-SD3.5-medium-eval]
+                tests/models/stable_diffusion/test_stable_diffusion_3_5.py::test_stable_diffusion_3_5[full-SD3.5-large-eval]
+                tests/models/stable_diffusion/test_stable_diffusion_transformer.py::test_stable_diffusion_transformer[full-SD3.5-large-transformer-eval]
             "
           },
         ]
--- a/.github/workflows/run-op-by-op-model-tests-nightly.yml
+++ b/.github/workflows/run-op-by-op-model-tests-nightly.yml
@@ -163,6 +163,13 @@ jobs:
               tests/models/flux/test_flux.py::test_flux[op_by_op_torch-flux_schnell-eval]
               tests/models/flux/test_flux.py::test_flux[op_by_op_torch-flux_dev-eval]
               "
+          },
+          {
+            runs-on: wormhole_b0, name: "stable-diffusion-v3.5", tests: "
+              tests/models/stable_diffusion/test_stable_diffusion_3_5.py::test_stable_diffusion_3_5[op_by_op_torch-SD3.5-medium-eval]
+              tests/models/stable_diffusion/test_stable_diffusion_3_5.py::test_stable_diffusion_3_5[op_by_op_torch-SD3.5-large-eval]
+              tests/models/stable_diffusion/test_stable_diffusion_transformer.py::test_stable_diffusion_transformer[op_by_op_torch-SD3.5-large-transformer-eval]
+              "
           }
         ]
     runs-on:
--- a/tests/models/stable_diffusion/test_stable_diffusion_3_5.py
+++ b/tests/models/stable_diffusion/test_stable_diffusion_3_5.py
@@ -1,0 +1,105 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+from tests.utils import ModelTester, skip_full_eval_test
+from diffusers import StableDiffusion3Pipeline, AutoencoderTiny
+
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        model_dict = dict(model_info_list)
+        model_path = model_dict[self.model_name]
+        pipe = StableDiffusion3Pipeline.from_pretrained(
+            model_path,
+            text_encoder_3=None,
+            tokenizer_3=None,
+            torch_dtype=torch.bfloat16,
+            low_cpu_mem_usage=True,
+        )
+        # memory optimization recommended by: https://huggingface.co/docs/diffusers/en/api/pipelines/stable_diffusion/stable_diffusion_3#tiny-autoencoder-for-stable-diffusion-3
+        pipe.vae = AutoencoderTiny.from_pretrained(
+            "madebyollin/taesd3", torch_dtype=torch.bfloat16, low_cpu_mem_usage=True
+        )
+        pipe.enable_attention_slicing()
+        return pipe
+
+    def _load_inputs(self):
+        prompt = [
+            "a photo of an astronaut riding a horse on mars",
+        ]
+
+        negative_prompt = ""
+        height = 512
+        width = 512
+        guidance_scale = 7.0
+        arguments = {
+            "prompt": prompt,
+            "negative_prompt": negative_prompt,
+            "height": height,
+            "width": width,
+            "guidance_scale": guidance_scale,
+        }
+
+        return arguments
+
+
+model_info_list = [
+    ("SD3.5-medium", "stabilityai/stable-diffusion-3.5-medium"),
+    ("SD3.5-large", "stabilityai/stable-diffusion-3.5-large"),
+]
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "model_info",
+    model_info_list,
+    ids=[model_info[0] for model_info in model_info_list],
+)
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
+def test_stable_diffusion_3_5(record_property, model_info, mode, op_by_op):
+    model_group = "red"
+    model_name, model_path = model_info
+
+    cc = CompilerConfig()
+    cc.enable_consteval = True
+    # consteval_parameters is disabled because it results in a memory related crash
+    if op_by_op:
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    skip_full_eval_test(
+        record_property,
+        cc,
+        model_name,
+        bringup_status="FAILED_RUNTIME",
+        reason="Cannot lower StableHLO --> TTIR : results/mlir_tests/stable_hlo/aten::round.decimals_0.mlir:3:10: error: failed to legalize operation 'stablehlo.round_nearest_even' - https://github.com/tenstorrent/tt-torch/issues/769",
+        model_group=model_group,
+    )
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        assert_atol=False,
+        assert_pcc=False,
+        model_group="red",
+    )
+    results = tester.test_model()
+    if mode == "eval":
+        image = results.images[0]
+        image.save("generated_image.png")
+
+    tester.finalize()
--- a/tests/models/stable_diffusion/test_stable_diffusion_transformer.py
+++ b/tests/models/stable_diffusion/test_stable_diffusion_transformer.py
@@ -1,0 +1,128 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+from tests.utils import ModelTester, skip_full_eval_test
+from diffusers import (
+    StableDiffusion3Pipeline,
+    FlowMatchEulerDiscreteScheduler,
+    SD3Transformer2DModel,
+)
+
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        model_dict = dict(model_info_list)
+        model_path = model_dict[self.model_name]
+        self.pipe = StableDiffusion3Pipeline.from_pretrained(
+            model_path,
+            torch_dtype=torch.bfloat16,
+            low_cpu_mem_usage=True,
+        )
+        self.scheduler = FlowMatchEulerDiscreteScheduler.from_pretrained(
+            model_path, subfolder="scheduler"
+        )
+        self.transformer = SD3Transformer2DModel.from_pretrained(
+            model_path, subfolder="transformer", torch_dtype=torch.bfloat16
+        )
+        return self.transformer
+
+    def _load_inputs(self):
+        prompt = "A futuristic cityscape at sunset"
+        num_images_per_prompt = 1
+        height = 512
+        width = 512
+        (
+            prompt_embeds,
+            negative_prompt_embeds,
+            pooled_prompt_embeds,
+            negative_pooled_prompt_embeds,
+        ) = self.pipe.encode_prompt(
+            prompt=prompt,
+            prompt_2=prompt,  # Using the same prompt for all encoders for simplicity
+            prompt_3=prompt,
+            num_images_per_prompt=num_images_per_prompt,
+            do_classifier_free_guidance=False,  # For simplicity in this direct call
+        )
+
+        num_channels_latents = self.transformer.config.in_channels
+        latents = self.pipe.prepare_latents(
+            batch_size=num_images_per_prompt,
+            num_channels_latents=num_channels_latents,
+            height=height,
+            width=width,
+            dtype=prompt_embeds.dtype,
+            device=None,
+            generator=None,
+        )
+
+        self.scheduler.set_timesteps(28)
+        timestep = self.scheduler.timesteps[0].expand(latents.shape[0])
+        arguments = {
+            "hidden_states": latents,
+            "timestep": timestep,
+            "encoder_hidden_states": prompt_embeds,
+            "pooled_projections": pooled_prompt_embeds,
+            "joint_attention_kwargs": {},
+            "return_dict": False,
+        }
+        return arguments
+
+
+model_info_list = [
+    ("SD3.5-medium-transformer", "stabilityai/stable-diffusion-3.5-medium"),
+    ("SD3.5-large-transformer", "stabilityai/stable-diffusion-3.5-large"),
+]
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "model_info",
+    model_info_list,
+    ids=[model_info[0] for model_info in model_info_list],
+)
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
+def test_stable_diffusion_transformer(record_property, model_info, mode, op_by_op):
+    model_group = "red"
+    model_name, model_path = model_info
+
+    cc = CompilerConfig()
+    cc.enable_consteval = True
+    cc.consteval_parameters = True
+    if op_by_op:
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    skip_full_eval_test(
+        record_property,
+        cc,
+        model_name,
+        bringup_status="FAILED_RUNTIME",
+        reason="Model compilation exceeds CI timeout limit (240 minutes), compiles e2e locally. Model execution fails with `Not enough space to allocate 47316992 B DRAM buffer across 12 banks.`",
+        model_group=model_group,
+        model_name_filter="SD3.5-large-transformer",
+    )
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        assert_atol=False,
+        assert_pcc=False,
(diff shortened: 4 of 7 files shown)


.github/workflows/run-full-model-execution-tests-nightly.yml
```diff
@@ -110,6 +110,7 @@ jobs:
             # Approximately 40 minutes.
             runs-on: wormhole_b0, name: "eval_6", tests: "
                   tests/models/stable_diffusion/test_stable_diffusion_unet.py::test_stable_diffusion_unet[full-eval]
+                  tests/models/stable_diffusion/test_stable_diffusion_transformer.py::test_stable_diffusion_transformer[full-stable-diffusion-v3.5-medium-eval]
```
kmabeeTT, line 113: We can do a follow up PR to address this, but if the plan is to make visible that we have the other variants existing but not yet running, then the idea is to add them here as well , with the skip_full_eval_test() in each test with reasons and bringup_status, I see one of them was added, but not used since not added here.

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
```diff
@@ -0,0 +1,127 @@
...
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    skip_full_eval_test(
+        record_property,
+        cc,
+        model_name,
+        bringup_status="FAILED_RUNTIME",
+        reason="Model is too large to fit on single device during execution.",
```
kmabeeTT, line 111: This is probably eventually true (since seems like an 8B param model?), but is this really the failure reason, or something else?  When we encounter this, it's of this type of error at runtime (does it make it to runtime)?

```
Out of Memory: Not enough space to allocate 165183488 B DRAM buffer across 12 banks, where each bank needs to store 13766656 B
```

kmabeeTT (commented):
Nice changes! Left some comments.

Do you have more details on the failure reasons here other than "cannot run" we can document in this PR to help save time in the future?

> Even then, the tests/models/stable_diffusion/test_stable_diffusion_3_5.py test can only run the medium variant op-by-op. It cannot run compile/ execute, and it cannot run the large variant.


My attempt to document the 4 variants added here and where they are added:

```
tests/models/stable_diffusion/test_stable_diffusion_transformer.py medium => full-model-execute
tests/models/stable_diffusion/test_stable_diffusion_transformer.py large => op-by-op
tests/models/stable_diffusion/test_stable_diffusion_3_5.py medium => op-by-op
tests/models/stable_diffusion/test_stable_diffusion_3_5.py large => not run anywhere
```

.github/workflows/run-op-by-op-model-tests-nightly.yml
```diff
@@ -163,6 +163,16 @@ jobs:
...
               tests/models/flux/test_flux.py::test_flux[op_by_op_torch-flux_dev-eval]
               "
+          },
+          {
+            runs-on: wormhole_b0, name: "stable-diffusion-v3.5", tests: "
+              tests/models/stable_diffusion/test_stable_diffusion_transformer.py::test_stable_diffusion_transformer[op_by_op_torch-stable-diffusion-v3.5-large-eval]
+              "
+          },
+          {
+            runs-on: wormhole_b0, name: "stable-diffusion-v3.5", tests: "
```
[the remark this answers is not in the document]
jameszianxuTT: I think that has a chance of blowing up a couple of actions like fetch job-id and upload artifact, which enforce non-duplication policies and pass in as input the `matrix.build.name`; agree to differentiate these names

jameszianxuTT (approved):
Other than duplicate group naming, I think this looks fine. Thanks!

tt_torch/dynamo/passes.py
```diff
@@ -154,4 +157,5 @@ def pass_pipeline(gm: torch.fx.GraphModule, example_inputs, compiler_config):
 
     # Need to run shape_prop again to populate tensor_meta
     run_shape_prop(program.graph_module, constant_inputs + example_inputs)
+    gc.collect()
```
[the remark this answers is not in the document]
ddilbazTT: This and also in constant_fold, I delete the unused graph.

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
ddilbazTT: This is the exact error it got:
`Out of Memory: Not enough space to allocate 47316992 B DRAM buffer across 12 banks, where each bank needs to store 3949568 B` [link](https://github.com/tenstorrent/tt-torch/actions/runs/14888956747/job/41817359246)

.github/workflows/run-op-by-op-model-tests-nightly.yml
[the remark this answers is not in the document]
ddilbazTT: Oh sorry, yeah i overlooked this. I'll append one of them with -transformer

.github/workflows/run-full-model-execution-tests-nightly.yml
ddilbazTT: Yeah, could you let me know what exactly I should do? I had a feeling I needed to do something like this for the large one, but if you could write it out I'd appreciate it.

.github/workflows/run-op-by-op-model-tests-nightly.yml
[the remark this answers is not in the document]
ddilbazTT: Or actually I'll just merge them in the same group, what am I saying

tests/models/stable_diffusion/test_stable_diffusion_3_5.py
```diff
@@ -0,0 +1,94 @@
...
+            "height": height,
+            "width": width,
+            "guidance_scale": guidance_scale,
+        }
+
+        return arguments
+
+
+model_info_list = [
+    ("stable-diffusion-v3.5-medium", "stabilityai/stable-diffusion-3.5-medium"),
```
kmabeeTT, line 50: Another problem I just noticed, these have same model names ie "stable-diffusion-v3.5-medium" as the _transformer.py test, so they will clash in reporting in various places (XLS, database), I think.  There is also a character limit in excel sheet names so we only show 28 characters that we have to work with.  Can we use short names something like these as the model names (key of model_info_list) in these tests?

```
SD3.5-medium
SD3.5-large
SD3.5-medium-transformer
SD3.5-large-transformer
```

.github/workflows/run-full-model-execution-tests-nightly.yml
kmabeeTT: Sure, I don't have the failing reasons you've encountered so I can't say for sure about the content, but the idea is to follow that commit I shared https://github.com/tenstorrent/tt-torch/commit/48c2ca89c8f30ee85d6b4766b6a77abbdec03671 as in:

1. Add `skip_full_eval_test()` targeting each case that has issues we want to pursue with details on current failing reason as best we know today and correct `bringup_status` (will be used for reporting where this model is failing). Should be placed before model is loaded, right after `CompilerConfig` is created, just like you did in this PR already for 1 of the tests (looks great). Available bringup statuses:

```
tests/utils.py:        bringup_status: The bringup status of the test (FAILED_FE_COMPILATION, FAILED_TTMLIR_COMPILATION, FAILED_RUNTIME, INCORRECT_RESULT, PASSED)
```

2. add the test to `run-full-model-execution-tests-nightly.yml` group called `skip_tests_red_models` where all the other similiar tests are listed that have issues.


For bullet 1, can filter by single string model name, list of model names, or no filter at all, example:

```
    skip_full_eval_test(
        record_property,
        cc,
        model_name,
        bringup_status="FAILED_RUNTIME",
        reason="Model is too large to fit on single device during execution.",
        model_group=model_group,
        model_name_filter=[
            "mistralai/Mistral-7B-v0.1",
            "mistralai/Ministral-8B-Instruct-2410",
        ],
    )
```

If we care about all 4 variants, since 1 is already running, would expect to see `skip_full_eval_test()` covering 3 different failures, and 3 entries added to `skip_tests_red_models` group.

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
kmabeeTT: Okay cool, looks good, thx for link and confirming!

tests/models/stable_diffusion/LICENSE.txt
```diff
@@ -0,0 +1,25 @@
+STABILITY AI COMMUNITY LICENSE AGREEMENT Last Updated: July 5, 2024
```
AleksKnezevic, line 1: Could have sworn I already commented here, but I don't see my comments... Anyway, we're not reproducing their code or their model. We don't need a licence.

tests/models/stable_diffusion/NOTICE.txt
```diff
@@ -0,0 +1,3 @@
+This repository includes Stability AI Model.
```
AleksKnezevic, line 1: Ditto on the notice.

ddilbazTT:
I wanted to post an update. I ran the designs in a nightly flow, and except for one, none of them executed. However, the nightly runs didn';t really capture why they weren't executing. When I ran `pytest -s tests/models/stable_diffusion/test_stable_diffusion_3_5.py::test_stable_diffusion_3_5[full-SD3.5-large-eval]` locally, I noticed it throws the following error: 
```
E       RuntimeError: TT_THROW @ /localdev/ddilbaz/tt-torch-2/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/program/program.cpp:114: tt::exception
E       info:
E       Failed to generate binaries for tilize TT_THROW @ /localdev/ddilbaz/tt-torch-2/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/jit_build/build.cpp:69: tt::exception
E       info:
E       Failed to open compile failure log file /home/ddilbaz/.cache/tt-metal-cache/7e6553092d/4097/kernels/tilize/13607550924644667372/trisc0/build.log
```

AleksKnezevic:
> I discovered that the reason SD3.5-medium is failing during execute is this:
> 
> ```
> python3.10: /localdev/ddilbaz/tt-torch-2/tt_torch/csrc/bindings.cpp:65: tt::target::DataType torch_scalar_type_to_dt(torch::ScalarType): Assertion `false && "Unsupported scalar type"' failed.
> Fatal Python error: Aborted
> ```
> 
> I'll try to use gdb and see why it's hitting this.

Fails op by op or e2e?

ddilbazTT:
> Fails op by op or e2e?

e2e

kmabeeTT:
> I'll try to use gdb and see why it's hitting this.

If full model execute no need to debug here, goal is op by op tests, skipped during full model execute for tracking purposes ideally with some minimal reason for reference.

ddilbazTT:
@kmabeeTT this could be a bug though, `torch_scalar_type_to_dt` shouldn't really hit an uncovered case when the model is using bfloat16. I would like to debug further why this is happening

AleksKnezevic:
There's an intermediate value that's not bfloat16, it's fine. I agree with @kmabeeTT that we can just add the op by op test and continue.

kmabeeTT:
Also remember that passing op-by-op flow exists to debug, and is what typically shows compiler/runtime failures per op along with repro details for single ops.  To use it, need to generate and open up the op-by-op flow generated XLS to see details, in this case the linked CI [run](https://github.com/tenstorrent/tt-torch/actions/runs/14983356034/job/42092715437) above accidentally had the XLS generation step commented out though. It's useful to keep, this part:

```
  download-report:
    if: success() || failure()
    needs: test_op_by_op
    uses: ./.github/workflows/generate-model-report.yml
    secrets: inherit
```

Upon downloading the artifacts and generating it myself manually (steps below) I see a few small problems... 

```
 4658  13/05/25 17:35:20 python3 results/download_artifacts.py --run-id 14983356034 --filter reports
 4661  13/05/25 17:35:37 cd nightly-tests_artifacts_20250512_run_id_14983356034
 4662  13/05/25 17:35:41 for f in `find *.tar`; do tar -xvf $f; done;
 4667  13/05/25 17:36:17 mv models ../results
 4668  13/05/25 17:36:18 cd ..
 4671  13/05/25 17:36:30 python3 results/parse_op_by_op_results.py

Processing json (1/4) stabilityai_stable_diffusion
Processing json (2/4) stabilityai_stable_diffusion_1
Processing json (3/4) stabilityai_stable_diffusion_2
Processing json (4/4) stabilityai_stable_diffusion_3
Creating summary for 4 models
```

Screenshot and attached XLS ([models_op_per_op_sd3.5.xlsx](https://github.com/user-attachments/files/20193870/models_op_per_op_sd3.5.xlsx)):

image


1. The model names (with cutoff at 28 characters) still appear the same, my previous suggestion wasn't enough I just realized, need to find some way to have `ModelTester` have `self.model_name` set to to the abbreviated `SD3.5-medium` version and the model loading use the longer string `stabilityai/stable-diffusion-3.5-medium`. 
2. Something is off here in this run, and 99% of the error messages on per-model sheets aren't able to be extracted from stderr. Not sure what is up there, never seen this before... not seen in nightly regression,  so I wonder if something wrong with the flow or model, or branch.  Medium SD3.5 test is quick to run (15 mins on CI) in op-by-op flow to reproduce, I can give some tips to make it even shorter by targeting just a specific op type.
3. For my earlier point, `stabilityai_stable_diffusion_2 and stabilityai_stable_diffusion_3` must be the full (non-transformer) models you mentioned have real errors, so they do not look close at all to be ready for full-model-execute running and debug, which is okay.

ddilbazTT:
@kmabeeTT I have been running again and I still can't find the actual error why SD3.5 (the complete model) fail. It's like every time I run I get a different error. Even if not debugging, I just want to note down the exact reason why they're failing in the tests themselves. For exampl,e when I reran SD3.5 medium, I got the following error this time:

```
E       RuntimeError: TT_THROW @ /localdev/ddilbaz/tt-torch-2/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/program/program.cpp:114: tt::exception
E       info:
E       Failed to generate binaries for tilize TT_THROW @ /localdev/ddilbaz/tt-torch-2/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/jit_build/build.cpp:67: tt::exception
E       info:
E       trisc0 build failed. Log: /localdev/ddilbaz/tt-torch-2/install/tt-metal/runtime/sfpi/compiler/bin/../lib/gcc/riscv32-unknown-elf/12.4.0/../../../../riscv32-unknown-elf/bin/ld:/localdev/ddilbaz/tt-torch-2/install/tt-metal/runtime/h
w/toolchain/wormhole/kernel_trisc0.ld:18: non constant or forward reference address expression for section .text
E       collect2: error: ld returned 1 exit status
```

kmabeeTT:
> Even if not debugging, I just want to note down the exact reason why they're failing in the tests themselves. For example, when I reran SD3.5 medium, I got the following error this time:

op-by-op flow is useful here too... especially in this case where model is failing for many reasons (multiple depths in compiler, and also at runtime) I have been picking the assertion errors seen from runtime/tt-metal (or highest compile depth) since they typically have the longest path/time to fix (and the frontend/compiler errors end up getting fixed sooner). Try running the SD3.5 medium test in op by op flow locally and seeing if you get some real errors reported on in the generated .json files and XLS.

kmabeeTT:
Follow up to my previous points 1,2 to 

1. model name cutoff, I don't see quick and easy and safe fix for this (modifying model_name to ModelTester wil impact reported models to superset which we don't want to do), it's purely XLS generation limited,  so I vote to leave it for now, revisit afterwards with fix for all models once discussed and some plan figured out to address .json/xls generation.
2. Your CI run for op-by-op flow at least had machine issues of some kind, since when I run SD3.5 med test locally, so far all ops are running fine (depth 7), certainly more than the 47 successful ops in the screenshot...

```
gron results/models/tests/models/stable_diffusion/test_stable_diffusion_3_5_py__test_stable_diffusion_3_5_op_by_op_torch_SD3_5_medium_eval___call__unique_ops.json | grep status | grep "= 7" | wc -l
132
```

kmabeeTT:
> Yeah, this is why I'm rerunning locally and on CI. At the very least, I want to be sure I understand when exactly the models fail during e2e. So far, results are non-deterministic.

Locally is good idea, I agree - it worked great for me - don't know what is up with CI not showing actual errors (pytest cmd and logging did change a bit in past week, maybe related), but I think that might be seperate issue from this PR.

The op-by-op flow results are important to consider first though, because if model cannot compile all ops it generally doesn't make sense to try and run full model execute since compiling is prerequisitie to running in `full-eval` mode.  In this case, the op-by-op flow for SD3.5 (med and large) reports this compile error (on stablehlo -> ttir conversion) for 1 op (`aten::round.decimals`, all other ops pass execute in my local run from yesterday, see screenshot) and e2e compile (`TT_TORCH_COMPILE_DEPTH=TTNN_IR`) also reports the same error, pasted here:

```
error: failed to legalize operation 'stablehlo.round_nearest_even'
```

That would be the reason to use for what is blocking full model execute since it is blocking compile. Note that the original SD1.5 model we've had for a while is hitting that compiler error above too lately. I opened ticket for this here: https://github.com/tenstorrent/tt-torch/issues/769 for debug and reference it below.

So for SD3.5 med/large models, this should be the exact line (inspired by similar examples in other tests) you want to add to `test_stable_diffusion_3_5.py` for tracking purposes alongside adding them to`skip_tests_red_models` group in `.github/workflows/run-full-model-execution-tests-nightly.yml`, and tests added to op-by-op nightly list:

```
    skip_full_eval_test(
        record_property,
        cc,
        model_name,
        bringup_status="FAILED_TTMLIR_COMPILATION",
        reason="results/mlir_tests/stable_hlo/aten::round.decimals_0.mlir:3:10: error: failed to legalize operation 'stablehlo.round_nearest_even' - https://github.com/tenstorrent/tt-torch/issues/769",
        model_group=model_group,
    )
```

For reference, attaching new XLS [models_op_per_op_sd3.5_local.xlsx](https://github.com/user-attachments/files/20211298/models_op_per_op_sd3.5_local.xlsx) where I ran all 4 variants without any non-deterministic issues showing the real failures:

image
image

All this said, I think we should cleanup this branch, apply changes above and merge today, what do you think?

kmabeeTT:
About model name issues... after more thought, I realized an additional problem: right now they are duplicated between the two different python tests which would clash during reporting to superset. So I put together a small change to fix both that problem and the "names are too long for XLS" issue at 15214541457f68868c5f1310f16273e6aab83007, you can cherry pick or adjust as needed to solve problem, please take a look.

tt_torch/dynamo/passes.py
```diff
@@ -103,11 +104,13 @@ def bypass_dtype_promotion(gm, compiler_config):
 
 
 def constant_fold(gm):
-    gm = const_fold.split_const_subgraphs(gm)
-    gm.run_folding()
+    split_gm = const_fold.split_const_subgraphs(gm)
```
AleksKnezevic, line 107: If the idea is to delete the original gm before you run folding (if this is to solve potential out of memory issues) I would prefer something like:
```python
gm = const_fold.split_const_subgraphs(gm)
gc.collect()
gm.run_folding()
```

tt_torch/dynamo/passes.py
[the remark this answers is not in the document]
AleksKnezevic: Is this really needed, everything goes out of scope after the return?

tt_torch/dynamo/passes.py
```diff
@@ -104,13 +104,12 @@ def bypass_dtype_promotion(gm, compiler_config):
 
 
 def constant_fold(gm):
-    split_gm = const_fold.split_const_subgraphs(gm)
-    del gm
-    split_gm.run_folding()
-
-    split_gm.graph.eliminate_dead_code()
+    gm = const_fold.split_const_subgraphs(gm)
+    gc.collect()
```
AleksKnezevic, line 108: This collect is needed?

ddilbazTT: We need some memory cleanup here to not crash due to leaked semaphore. My reasoning was that `gm = const_fold.split_const_subgraphs(gm)` assigns a new value to gm without releasing memory occupied by the previous gm.

AleksKnezevic: sounds good. I guess the garbage collector happens to not run before run_folding.

AleksKnezevic (approved):

kmabeeTT (changes requested):
Hmm some unexpected changes here to comment out tests in run-e2e-compile-tests.yml?

tests/models/stable_diffusion/test_stable_diffusion_3_5.py
```diff
@@ -0,0 +1,108 @@
...
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    skip_full_eval_test(
+        record_property,
+        cc,
+        model_name,
+        bringup_status="FAILED_RUNTIME",
+        reason="Cannot lower StableHLO --> TTIR",
+        model_group=model_group,
+    )
```
kmabeeTT, line 92: bringup_status is wrong here, plz use this (I merged the text from my previous comment with your reason here):

```
    skip_full_eval_test(
        record_property,
        cc,
        model_name,
        bringup_status="FAILED_TTMLIR_COMPILATION",
        reason="Cannot lower StableHLO --> TTIR : results/mlir_tests/stable_hlo/aten::round.decimals_0.mlir:3:10: error: failed to legalize operation 'stablehlo.round_nearest_even' - https://github.com/tenstorrent/tt-torch/issues/769",
        model_group=model_group,
    )
```

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
```diff
@@ -0,0 +1,131 @@
...
+            "timestep": timestep,
+            "encoder_hidden_states": prompt_embeds,
+            "pooled_projections": pooled_prompt_embeds,
+            "joint_attention_kwargs": {},
+            "return_dict": False,
+        }
+        return arguments
+
+
+model_info_list = [
```
kmabeeTT, line 78: Not broken, but I don't love that short model name to weights name is listed here (not used) and then duplicated above in model_dict. The change I gave as cherry pick simplified this part so it was just model_name list.

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
```diff
@@ -0,0 +1,131 @@
...
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    skip_full_eval_test(
+        record_property,
+        cc,
+        model_name,
+        bringup_status="FAILED_RUNTIME",
+        reason="Model compilation exceeds CI timeout limit (240 minutes), compiles e2e locally. Model execution fails with `Not enough space to allocate 47316992 B DRAM buffer across 12 banks.`",
+        model_group=model_group,
+        model_name_filter="stabilityai/stable-diffusion-3.5-large",
```
kmabeeTT, line 117: After the most recent force push which started to use the shorter model names and remove duplicate model names across py tests (thank you), this is incorrect:

```
model_name_filter="stabilityai/stable-diffusion-3.5-large",
```

should be:


```
model_name_filter="SD3.5-large-transformer",
```

tests/models/stable_diffusion/test_stable_diffusion_transformer.py
ddilbazTT: replaced the static dict with `dict(model_info)`

kmabeeTT: Nice! Even better - thanks a lot!

kmabeeTT (approved):
Thank you!

ddilbazTT:
I am rebasing after Aleks' changes, will push this weekend

PR #750, opened 2025-05-09
Jzx/torch2.7 test (Dummy)

jameszianxuTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


jameszianxuTT:
Will revert this change if it breaks nightlies.

tt_torch/dynamo/executor.py
```diff
@@ -305,8 +305,6 @@ def __call__(self, *inputs):
             )
 
         assert self.binary is not None
-        if self.compiler_config.typecast_inputs:
```
ddilbazTT, line 308: nit: did you accidentally changed where typecast_inputs is called or is there a reason?

requirements.txt
```diff
@@ -1,4 +1,5 @@
-torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
+torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.7.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
+torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl
```
ddilbazTT, line 2: Just out of curiosity, did we fix the torchvision version for stability? Or does it break with other versions?

.github/workflows/run-tests.yml
```diff
@@ -92,6 +91,12 @@ jobs:
         source env/activate
         pip install ${{ steps.strings.outputs.dist-dir }}/*.whl
 
+    - name: Show actual installed python dependencies
+      shell: bash
+      run: |
+        source env/activate
+        pip freeze
```
ddilbazTT, line 98: i like the pip freeze :) good thinking!

ddilbazTT (approved):

tt_torch/dynamo/executor.py
ddilbazTT: nvm it's in the description which I didn't read before reviewing

requirements.txt
jameszianxuTT: The torchvision version is in lockstep with torch, so both need to be the same type (CPU cxx11 py310) and uplifted at the same time

ddilbazTT (commented):
Actually now that I played around with this, I think all tests with `model.generate` will fail. We need to update to `transformers>=4.50.0` Even then, I think there is some instability with tracing. [https://github.com/pytorch/pytorch/issues/141196](https://github.com/pytorch/pytorch/issues/141196)

jameszianxuTT:
@ddilbazTT noted. That is probably enough reason to hold off and do more testing. Thanks for investigating this!

jameszianxuTT:
Assessing impact on generate models with unfixed transformers @ wide sweep - https://github.com/tenstorrent/tt-torch/actions/runs/14975490557

ddilbazTT:
So I looked into this a little bit more. I tried removing any fixed versions in the `requirements.txt` and ran beit again, as this was failing with later transformers versions. This is passing for me now. Here is my pip freeze output for reference:

```
accelerate==1.6.0
aiohappyeyeballs==2.6.1
aiohttp==3.11.18
aiosignal==1.3.2
async-timeout==5.0.1
attrs==25.3.0
audioread==3.0.1
bi-lstm-crf==0.2.1
black==25.1.0
certifi==2025.4.26
cffi==1.17.1
cfgv==3.4.0
charset-normalizer==3.4.2
click==8.2.0
coloredlogs==15.0.1
contourpy==1.3.2
coverage==7.8.0
cycler==0.12.1
datasets==3.6.0
decorator==5.2.1
diffusers==0.33.1
dill==0.3.8
distlib==0.3.9
distro==1.9.0
efficientnet_pytorch==0.7.1
einops==0.8.1
et_xmlfile==2.0.0
exceptiongroup==1.3.0
filelock==3.18.0
flatbuffers==25.2.10
fonttools==4.58.0
frozenlist==1.6.0
fsspec==2025.3.0
gliner==0.2.19
hf-xet==1.1.0
huggingface-hub==0.31.1
humanfriendly==10.0
identify==2.6.10
idna==3.10
importlib_metadata==8.7.0
iniconfig==2.1.0
Jinja2==3.1.6
joblib==1.5.0
kiwisolver==1.4.8
kornia==0.8.1
kornia_rs==0.1.9
lazy_loader==0.4
librosa==0.11.0
llvmlite==0.44.0
loguru==0.7.3
MarkupSafe==3.0.2
matplotlib==3.10.3
mdutils==1.6.0
ml_dtypes==0.5.1
mlp-mixer-pytorch==0.2.0
mpmath==1.3.0
msgpack==1.1.0
multidict==6.4.3
multiprocess==0.70.16
mypy_extensions==1.1.0
nanobind==2.7.0
networkx==3.4.2
ninja==1.11.1.4
nodeenv==1.9.1
numba==0.61.2
numpy==2.2.5
onnx==1.17.0
onnxruntime==1.22.0
opencv-contrib-python==4.11.0.86
opencv-python==4.11.0.86
openpyxl==3.1.5
packaging==25.0
pandas==2.2.3
pathspec==0.12.1
pillow==11.2.1
platformdirs==4.3.8
pluggy==1.5.0
pooch==1.8.2
pre_commit==4.2.0
propcache==0.3.1
protobuf==6.30.2
psutil==7.0.0
py-cpuinfo==9.0.0
pyarrow==20.0.0
pybind11==2.13.6
pycparser==2.22
pyparsing==3.2.3
pytest==8.3.5
pytest-cov==6.1.1
python-dateutil==2.9.0.post0
pytorchcv==0.0.73
pytz==2025.2
PyYAML==6.0.2
regex==2024.11.6
requests==2.32.3
safetensors==0.5.3
scikit-build==0.18.1
scikit-learn==1.6.1
scipy==1.15.3
seaborn==0.13.2
sentencepiece==0.2.0
six==1.17.0
soundfile==0.13.1
soxr==0.5.0.post1
stablehlo==1.0.0.1715728102+6051bcdf
sympy==1.14.0
tabulate==0.9.0
threadpoolctl==3.6.0
tiktoken==0.9.0
timm==1.0.15
tokenizers==0.21.1
tomli==2.2.1
torch @ https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.7.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl#sha256=2ae99d57ce02d9ea73683288369b61c77805a93596d5d4a50d2031e114057e17
torchaudio==2.7.0+cpu
torchvision @ https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl#sha256=b1070b55bed045e0f41afe8d6617fc1b6d47088aa3b3430d87f8fcac48f43386
tqdm==4.67.1
transformers==4.51.3
typing_extensions==4.13.2
tzdata==2025.2
ultralytics==8.3.132
ultralytics-thop==2.0.14
urllib3==2.4.0
virtualenv==20.31.2
XlsxWriter==3.2.3
xxhash==3.5.0
yarl==1.20.0
zipp==3.21.0
```
As for the `model.generate` errors, something I found was from [PyTorch - troubleshooting data dependent operations](https://docs.pytorch.org/docs/stable/torch.compiler_troubleshooting.html#data-dependent-operations) I think `model.generate` will almost always have `.item()` access. As following the recommendation, I added the following to the beginning of 'compile_model`:
```
torch._dynamo.config.capture_scalar_outputs = True
```

This was passing `test_gpt_neo.py` for me.

TLDR: upgrading `transformers` and adding `torch._dynamo.config.capture_scalar_outputs = True` to `compile_model` under `tests/utils.py`might be enough to fix model.generate related issues for PyTorch 2.7 upgrade.

jameszianxuTT:
Thanks @ddilbazTT  for looking into this.

I triggered a [small nightly sweep with unpinned transformers](https://github.com/tenstorrent/tt-torch/actions/runs/14975490557/job/42067877543)  which saw a couple of different failures in codegen / llama3 in a limited sweep, though these seem to be different than what you're seeing, though I didn't test gpt_neo. The profile test failures can be ignored.

I think something related to kv cache default behavior has changed with llama as a dynamic cache object is now returned and breaks on asserting PCC (Should be an easy fix). 

The codegen compile issue also seems to relate to some changed default generation behaviour - `AssertionError: Please convert all Tensors to FakeTensors first or instantiate FakeTensorMode with 'allow_non_fake_inputs'. Found in aten.clone.default(tensor(1, size=()))`  - is related to the generation config and us not passing in an eos token and falling back to a new weird behaviour.

Will continue making point fixes tomorrow!

tests/utils.py
```diff
@@ -213,7 +213,26 @@ def flatten(t):
...
+            non_tensor_objects = [
+                t for t in flattened_outputs if not isinstance(t, torch.Tensor)
+            ]
+
+            # Warn about discarded non-tensor objects
+            if non_tensor_objects:
+                discarded_types = {type(obj).__name__ for obj in non_tensor_objects}
+                print(
+                    f"Discarding non-tensor objects of types: {', '.join(discarded_types)}"
+                )
```
jameszianxuTT, line 232: Eg. 
```
Discarding non-tensor objects of types: DynamicCache // (from golden)
Discarding non-tensor objects of types: DynamicCache // (from hardware)
```

To unblock llama, qwen, mistral

tests/utils.py
```diff
@@ -213,7 +215,26 @@ def flatten(t):
                     else [item for sublist in t for item in flatten(sublist)]
                 )
 
-            return flatten(output_object.to_tuple())
+            # The output of some transformers models is CausalLMOutputWithPast,
+            # which when converted to_tuple and flattened which may contain DynamicCache objects
+            # since transformers 4.51
```
ddilbazTT, line 220: Actually the DynamicCache might contain some data we can use. We can call `_flatten_dynamic_cache` and then append that to outputs/ golden. [https://github.com/huggingface/transformers/blob/main/src/transformers/cache_utils.py](https://github.com/huggingface/transformers/blob/main/src/transformers/cache_utils.py)

tests/utils.py
```diff
@@ -364,6 +395,8 @@ def get_framework_model(self):
             if hasattr(self.framework_model, "eval")
             else self.framework_model
         )
+        if hasattr(model, "can_generate") and model.can_generate():
```
AleksKnezevic, line 398: Just because a model is generative, doesn't necessarily mean we want to run generate. It should be an argument to the tester.

tests/utils.py
jameszianxuTT: Yeah, that's a good idea, thanks!

tests/utils.py
jameszianxuTT: Thanks, yeah, cherry-picked in [Replace is_generate with run_generate](https://github.com/tenstorrent/tt-torch/commit/4bf539810e14676e966c4a2728eb5e1395de69f1)

jameszianxuTT:
Merging with #765, which is rebased on this change.

Issue #663, opened 2025-04-26
Enable PCC/ATOL for test_torchvision_image_classification_red[full-eval-swin_b]


kmabeeTT:
This test is being added to `.github/workflows/run-full-model-execution-tests-nightly.yml` today since it can run and pass with workaround to disable PCC and/or ATOL, so using that as a workaround for now to protect the model, and opening this ticket to track removal of workaround.

```
FAILED tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification_red[full-eval-swin_b] -
AssertionError: PCC of output 0: 0.0378, threshold: 0.97 ❌
ATOL of output 0: 8.8750, threshold: 0.01 ❎ (assert_atol == False)
```

kmabeeTT:
This takes 2 minutes on CI and is customer priority model, so let's look at it ~next please:

```
127.38s call     tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification_red[full-eval-swin_b]

```

kmabeeTT:
Actually, we found today this model was mislabeled, it is not a red model. It will be corrected tonight.

PR #756, opened 2025-05-12
Demote int64 constant value to int32

sshonTT:
### Ticket
[Link to Github Issue
](https://github.com/tenstorrent/tt-torch/issues/663)

### Problem description
Our device misinterprets int64 tensors when they are provided as constants, resulting in like [1, 0, 2, 0, 3, 0, ...] instead of [1, 2, 3, ...]. 

### What's changed
To prevent this, this function safely casts int64 tensors to int32 if all values fit within the int32 range, ensuring correct data is passed to the device.

### Checklist
- [v] Run swin_b without PCC drops with https://github.com/tenstorrent/tt-mlir/pull/3320

### DIFF
--- a/tt_torch/dynamo/executor.py
+++ b/tt_torch/dynamo/executor.py
@@ -305,8 +305,6 @@ class Executor:
             )
 
         assert self.binary is not None
-        if self.compiler_config.typecast_inputs:
-            inputs = self.typecast_inputs(inputs)
 
         activations_len = len(inputs)
         if (
@@ -314,6 +312,9 @@ class Executor:
             and self.preprocessed_graph_constants is None
         ):
             inputs = self.graph_constants + inputs
+
+        if self.compiler_config.typecast_inputs:
+            inputs = self.typecast_inputs(inputs)
 
         inputs = list(inputs)
         device = self._get_device()


tt_torch/dynamo/passes.py
```diff
@@ -110,6 +110,27 @@ def constant_fold(gm):
...
+"""
+    Workaround for int64 constant feeding issue on our device.
+    Our device misinterprets int64 tensors when they are provided as constants,
+    resulting in corrupted data like [1, 0, 2, 0, 3, 0, ...] instead of [1, 2, 3, ...].
+    To prevent this, this function safely casts int64 tensors to int32 if all values
+    fit within the int32 range, ensuring correct data is passed to the device.
+"""
+
+
+def sanitize_int64_constants_for_device(constant_inputs):
```
AleksKnezevic, line 122: We already have infra for this, see `typecast_inputs` in executor.py. We should add this to that.

sshonTT: Right, I had a discussion with Defne about this. Thanks, I will check.

sshonTT: I was able to fix it by using `typecast_inputs`.

AleksKnezevic: How about you move the call to typecast just after line 316 in [executor.py](https://github.com/tenstorrent/tt-torch/blob/5a3289e89febc726b2da28016362da13f4048621/tt_torch/dynamo/executor.py#L316). Then it's all in one place.

AleksKnezevic (changes requested):

tt_torch/dynamo/passes.py
sshonTT: That makes sense. I was also thinking about how to handle `self.compiler_config.typecast_inputs`. Since it appears to be enabled by default, I think I can go ahead and move it.

sshonTT: Fixed it.

ddilbazTT (approved):

AleksKnezevic (approved):

Issue #763, opened 2025-05-13
Extend model tests to support multiple batch


AleksKnezevic:

tapspatel:
Priority red models that have good PCC
- BERT_Turkish
- MobileNetV2
- SegFormer
- ViT
- YOLOv4
- efficientnet-b0

@gengelageTT These models are dependent on this issue: https://github.com/tenstorrent/tt-torch/issues/758. For any models that fail compilation, please let me know and I can triage on backend compiler side. Thanks!

PR #764, opened 2025-05-13
Add support for automatic data parallelization guarded by compiler config option

tapspatel:
Issue: https://github.com/tenstorrent/tt-torch/issues/763

Usage
```
cc.automatic_parallelization = True
cc.mesh_shape = [1, 8]
```

Example
```
pytest -ssv tt-torch/tests/models/resnet50/test_resnet50_llmbox.py::test_resnet[full-eval]
```

### DIFF
--- a/tt_torch/csrc/tt-mlir-interface.cpp
+++ b/tt_torch/csrc/tt-mlir-interface.cpp
@@ -38,6 +38,7 @@
 #include "ttmlir/Dialect/TT/IR/TTOps.h"
 #include "ttmlir/Dialect/TT/IR/TTTraits.h"
 
+#include "ttmlir/Dialect/StableHLO/Pipelines/StableHLOPipelines.h"
 #include "ttmlir/Dialect/TTIR/Pipelines/TTIRPipelines.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h"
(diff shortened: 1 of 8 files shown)


tests/models/resnet50/test_resnet50.py
```diff
@@ -82,4 +83,28 @@ def empty_record_property(a, b):
 
 # Run pytorch implementation
 if __name__ == "__main__":
-    test_resnet(empty_record_property)
+    parser = argparse.ArgumentParser(description="Run resnet50")
```
AleksKnezevic, line 86: what's this for?

AleksKnezevic: These tests are usually just called via pytest.

tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -53,6 +54,77 @@
...
+    automatic_sharding_pipeline_pm.getContext()->disableMultithreading();
+    automatic_sharding_pipeline_pm.enableIRPrinting();
+  }
+
+  mlir::tt::stablehlo::AutomaticShardingPipelineOptions
+      automatic_sharding_pipeline_options;
+  automatic_sharding_pipeline_options.meshShape = mesh_shape;
+
+  // Set argument types.
+  if (len_activations > 0 || len_graph_constants > 0) {
```
AleksKnezevic, line 96: Can you please pull this out into a helper and call from both functions?

tests/models/resnet50/test_resnet50.py
AleksKnezevic: @brataTT, how complicated is it to create a system descriptor on compile, not on pytest session start? In general, I'm not a huge fan of using pytest infra for this since user scripts won't work as expected.

AleksKnezevic: @tapspatel so this PR won't have any test coverage until we update the infra, correct?

tapspatel: @AleksKnezevic that's right. 

We can update this function
```
void create_system_desc(std::optional<tt::runtime::Device> device)
```

which pytest calls to pass in the mesh shape but it would still require some changes to pytest sessions (maybe add a parameterized variable to tests).

The other option is to do it on compile as you suggested so there's a single location which generates the system desc, we would have to pass down the mesh in this function.

```
compileTTIRToTTNN
```

Both work but I wasn't sure which one was preferred so I left it out of this PR.

tt_torch/csrc/tt-mlir-interface.cpp
tapspatel: fixed.

tests/models/resnet50/test_resnet50.py
brataTT: @AleksKnezevic not complicated.
We can do it the same way you implemented executing using the same device
Does that blocks this PR? I can prioritize it if it does

AleksKnezevic: @tapspatel, @brataTT will add the required infra change to your PR so can merge it in with some test coverage.

tapspatel: thanks @brataTT your changes worked.

brataTT: So for tests, all that's left to do is add the new pytest argument and run it in `run-full-model-execution-tests.yml`?

And for future changes to system_desc, I'm guessing `mlir::tt::stablehlo::createAutomaticShardingPipeline` has nothing to do with system_desc, it's just about the mesh_shape we were creating by default that was wrong?

tapspatel: 1. We need some infrastructure to run these types of tests on multi chip. For that, we need to know the type of device we are running on so we can set the mesh shape appropriately and a way to increase the batch size depending on num of chips
2. createAutomaticShardingPipeline doesn't care about anything but mesh shape you give to it. The issue was opening a device in multi-chip mesh not matching the system_desc, which was exposed by this PR but not related.

AleksKnezevic: If the parsing is no longer needed, please remove.

tt_torch/dynamo/executor.py
```diff
@@ -160,6 +160,20 @@ def __init__(
...
+
+    def _remove_system_desc(self):
+        try:
+            os.remove(os.getenv("SYSTEM_DESC_PATH"))
+        except OSError:
+            pass
+
+    def _create_system_descriptor(self):
+        self._remove_system_desc()
+        device = self.device if self.device is not None else self._get_device()
```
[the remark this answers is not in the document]
brataTT: Thanks for catching this @jameszianxuTT
Talked to @tapspatel and we decided I'll create a separate change for rewriting system_desc creation for this PR to rebase on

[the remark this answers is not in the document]
tapspatel: thanks @brataTT with the fix. updated all changes.

vmilosevic:

Issue #714, opened 2025-05-05
torch audio


AleksKnezevic:
Let's bringup a model that needs torch_audio (maybe [this one](https://huggingface.co/nari-labs/Dia-1.6B), but please double check that it actually does use torch_audio). The main complexity is that torch audio (like torchvision) needs a cxx11ABI build, but is delivered by default in with pre cxx11ABI. 

Do we need to build our own torch audio wheel that uses cxx11 ABI like we do for torchvision? Alternatively, and preferably, AFAIK, in pytorch >= 2.7, pre cxx11ABI has been disabled. Can we just bump to 2.7.0 and everything works?

ddilbazTT:
This project is using [torch audio 2.6.0](https://github.com/nari-labs/dia/blob/main/pyproject.toml)

AleksKnezevic:
@ddilbazTT can you try it out w/ our stack?

ddilbazTT:
@AleksKnezevic 
I found an easier model to try out with since this one needed me to clone the repo. 

```
from transformers import AutoProcessor, SeamlessM4Tv2Model
import torchaudio

processor = AutoProcessor.from_pretrained("facebook/seamless-m4t-v2-large")
model = SeamlessM4Tv2Model.from_pretrained("facebook/seamless-m4t-v2-large")

# from text
text_inputs = processor(text = "Hello, my dog is cute", src_lang="eng", return_tensors="pt")
audio_array_from_text = model.generate(**text_inputs, tgt_lang="rus")[0].cpu().numpy().squeeze()

# from audio
audio, orig_freq =  torchaudio.load("tests/models/Dia/preamble.wav")
audio =  torchaudio.functional.resample(audio, orig_freq=orig_freq, new_freq=16_000) # must be a 16 kHz waveform array
audio_inputs = processor(audios=audio, return_tensors="pt")
audio_array_from_audio = model.generate(**audio_inputs, tgt_lang="tur")[0].cpu().numpy().squeeze()

import scipy

sample_rate = model.config.sampling_rate
scipy.io.wavfile.write("out_from_text.wav", rate=sample_rate, data=audio_array_from_audio)
```
It's coming from [facebook/seamless-m4t-v2-large](https://huggingface.co/facebook/seamless-m4t-v2-large)

I needed to `pip install torchaudio` I didn't try running this in our ModelTester framework since just trying out cpu execution in our virtual environment would be enough for the scope of this issue for now. 

[...]

PR #765, opened 2025-05-13
torchaudio #714 and generative models in torch 2.7

ddilbazTT:
This is a draft PR to discuss a way to handle generative models in torch 2.7 uplift and tt-torch backend. 

I am providing an example in this PR, which is a text-to-speech model by facebook. This is a generative model, which also can be determined by running `model.can_generate()` 

In this branch, what I did was I compiled `model.forward` with `torch.compile`, instead of the entire model. Then invoked `generate` on it. This might be a better approach for generative models. We can generalize this by using `can_generate` in `compile_model` and `run_model` functions, or somewhere else. Depending on generative model or not, we could follow two separate paths. 

When I tried compiling the model instead of the forward pass, then calling generate on it; it didn't run on our backend. Thanks to @LPanosTT for suggesting to use the forward pass instead.


ddilbazTT:
@AleksKnezevic @jameszianxuTT I would appreciate if you could take a look at this PR. My steps to follow are:
- [ ] debug `tests/models/speecht5_tts/test_speecht5_tts.py` - it's failing during verification
- [ ] run a nightly test sweep with changes to see errors (do this with/ without torch upgrade)
- [ ] undo torch upgrade and put up PR so James can put a torch upgrade PR separately. This is contingent on things working when reverting to old torch version.

tests/utils.py
```diff
@@ -373,6 +385,8 @@ def get_framework_model(self):
             if hasattr(self.framework_model, "eval")
             else self.framework_model
         )
+        if hasattr(model, "can_generate") and model.can_generate():
+            self.is_generative = True
```
jameszianxuTT, line 389: Will this automatically force models that can be run in generate, to do so? 

If so, it will change the tested behaviour of some other models that we don't currently test generation on, to start becoming generative. test_llama_3b I believe can run in generate, but we only run a single forward pass currently.

ddilbazTT: Yes, it will force them to generate. Something that would overcome this is setting `self.is_generative` when creating the ModelTester. Let me know if that sounds better

.github/Dockerfile.base
```diff
@@ -36,11 +36,11 @@ RUN apt-get update && apt-get install -y \
     libgl1 \
     libglx-mesa0
 
-# Install python3.11 packages
+# Install python3.10 packages
 RUN python3.10 -m pip install \
     setuptools==59.6.0 \
     wheel \
-    torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
+    torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.7.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
```
AleksKnezevic, line 43: We shouldn't need this anymore. it should just be torch==2.7.0

ddilbazTT: I just used James' branch in the beginning, I can change this here but if we're doing two separate PRs @jameszianxuTT you should keep this in mind too.

requirements.txt
```diff
@@ -1,4 +1,6 @@
-torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
+torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.7.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl
```
AleksKnezevic, line 1: Ditto for here, should just be 
```
torch==2.7.0
torchvision==0.22.0
```

setup.py
```diff
@@ -100,7 +100,7 @@ def run(self):
     },
     zip_safe=False,
     install_requires=[
-        "torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl",
+        "torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.7.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl",
```
AleksKnezevic, line 103: torch==2.7.0

requirements.txt
jameszianxuTT: This is my change. In my local experiments, I saw that the automatically installed torch version would pull in some Nvidia dependencies if I didn't specify the CPU wheel. Will double check, but I think this is likely the default behavior.

tests/utils.py
AleksKnezevic: I agree with this. It's awesome that this works, but we need a way to specify what we want to test. Not all models that support generation will be called from the generative loop. One option would be to provide a `run_generate` argument to the model_tester that would force this behaviour.

tests/utils.py
```diff
@@ -135,6 +135,8 @@ def __init__(
         self.compiler_config.model_name = model_name
         self.compiler_config.model_group = model_group
 
+        self.is_generative = False  # a model can be generative or discriminative
```
AleksKnezevic, line 138: TIL the term discriminative in this context 🙂

tests/utils.py
ddilbazTT: I can remove/ leave`self.is_generative` and create something like `self.run_generate` to determine what to do in compile/ run model. 

If we keep both:
- if `is_generative`==True but `run_generate`==False --> print out warning message
- if `is_generative`==False but `run_generate`==True --> error out

requirements.txt
ddilbazTT: @jameszianxuTT we need to do some cooperation for the uplift/ generative model changes. I will address the comments here but as per my offline chat with Aleks, the changes here will be merged after torch uplift. Maybe we can merge PRs then but let's communicate. My PR addressed two things on top of yours:
- removed transformers library version so we use the latest 
- fixed torch compile issues with generative models

tests/utils.py
ddilbazTT: I referenced [this article](https://dataforest.ai/blog/generative-and-non-generative-ai-the-strengths-of-both-approaches) for the terminology - not that I knew it before :D

AleksKnezevic (approved):

jameszianxuTT:
I think the transformers upgrade broke all our attempts to compile `model.generate`. So the change to only compile the generative models' forward passes looks like one path to fixing a large swath of failing models with the uplift.

ddilbazTT:
> I think the transformers upgrade broke all our attempts to compile `model.generate`. So the change to only compile the generative models' forward passes looks like one path to fixing a large swath of failing models with the uplift.


@jameszianxuTT Yeah compiling model.generate on its own wouldn't work, that's why I switched to compiling the forward pass and calling generate on it. Did you mean that the new approach is causing failures?

jameszianxuTT:
> Did you mean that the new approach is causing failures?

No, I mean it's fixing the failures, so I've cherry picked the generate changes into my uplift and am resweeping

ddilbazTT:
Also something we could do is to comment out designs failing after uplift from our workflows and then fix them in a separate PR. @jameszianxuTT So that the uplift could go in quicker. I assume most would be the generative models.

ddilbazTT:
[James' run](https://github.com/tenstorrent/tt-torch/actions/runs/15069685997/job/42363544405) 

Errors seen:
- tests/models/Qwen/test_qwen2_casual_lm.py::test_qwen2_casual_lm[full-Qwen/Qwen2.5-1.5B-eval] - was due to misplaced `run_generate`, fix pushed
- tests/models/vilt/test_vilt.py::test_vilt[full-eval] - James' branch is behind the commit where Kyle changed verify_outputs logic, it should pass now since verify_outputs shouldn't run with compile_depth ttnn_ir. We don't need to disable assert_atol/ assert_pcc.
- tests/models/mistral/test_pixtral.py::test_pixtral[full-eval] - RuntimeError: Input type (torch.FloatTensor) and weight type (CPUBFloat16Type) should be the same or input should be a MKLDNN tensor and weight is a dense tensor - I'm stuck with this one. Something I investigated was also moving weights and inputs to the same device (by default it should be but I still forced it) I did a few things and am still stuck, I'll note them down later but for now, I'll skip this test.
- tests/models/vit/test_vit_onnx.py::test_vit_onnx[full-eval] - TypeError: 'str' object is not callable - haven't looked into it yet but looks simple to fix

ddilbazTT:
I provided 3 different ways to handle DynamicCache output in my last 3 commits.
[C-2:](https://github.com/tenstorrent/tt-torch/pull/765/commits/91f951ac10cfd9057ede96a200151cd02cc2dd23) 
- In this case, we set `use_cache=False` when the model is generative, regardless of whether the model is running generate or not. This returns a `None` item for `output_object['past_key_values']`
[C-1:](https://github.com/tenstorrent/tt-torch/pull/765/commits/d13cd1b67bf48ff56b85da6ba2eb9d838d3bdc92)
- In this case, `use_cache=True` by default since we do not set it to `False`. This means that the output object will have a `past_key_values` attribute that is a `DynamicCache` object. I am flattening the `DynamicCache` object using built in function (returns a tuple), and only using the first element since the second is a list of string (`['key_cache', 'value_cache']`) Since the first element is a list, we need to recursively call `flatten` again. 
[C-0:](https://github.com/tenstorrent/tt-torch/pull/765/commits/d13cd1b67bf48ff56b85da6ba2eb9d838d3bdc92) 
- In this case, `use_cache=True` by default since we do not set it to `False`.  However, `DynamicCache` object is ignored since it might not be relevant to verify cached values for atol/ pcc. 

@AleksKnezevic @jameszianxuTT could you let me know if you have a preference between any of these methods?

jameszianxuTT:
@ddilbazTT Thanks for laying out these options. 

I think C-0 is the best option:

C1: I'm not sure if we want to fail PCC because contents of the caches differ, as that will obscure the more important measure of output PCC comparison. I think it's possible that the caches (which reflect internal K/V weight matrices in each attention head) could drift significantly, and I agree it's not the most relevant measure for "correctness". 

C2: I think we'll eventually need to run generative models with cache enabled and disabled in the future based on my task. 
I think this option effectively overrides the default `use-cache` to false, but if this is desirable it might be better to explicitly mark this in `_load_model` itself rather than within the ModelTester.

ddilbazTT:
@jameszianxuTT Yeah I don't know if we want to fail pcc due to dynamic cache either. However, it seems like pcc assertion works fine when we verify results with cache results included. So there might be minimal risk, or if things fail it might be indicative of good debugging information?

jameszianxuTT:
@ddilbazTT I wonder what the "output" dynamic cache is, since we definitely cannot compile it. I wonder if it's just calculated on CPU anyways through a graph break, which might be confirmed if the output and golden dynamic caches are suspiciously similar.

ddilbazTT:
@jameszianxuTT I think kv_cache might be stored in CPU, but the calculation comes from what we run on our backend (?) I based it on [this article](https://huggingface.co/blog/not-lain/kv-caching)

ddilbazTT:
Talked with Aleks off-line and we decided it's best to use the `DynamicCache` values in the output since they are an output of the model.

ddilbazTT:
Is docker build-image supposed to take so long?

[link](https://github.com/tenstorrent/tt-torch/actions/runs/15144849732/job/42577942428)

jameszianxuTT:
>Is docker build-image supposed to take so long?

Sorry, meant to respond to this - the logic here is that the docker-build step will:
1. hash the dockerfile.base and dockerfile.ci
2. go out to ghcr cloud storage and see if files with those hashes exist
3. if those files exist, pull the image keyed with that hash 
4. if those files don't exist, rebuild the ones that don't, upload them to ghcr 

So you probably went into branch (4) if you modified the dockerfiles

ddilbazTT:
mistral models are failing due to uplifting transformers and pytorch together. When each individually uplifted, it works fine. 

Spoke with James off-line and something I will try is to use a downgraded transformers for mistral specifically, like we do in timm. 

torch uplift requires transformers>=4.50 for generative models. The latest one, and the one grabbed by dependency resolver, is 4.52. Even 4.50 errors out mistral, so I think the test specific transformers version idea sounds better.

Issue #770, opened 2025-05-14
Pipeline parallel support


AleksKnezevic:

ddilbazTT:
I assume this is the reference for this issue: [PyTorch - big_model_inference](https://huggingface.co/docs/accelerate/v0.25.0/en/concept_guides/big_model_inference)

AleksKnezevic:
yes

PR #771, opened 2025-05-14
Automatic pipeline parallel 

AleksKnezevic:
Added support for automatic pipeline parallel based on transformers device_map infrastructure. Transformers library allows us to shard a model on multiple devices if it does not fit into the DRAM of a single device. This appears to be a pretty rudimentary pass which just takes model weights (in the order that they're allocated in the model) and sums them up, and once they no longer fit on one device, throws them onto the next device. This is enough for us to start a high level pipeline parallel pass. 

A follow up will be to simply create the separate shards, but decide which device they go on based on a topological sort we perform ourselves. This will remove the manual override in test_llama_7b_pipeline_parallel. See #779.

With this PR, I added the functionality to read the device map and create multiple graphs (from the original FX graph which contains the entire model), one for each device. These graphs are then compiled separately (into individual flatbuffers) and executed consecutively. 

### Ticket
#770 

### What's changed
Models can be pipelined across devices by supplying a device map to compiler_config
Refactored backend and executor to allow for multiple device graphs.

### DIFF
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -77,6 +77,7 @@ jobs:
                   tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-regnet_x_8gf]
                   tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-regnet_x_16gf]
                   tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-regnet_x_32gf]
+                  tests/models/mistral/test_mistral.py::test_mistral[full-ministral3b-eval]
             "
           },
           {
@@ -97,10 +98,11 @@ jobs:
                 tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-wide_resnet101_2]
             "
           },
+          # This test group needs to be moved. This will be done once multichip tests are refactored: #780
           {
             runs-on: n300, name: "eval_6", tests: "
                 tests/models/llama/test_llama_7b_pipeline_parallel.py::test_llama_7b_pipeline_parallel[huggyllama/llama-7b-eval]
-                tests/models/mistral/test_mistral.py::test_mistral[full-ministral3b-eval]
+                tests/torch/test_basic_async.py
                 tests/torch/test_basic_multichip.py
             "
           },
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -101,6 +101,7 @@ jobs:
             runs-on: n300, name: "eval_6", tests: "
                 tests/models/llama/test_llama_7b_pipeline_parallel.py::test_llama_7b_pipeline_parallel[huggyllama/llama-7b-eval]
                 tests/models/mistral/test_mistral.py::test_mistral[full-ministral3b-eval]
+                tests/torch/test_basic_multichip.py
             "
           },
         ]
--- a/tests/models/llama/test_llama_7b_pipeline_parallel.py
+++ b/tests/models/llama/test_llama_7b_pipeline_parallel.py
@@ -16,6 +16,8 @@ from transformers.modeling_outputs import (
 from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
 from transformers.cache_utils import Cache, DynamicCache
 from transformers.processing_utils import Unpack
+from accelerate import infer_auto_device_map
+from tt_torch.tools.verify import verify_against_golden
 
 
 def get_model_and_tokenizer(model_name):
@@ -27,7 +29,6 @@ def get_model_and_tokenizer(model_name):
     m = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.bfloat16)
     for param in m.parameters():
         param.requires_grad = False
-    m.eval()
     return m, tokenizer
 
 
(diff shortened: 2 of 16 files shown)


.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -96,6 +96,7 @@ jobs:
             runs-on: n300, name: "eval_6", tests: "
                 tests/models/llama/test_llama_7b_pipeline_parallel.py::test_llama_7b_pipeline_parallel[huggyllama/llama-7b-eval]
                 tests/models/mistral/test_mistral.py::test_mistral[full-ministral3b-eval]
+                tests/torch/test_basic_multichip.py
```
jameszianxuTT, line 99: Did this file not get git added?

Also, `run-tests.yml` picks up every test under `tests/torch` but doesn't guarantee run on n300, so you may want to pass `--ignore=tests/torch/test_basic_multichip.py` to the invocation of pytest [here](https://github.com/tenstorrent/tt-torch/blob/82aafff4e072d761d1f9f1a17dd17ccb6048a40f/.github/workflows/run-tests.yml#L105), or relocate the test file to another directory.

tt_torch/dynamo/executor.py
```diff
@@ -304,51 +350,75 @@ def __call__(self, *inputs):
...
+            preprocessed_weights, preprocessed_activations = self.get_inputs(
+                *device_inputs,
+                binary=binary,
+                device_idx=device_idx,
+                program_idx=program_idx,
             )
 
-        if self.compiler_config._enable_intermediate_verification:
-            # put this as close to the binding as possible to ensure the GM is not mutated past this point
-            self._generate_golden_intermediate_cache(self.program, inputs)
```
jameszianxuTT, line 339: Removing the call to `_generate_golden_intermediate_cache` will break runtime intermediate checking, but I think supporting multidevice runtime intermediates will be a nontrivial change. 

One quick change would be to only allow this in the single-device case, and I will need to run some experiments to determine how to uniquely tag intermediates from nodes in different programs across devices. 

Or perhaps, it is unnecessary to validate runtime intermediates for a pipeline-parallel run, and we'd be better off validating intermediates on a single device only.

ddilbazTT (commented):
I hate to be that person but either here or in the issue itself, could you give more details about the changes?

tests/torch/test_basic_multichip.py
```diff
@@ -0,0 +1,63 @@
...
+        end = time.time()
+        times.append(end - start)
+    min_time = min(times)
+    max_time = max(times)
+    mean_time = sum(times) / len(times)
+    std_time = np.std(times)
+    print(f"Min time taken: {min_time}")
+    print(f"Max time taken: {max_time}")
+    print(f"Mean time taken: {mean_time}")
+    print(f"Standard deviation time taken: {std_time}")
```
jnie-TT, line 61: Would it make sense to perhaps create a helper API/python context that collects/displays these measurements? Might be good for consistency and we could possibly hook it up to our CI data collection infrastructure to postprocess/pipeline into our dashboard.

tt_torch/dynamo/executor.py
```diff
@@ -207,30 +198,34 @@ def typecast_inputs(self, inputs):
...
 
     def _cleanup_resources(self, preprocessed_activations, device):
         for t in preprocessed_activations:
             tt_mlir.deallocate_tensor(t, force=True)
 
-        if self.device is None:
+        # if the user didn't provide a device, we opened one, so we need to close it
+        if device in self.owned_devices:
             tt_mlir.close_mesh_device(device)
+            self.owned_devices.remove(device)
```
jnie-TT, line 228: Seems like we're adding the owned device to self.devices, should we remove it from `self.devices` as well?

tt_torch/dynamo/executor.py
```diff
@@ -133,24 +133,14 @@ def execute_process(receiver, sender, exec_event):
...
     def __init__(
         self,
-        program: Union[torch.export.ExportedProgram, None] = None,
-        graph_constants=None,
+        mcg: MultiChipGraph,
         compiler_config=None,
         required_pcc=0.99,
         required_atol=1e-2,
-        device=None,
+        devices=[None],
```
jnie-TT, line 140: I think this is dangerous in python - all class instances will share the same list. If I do something like
```
e1 = Executor()
e2 = Executor()

e1.devices.append("device1")
print(e2.devices)
```
I think obj2.devices will also have the appended device1. Could we maybe default to None and set it to [None] within __init__ separately?

tests/models/llama/test_llama_7b_pipeline_parallel.py
```diff
@@ -49,308 +50,41 @@ def decode_output(outputs, tokenizer):
...
-
-    # Compile the second half, targeting device2
-    options2 = BackendOptions()
-    cc2 = CompilerConfig()
-    options2.compiler_config = cc2
-    options2.device = device2
-    second_half = LlamaSecondHalf(original_model)
-    tt_second_half = torch.compile(
-        second_half, backend=backend, dynamic=False, options=options2
+    device_map = infer_auto_device_map(
```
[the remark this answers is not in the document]
AleksKnezevic: Yeah, this is a great idea. Also, as it stands we just load the model in whatever data format we have onto the device, but there will eventually be some AMP passes which can reduce data formats further so more layers can fit. We'll need a way to account for that.

.github/workflows/run-full-model-execution-tests.yml
AleksKnezevic: Interesting. How did it skip the pipeline parallel llama test before?

tt_torch/dynamo/executor.py
AleksKnezevic: Great catch! I moved this to be the last part in passes.py. I recall you saying that you put it just before execute to avoid any graph transformations. AFAIK, no further graph transformations happen after the program is generated. Is there a test I can run to confirm that I'm generating intermediates correctly.

tests/torch/test_basic_multichip.py
AleksKnezevic: It definitely would! This PR is large enough as is, so I created #777 to track it.

tt_torch/dynamo/executor.py
AleksKnezevic: Great catch.

AleksKnezevic:
@ddilbazTT, I expanded the description and added some followup todos. I appreciate you calling me out for this! This is how we get better.

tt_torch/dynamo/passes.py
```diff
@@ -110,48 +108,397 @@ def constant_fold(gm):
...
+            devices = list(set(user_devices))
+            if len(devices) == 0:
+                devices = [prev_device_idx]
+            for device_idx in devices:
+                prev_device_idx = device_idx
+                inp_node = (
+                    mcg.device_graphs[device_idx].get_attr(node.target)
+                    if node.op == "get_attr"
+                    else mcg.device_graphs[device_idx].placeholder(node.target)
+                )
```
hshahTT, line 273: Is there a case where the `device_map` has multiple devices, but `node_to_device()` returns None for all nodes in the graph (maybe due to some parsing error)? In that case L270 will fail due to trying to access an invalid index.

Maybe we can throw an assert to make sure we have at least 1 valid device in `user_devices`?

hshahTT (changes requested):

tests/models/llama/test_llama_7b_pipeline_parallel.py
```diff
@@ -49,308 +50,47 @@ def decode_output(outputs, tokenizer):
...
-    print(
-        f"""
-      model_name: {model_name}
-      input: {prompt}
-      output: {decoded_output}
-      """
+    # The device map shards model chunks based on the order they're defined in the model
+    # not based on the topological order of the graph. We need to move rotary embeddings
+    # to the first device, otherwise we wouldn't be able to execute each device in full.
+    # That is, device 1 would need to run rotart_emb, device 0 the full graph, device 1,
```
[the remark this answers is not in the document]
AleksKnezevic: I was just saying we need to override it otherwise this is what would happen.

tt_torch/dynamo/passes.py
```diff
@@ -110,48 +108,397 @@ def constant_fold(gm):
...
+            )
+            .run_decompositions(decompositions)
+            .module()
         )
-        for i in range(len(program._graph_signature.input_specs)):
-            if program._graph_signature.input_specs[i].kind != InputKind.USER_INPUT:
-                program._graph_signature.input_specs[i].kind = InputKind.USER_INPUT
-    else:
-        constant_inputs = []
+        run_shape_prop(gm_device, sub_example_inputs)
```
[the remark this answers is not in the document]
AleksKnezevic: good catch, it is not.

tt_torch/dynamo/passes.py
AleksKnezevic: If we can't find the device index of any of the users, we revert to the last device index found. User devices are first filtered with
```python
user_devices = [d for d in user_devices if d is not None]
```
then if none are there, use prev_device_idx.

tt_torch/dynamo/executor.py
```diff
@@ -304,51 +296,75 @@ def __call__(self, *inputs):
...
+                if graph_output.io_type == IOType.INTER_DEVICE:
+                    mci = graph_output.linked_input
+                    graph_inputs[mci.originating_device][mci.consumer_index] = output
+                    intermediate_results.append(output)
+                else:
+                    final_outputs[graph_output.index] = output
+
+        self._cleanup_resources(preprocessed_activations, device_idx)
+        assert all([o is not None for o in final_outputs])
+        return final_outputs
```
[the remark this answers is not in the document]
AleksKnezevic: I don't think they are.

tt_torch/tools/utils.py
```diff
@@ -77,6 +77,45 @@ class OpByOpBackend(Enum):
...
+        self.device_graphs = {device: torch.fx.Graph() for device in devices}
+        self.graph_outputs = {device: [] for device in devices}
+        self.graph_inputs = {device: [] for device in devices}
+        self.programs = {}
+        self.binaries = {}
+        self.constant_inputs = {}
+        self.example_inputs = {}
+        self.shlo_modules = {}
+
+        self.gm = None
```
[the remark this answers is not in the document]
AleksKnezevic: Not only is there a case where we don't set gm, there isn't a case that we do.  🤣 Removed the property.

tt_torch/dynamo/backend.py
```diff
@@ -127,13 +127,13 @@ def _shlo_backend(
...
     program=None,
     graph_constants=None,
-    device=None,
+    devices=None,
     async_mode=False,
 ):
     executor = StablehloExecutor(
         module=shlo,
         compiler_config=compiler_config,
-        device=device,
```
brataTT, line 136: After taps' [data parallel change](https://github.com/tenstorrent/tt-torch/pull/764/files) is in, this `device` -> `devices` update will have conflicts.
I can help rebase + update system_desc creation after it goes in.

We can create a desc for each device and pass in the path during `ttir`->`ttnn` conversion

hshahTT (approved):

jnie-TT (approved):
🔥

jameszianxuTT (approved):

PR #802, opened 2025-05-20
Add shared runners to nightly test gpt and flux

jmcgrathTT:
Shift runners to shared infra pool

Tests:

GPT
https://github.com/tenstorrent/tt-torch/actions/runs/15117785612/job/42502648841#step:1:2

Flux
https://github.com/tenstorrent/tt-torch/actions/runs/15142557785/job/42570965726

### DIFF
--- a/.github/workflows/run-op-by-op-model-tests-nightly.yml
+++ b/.github/workflows/run-op-by-op-model-tests-nightly.yml
@@ -8,6 +8,10 @@ on:
         description: 'Docker image to use for build'
         required: true
         type: string
+      docker-image-harbor:
+        description: 'Shared infra Docker image to use for build'
+        required: true
+        type: string
   workflow_run:
     workflows: [Build]
     types: [completed]
@@ -39,9 +43,9 @@ jobs:
               "
           },
           {
-            runs-on: wormhole_b0, name: "gpt", tests: "
+            runs-on: n300, name: "gpt", tests: "
               tests/models/gpt_neo/test_gpt_neo.py::test_gpt_neo[op_by_op_torch-eval]
-              "
+              ", sh-run: true
           },
           {
             runs-on: wormhole_b0, name: "llama", tests: "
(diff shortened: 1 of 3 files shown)


.github/workflows/run-op-by-op-model-tests-nightly.yml
```diff
@@ -170,13 +174,13 @@ jobs:
           }
         ]
     runs-on:
-      - ${{ matrix.build.runs-on }}
+      - ${{ matrix.build.sh-run && format('tt-beta-ubuntu-2204-{0}-large-stable', matrix.build.runs-on) ||  matrix.build.runs-on }}
```
jameszianxuTT, line 177: Is the reason for introducing harbor instead of ghcr for image storage because these shared runners cannot access ghcr?

jmcgrathTT: Ah, so Harbor is an image cache pull-through proxy so we can get quicker pulls in our local shared GHA infra.

This will work for any upstream repo, but image pushes still need to go directly to the source repo e.g. ghcr

jameszianxuTT: I see, so we want to add another cache level to image retrieval between ghcr and container.

In a follow up PR, could we then make harbor our only source for the image in all files that only download it? Or are we forced to have this branch to pull from ghcr or harbor, based on which cache is quicker, based on the "type" of runner?

jmcgrathTT: That is my hope in the long term, but it might be a month or two out.

There are a couple of reasons: we need to slowly load the shed because of the capacity in the shared infra pool, and we also might run into some tests that have issues. This also allows us to easily fail back over to our old runners by changing `sh-run` to `false` in a DR scenario while still gaining some benefit versus an all-or-nothing approach. 

My goal is to shift a few jobs to the shared pool each day if I see low back pressure in the job queue in a 24-hour period. Unfortunately, access to this Grafana dashboard is limited atm, but once there is SSO access, I will let you know

image

jameszianxuTT (approved):

PR #803, opened 2025-05-20
Fix shlo op by op errors after introduction of MultiChipGraph

ddilbazTT:
### Ticket
None

### Problem description
We switched to `MultiChipGraph`. https://github.com/tenstorrent/tt-torch/pull/771 However, this PR didn't include changes for stablehlo executor concerning torch graphs. 
This was not caught because we do not have any torch models running with stablehlo op-by-op. 
Also, we currently install stablehlo in `env/activate` which is an exception compared to how everything else is installed through `requirements.txt`

### What's changed
The `StablehloExecutor` now takes in mcg object instead of program.
MNIST is ran with `--op_by_op_stablehlo` during on pr tests
Stablehlo now installed through `requirements.txt`

### Checklist
- [X] Edit `StablehloExecutor` and `backend.py` to pipe mcg etc properly
- [X] Add MNIST op by op stablehlo to on pr tests.
- [X] Moved stablehlo from `env/activate` to `requirements.txt`

### DIFF
--- a/tt_torch/dynamo/shlo_backend.py
+++ b/tt_torch/dynamo/shlo_backend.py
@@ -479,16 +479,15 @@ class StablehloExecutor(OpByOpExecutor):
 
         if self.binary is not None:
             try:
-                print("Executing binary")
                 output = tt_mlir.run_end_to_end(
                     self.typecast_inputs(inputs), self.binary
                 )
+                return output
             except Exception as e:
                 print(f"Failed to execute binary: {e}")
 
         if self.program is not None:
             try:
-                print("Executing program")
                 return self.program.graph_module(
                     *(self.graph_constants + tuple(self.program.buffers()) + inputs)
                 )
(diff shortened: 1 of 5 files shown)


tt_torch/dynamo/shlo_backend.py
```diff
@@ -479,6 +479,7 @@ def __call__(self, *inputs):
 
         if self.binary is not None:
             try:
+                print("Executing binary")
```
AleksKnezevic, line 482: Are these intentional?

AleksKnezevic (approved):

tt_torch/dynamo/shlo_backend.py
mmanzoorTT: +1 there are other print statements as well.

mmanzoorTT (approved):

tt_torch/dynamo/shlo_backend.py
ddilbazTT: I thought they might be helpful but now I see they might be confusing. I removed them

Issue #807, opened 2025-05-21
Resnet demo failing with default image on BH


staylorTT:
### Describe the bug

/demos/resnet/resnet50_demo.py fails to run on BH

```
[tt-torch] An error occurred during compilation.

Please file a bug report by opening an issue at:
https://github.com/tenstorrent/tt-torch/issues/new

Include the following details in your report:
- The source code or model that triggered the error
- Steps to reproduce the issue
- Relevant logs or stack traces

Traceback (most recent call last):
  File "/localdev/staylor/tt-torch/tt-torch/demos/resnet/resnet50_demo.py", line 62, in <module>
    main()
  File "/localdev/staylor/tt-torch/tt-torch/demos/resnet/resnet50_demo.py", line 46, in main
    top5, top5_indices = torch.topk(tt_model(img).squeeze().softmax(-1), topk)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
    return forward_call(*args, **kwargs)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 574, in _fn
    return fn(*args, **kwargs)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
    return forward_call(*args, **kwargs)
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/_dynamo/convert_frame.py", line 1380, in __call__
    return self._torchdynamo_orig_callable(
  File "/localdev/staylor/tt-torch/tt-torch/env/venv/lib/python3.10/site-packages/torch/_dynamo/convert_frame.py", line 1164, in __call__
    result = self._inner_convert(
[...]

PR #808, opened 2025-05-21
Fix resnet demo and add it to CI

hshahTT:
### Ticket
[807](https://github.com/tenstorrent/tt-torch/issues/807)

### Problem description
A [previous PR](https://github.com/tenstorrent/tt-torch/pull/771) added a `devices` parameter to the `BackendOptions` objects without adding it to the actual `BackendOptions` class. That broke the normal resnet demo and the async demo.

### What's changed
- Replaced `device` with `devices` in the `BackendOptions` class. This fixes the normal resnet demo.
- Fixed the resnet async demo. Also, made the code more readable by removing an unnecessary loop.
- Added the normal resnet demo to On-PR and On-Push CI.
- [QoL] added `third_party/tt_forge_models` to the .gitignore file.

### Checklist

### DIFF
--- a/demos/resnet/resnet50_demo.py
+++ b/demos/resnet/resnet50_demo.py
@@ -1,11 +1,11 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+import argparse
 import torch
 from tt_torch.tools.utils import CompilerConfig
 from tt_torch.dynamo.backend import backend, BackendOptions
 from PIL import Image
-import torch
 from torchvision import transforms
 import torchvision.models as models
 import tabulate
@@ -13,7 +13,7 @@ import requests
 from tt_torch.tools.device_manager import DeviceManager
 
 
-def main():
+def main(run_default_img):
     weights = models.ResNet152_Weights.IMAGENET1K_V2
     model = models.resnet152(weights=weights).to(torch.bfloat16).eval()
     classes = weights.meta["categories"]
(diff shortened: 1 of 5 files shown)


kmabeeTT (approved):
Awesome - thanks! only takes 90 seconds on CI, great!

demos/resnet/resnet50_demo.py
```diff
@@ -29,11 +29,12 @@ def main():
 
     headers = ["Top 5 Predictions"]
     topk = 5
-    prompt = 'Enter the path of the image (type "stop" to exit or hit enter to use a default image): '
-    img_path = input(prompt)
-    while img_path != "stop":
+
+    DEFAULT_URL = "http://images.cocodataset.org/val2017/000000039769.jpg"
```
AleksKnezevic, line 33: @ssaliceTT, when you add the helper to download images to cache, please include this too.

AleksKnezevic (approved):

demos/resnet/resnet50_demo.py
```diff
@@ -55,8 +56,23 @@ def main():
...
 
+    if run_default_img:
+        print("Running with default image URL: ", DEFAULT_URL)
+        process_image("")
+    else:
+        prompt = 'Enter the path of the image (type "stop" to exit or hit enter to use a default image): '
         img_path = input(prompt)
+        while img_path != "stop":
+            process_image(img_path)
+            img_path = input(prompt)
```
jameszianxuTT, line 67: This doesn't matter for this PR, but I think it's a bit wild that we're controlling the input path to the image via a python `input()` call... 

I think it's weird in general to mix control flow via CLI arguments + interactive input, and I'd like to see this moved to a CLI argument as well with logic like:
- If run without args, use default image
- If run with run default image specified, always use default image
- If run with custom image path CLI arg specified, use input this arg instead of the default
- Disallow running with both `run default image` and `custom image path`

PR #809, opened 2025-05-21
[DO NOT MERGE] 

jmcgrathTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


jameszianxuTT:
Just as fyi for others on this ticket

Something strange happens when attempting to use run-tests.yml on a shared runner host - everything looks intact (container, system/python dependencies, paths etc.) but pytest cannot discover any paths. 

The only difference I note is that the default `.cache` folder doesn't seem accessible to us for some reason on the machines (maybe it's mounted on host?), which could have higher order effects for code that expects it to be writable.

PR #812, opened 2025-05-21
Move system_desc creation to Executor

brataTT:
### Ticket
To unblock [Data parallelization PR](https://github.com/tenstorrent/tt-torch/pull/764)

### Problem description
System desc is currently created by pytest fixture.
We want to tie lifetime of system desc to lifetime of Executor instead.
This is so that correct device config can be used to create system desc.

### What's changed
- System desc is created at Executor __init__ instead of pytest fixture
- Remove dead code
  - Old system desc creation code
  - Unused compile flow
- misc fixes

### Checklist

Checked locally that string "linux-gnu" does not exist in compiled IR.
Since only hardcoded system desc contains this string.
This is to make sure we're generating the real system desc.

Using the following tests to cover each flow
- torch
- onnx
- multidevice
- op_by_op_torch
- op_by_op_stablehlo
#### flows to smoke test system desc 
```sh
TT_TORCH_IR_LOG_LEVEL=DEBUG pytest -svv tests/torch/test_conv2d.py |& tee logs/conv.log
TT_TORCH_IR_LOG_LEVEL=DEBUG pytest -svv tests/onnx/test_basic.py::test_add |& tee logs/onnx.log
TT_TORCH_IR_LOG_LEVEL=DEBUG pytest -svv tests/models/distilbert/test_distilbert.py::test_distilbert_multiloop[full-distilbert-base-uncased-eval-64] |& tee logs/multi.log
TT_TORCH_IR_LOG_LEVEL=DEBUG pytest -svv tests/models/mnist/test_mnist.py::test_mnist_train[op_by_op_torch-eval] |& tee logs/mnist.log
TT_TORCH_IR_LOG_LEVEL=DEBUG pytest -svv tests/models/MobileNetV2/test_MobileNetV2_onnx.py::test_MobileNetV2_onnx[op_by_op_stablehlo-eval] |& tee logs/mblnet.log
```

##### report if has correct system desc
```sh
logfiles="logs/conv.log logs/onnx.log logs/multi.log logs/mnist.log logs/mblnet.log"
for logfile in $logfiles; do
    echo "Log file: $logfile"
    
    # expect to be non-zero
    echo "Has system descriptor: $(grep -c 'system_desc' $logfile)"
    
    # expect to be zero
    echo "Has fake system descriptor: $(grep -c linux-gnu $logfile)"
done
```

### DIFF
--- a/tests/torch/test_constant_fold.py
+++ b/tests/torch/test_constant_fold.py
@@ -60,7 +60,6 @@ def test_interp():
     cc = CompilerConfig()
     cc.enable_consteval = True
     cc.consteval_parameters = True
-    cc.compile_depth = CompileDepth.TORCH_FX
     verify_module(Basic(), inputs=[small], compiler_config=cc)
 
 
--- a/env/activate
+++ b/env/activate
@@ -64,4 +64,3 @@ fi
 export PYTHONPATH="$(pwd):$(pwd)/env/venv:$(pwd)/env/venv/lib:$(pwd)/install/lib:$(pwd)/install:$(pwd)/.local/toolchain/python_packages/mlir_core:${TT_METAL_HOME}:${TT_METAL_HOME}/tt_eager:${TT_METAL_BUILD_HOME}/tools/profiler/bin"
 export ARCH_NAME="${ARCH_NAME:-wormhole_b0}"
 export TT_METAL_LOGGER_LEVEL="ERROR"
-export SYSTEM_DESC_PATH="/tmp/system_desc.ttsys"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,21 +38,6 @@ def manage_dependencies(request):
     )
 
 
-@pytest.fixture(scope="session", autouse=True)
-def create_system_descriptor(request):
-    try:
-        tt_mlir.create_system_desc()
-    except Exception as e:
-        print(f"Failed to create system descriptor: {e}")
-
-    yield
-
-    try:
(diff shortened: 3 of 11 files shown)


AleksKnezevic (approved):

tapspatel (approved):
thanks @brataTT for the quick fix!

jameszianxuTT (approved):

brataTT:
Thanks for the lightning fast reviews 🙏 
I'll set to auto-merge since CI looks busy

ddilbazTT (approved):

tests/torch/test_constant_fold.py
```diff
@@ -60,7 +60,6 @@ def forward(self, x):
     cc = CompilerConfig()
     cc.enable_consteval = True
     cc.consteval_parameters = True
-    cc.compile_depth = CompileDepth.TORCH_FX
```
jameszianxuTT, line 63: Won't this still run up to EXECUTE if unspecified?

brataTT: Looks like it already was [[Filed issue to investigate]](https://github.com/tenstorrent/tt-torch/issues/813)
We call [[torch_to_shlo]](https://github.com/tenstorrent/tt-torch/blob/ef5b97255520bd057241811c43d99688b07bb138/tt_torch/dynamo/backend.py#L162), never check against `CompileDepth.TORCH_FX` and just keep going to `EXECUTE`
We didn't know before because the test actually passes `EXECUTE` 😅

brataTT: image

PR #815, opened 2025-05-21
Skip onPR build + test for more file types and add support for #skip-onpr-tests in commit description

kmabeeTT:
### Ticket
None

### Problem description
- Sometimes CI gets bogged down building and running for changes that don't even affect build + test in onPR|
- Sometimes there are PR feedback for new models being added to Nightly, and every minor tweak requires full rerun of onPR CI which doesn't even test the model.
- Would be great to extend existing pattern of files that can skip CI, and add way to skip on demand by user.

### What's changed
- Extend list of skip patterns to cover bunch of .yml files only used in nightly/weekly and some scripts
- Add flag #skip-onpr-tests which can be put in commit description by user to skip build + test if they are sure there change is no risk and doesn't impact CI.  Github docs show how other strings are used to skip entire CI, but we want to keep lint, spdx check etc in place.

### Checklist
- [x] Test on branch, screenshots links in PR

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -45,9 +45,24 @@ jobs:
           all_match=true
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             CHANGED_FILES=$(gh pr diff ${{ github.event.pull_request.number }} --name-only --repo ${{ github.repository }})
+
+            # Define patterns for files that should skip the build
+            SKIP_PATTERNS=(
+              '^.*\.md$'
+              '^.*\.gitignore$'
+              '^.*LICENSE.*$'
+              '^\.github/workflows/.*nightly.*\.yml$'
+              '^\.github/workflows/.*weekly.*\.yml$'
+              '^\.github/workflows/.*depth.*bench.*\.yml$'
+              '^\.github/workflows/fail_inspector.yml$'
+              '^results/parse_op_by_op_results\.py$'
+            )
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Verification to show we can now modify more files in skip list, and build+test is skipped ([link](https://github.com/tenstorrent/tt-torch/actions/runs/15174883608?pr=815)).

image

kmabeeTT:
Now onPR on branch runs for 1 minute when modifying files unrelated to onPush/onPR:

image

... instead of 35 minutes:

image

Which can be useful sometimes.

kmabeeTT:
Verification showing new message in latest commit description works to skip build + test ([link](https://github.com/tenstorrent/tt-torch/actions/runs/15175489301?pr=815)):

image

needed to use the PR head commit rather than detached head commit (which is a merge commit) that PR's create.

jameszianxuTT (approved):

jameszianxuTT:
Thanks for the detailed verification screenshots! 

As an aside, the fact that this works + tracing through the logic, I think that last check-all-green job might be doing nothing. It's at best redundant to the mechanism that Github uses to annotate this changeset with "Changes approved"

kmabeeTT:
> Thanks for the detailed verification screenshots!
> 
> As an aside, the fact that this works + tracing through the logic, I think that last check-all-green job might be doing nothing. It's at best redundant to the mechanism that Github uses to annotate this changeset with "Changes approved"

I mean, it's still doing a thing, to check that the remaining jobs (pre-commit, spdx, docker-build, ) pass right? If they fail, check-all-green would fail and block PR.

image

PR #817, opened 2025-05-22
move tt forge models install to non-toolchain build branch

jameszianxuTT:
### Ticket
None

### Problem description
Using ExternalProject_add(tt-forge-models...) occurs as a side effect of building the toolchain, which will break because the toolchain build branch doesn't include(ExternalProject).

### What's changed
Move tt-forge-models fetch to inside the non-toolchain build path.

### Checklist

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -116,17 +116,16 @@ else()
 
     install(DIRECTORY ${TORCH_MLIR_INSTALL_PREFIX}/python_packages/torch_mlir/torch_mlir DESTINATION "${CMAKE_INSTALL_PREFIX}" USE_SOURCE_PERMISSIONS)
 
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
TIL how this works

brataTT:
Thanks for running an On PR against metal uplift [[CI]](https://github.com/tenstorrent/tt-torch/actions/runs/15194276870/job/42734464148)
This run is building a new image, so should be good for testing both changes
- if image build passes, this torch change is good
- if rest passes, metal uplift is good
-

kmabeeTT (approved):
Crap, I don't know why I put it outside if/else... thanks for fix!

brataTT (approved):
Looks like new image build passed 🙌
But torch build failed due to uplift as expected

Issue #780, opened 2025-05-15
refactor multichip tests


AleksKnezevic:
Right now there are some multichip tests that run in `.github/workflows/run-full-model-execution-tests.yml::eval6`. These should be moved into their own test group probably within the run multidevice tests yml. 

@jameszianxuTT can help with how to structure the change.

PR #819, opened 2025-05-22
Add data parallel support to ModelTester, and some data parallel tests to CI

hshahTT:
### Ticket
[590](https://github.com/tenstorrent/tt-torch/issues/590)

### Problem description
Need to add support to run our test models in full execution mode using data parallel execution.

### What's changed
At a high level this PR modifies the `ModelTester` class to use the async mode functions added in [PR 696](https://github.com/tenstorrent/tt-torch/pull/696) to run the same input on multiple predefined devices in parallel.

Specifically:

- **tests/utils.py**:
    - Added two new member parameters to the `ModelTester` class:
        - `devices`: An optional list of devices to run the inputs on. If no device list is provided, the tester object will acquire all devices on the board and use them.
        - `data_parallel_mode`: Boolean, if True, will run the input on all of the provided devices.
    - Added a function to compile the given framework model for all provided devices (`compile_models_for_data_parallel()`)
    - Added a function (`_test_model_eval_data_parallel()`) to run the inputs asynchronously on all devices, and compare each output to the golden.
        - As a result of this, the `test_model()` function will return a list of outputs (one for each provided device) when running the test in data parallel mode.

- **tests/models/**:
    - Parameterized the MNIST, ResNet-50, and ResNet-18 model tests with a new `data_parallel_mode` parameter.

- **.github/workflows/run-full-model-execution-tests.yml**:
    - Added the data parallel versions of the MNIST, ResNet-50, and ResNet-18 model tests to the on-pr and on-push workflows. 

- All other file changes are no-ops and were made to ensure that they reflected the new `data_parallel_mode` parameter in the model test files.

### Notes
- Op by Op tests and runtime intermediate verification are not supported with data parallel mode.
- Once this PR is approved and submitted, I'll add more PRs to add other models to CI.

### Checklist

### DIFF
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -27,7 +27,7 @@ env:
 
 jobs:
   tests:
-    timeout-minutes: 63 # current onPR time runs ~62-63m
+    timeout-minutes: 90
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/run-full-model-execution-tests.yml
+++ b/.github/workflows/run-full-model-execution-tests.yml
@@ -27,81 +27,86 @@ env:
 
 jobs:
   tests:
-    timeout-minutes: 60 # current onPR time runs ~60m
+    timeout-minutes: 63 # current onPR time runs ~62-63m
     strategy:
       fail-fast: false
       matrix:
         build: [
-          # {
-          #   runs-on: wormhole_b0, name: "eval_1", tests: "
-          #         tests/models/autoencoder_linear/test_autoencoder_linear.py::test_autoencoder_linear[full-eval]
-          #         tests/models/distilbert/test_distilbert.py::test_distilbert[full-distilbert-base-uncased-eval]
-          #         tests/models/mlpmixer/test_mlpmixer.py::test_mlpmixer[full-eval]
-          #         tests/models/yolov3/test_yolov3.py::test_yolov3[full-eval]
-          #         tests/models/yolov4/test_yolov4.py::test_yolov4[full-eval]
-          #   "
-          # },
-          # {
-          #   runs-on: wormhole_b0, name: "eval_2", tests: "
-          #         tests/models/mnist/test_mnist.py::test_mnist_train[full-eval]
-          #         tests/models/MobileNetV2/test_MobileNetV2.py::test_MobileNetV2[full-eval]
-          #         tests/models/openpose/test_openpose_v2.py::test_openpose_v2[full-eval]
-          #         tests/models/resnet/test_resnet.py::test_resnet[full-eval]
-          #         tests/models/resnet50/test_resnet50.py::test_resnet[full-eval]
-          #         tests/models/distilbert/test_distilbert.py::test_distilbert_multiloop[full-distilbert-base-uncased-eval-64]
-          #         tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vit_l_16]
-          #         tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vit_b_16]
-          #         tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vit_l_32]
-          #         tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[full-eval-vit_b_32]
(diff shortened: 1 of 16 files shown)


hshahTT:
@kmabeeTT So far I've only added the MNIST, Resnet-50, and Resnet-18 models to the On-PR/On-Push CI since they don't take long to run (about ~30s - 1 min each). Could you let me know what models are worth adding to the nightlies and/or weeklies CIs? (i.e. which of the models in those workflows are the most important)

Also, how many data parallel tests should we aim to have in the CIs in total? I don't wanna make the workflows take too long.

tests/models/mnist/test_mnist.py
```diff
@@ -74,17 +78,25 @@ def test_mnist_train(record_property, mode, op_by_op):
     cc.enable_consteval = True
     cc.consteval_parameters = True
     if op_by_op:
+        if data_parallel_mode:
```
ddilbazTT, line 81: Is data parallel ever going to be supported during op-by-op? If never, maybe there should be an assert somewhere, probably `backend.py`

tests/models/mnist/test_mnist.py
```diff
@@ -74,17 +78,25 @@ def test_mnist_train(record_property, mode, op_by_op):
...
+            pytest.skip("Op-by-op not supported in data parallel mode")
         cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
         if op_by_op == OpByOpBackend.STABLEHLO:
             cc.op_by_op_backend = OpByOpBackend.STABLEHLO
-
+    devices = None
+    if data_parallel_mode:
```
ddilbazTT, line 87: Can't we do these in the ModelTester?

hshahTT: We could. I wanted to give more flexibility to the user on selecting which devices to use, rather than just defaulting to all devices. For example, when we expand this to LLMBox tests, then we can write tests that run Model A on 4 devices and Model B on the remaining 4 devices.

tests/models/mnist/test_mnist.py
hshahTT: I had previously added an assert in Executor's constructor: https://github.com/tenstorrent/tt-torch/blob/813ea5d7a0ca5d4975dedd1c996570f1d979a1e9/tt_torch/dynamo/executor.py#L188

hshahTT:
@ddilbazTT @jameszianxuTT I modified the `ModelTester` class so that the device list is now optional. If no device list is provided, the tester object will just acquire all devices on the board and use them for data parallel. I also updated the PR description to reflect this.

hshahTT:
Single device vs Data Parallel Benchmarking results (10 testing runs each):

**MNIST Test:**
image

**Resnet 18 Test:**
image

**Resnet 50 Test:**
image

.github/workflows/run-full-model-execution-tests.yml
```diff
@@ -27,7 +27,7 @@ env:
 
 jobs:
   tests:
-    timeout-minutes: 60 # current onPR time runs ~60m
+    timeout-minutes: 63 # current onPR time runs ~62-63m
```
AleksKnezevic, line 30: you should have some leeway here, I would go with 90

AleksKnezevic (approved):

jameszianxuTT (approved):

ddilbazTT (commented):
I have a few remarks to add. Thanks for addressing the PR comments

jameszianxuTT:
Crap, I see adding this extra parameterization makes the assumption we had before that cli option --op_by_op_* would run the single op by op test defined for the model, but now it splays out into +/- data parallel

We're gonna end up with a combinatorial explosion of different parallelization / multichip modes later - I am fine with just getting this PR in but we'll need to think of a proper testing strategy for multichip tests #780 . 

I'll push a quick fix to unblock this pr as we have some other stuff coming down the pipeline @hshahTT

tests/models/mnist/test_mnist.py
```diff
@@ -10,6 +10,7 @@
 import torch.nn.functional as F
 from tests.utils import ModelTester
 from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+from tt_torch.tools.device_manager import DeviceManager
```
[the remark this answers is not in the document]
hshahTT: Good catch! Removed it

vmilosevic:

PR #822, opened 2025-05-23
Clean wheel build

AleksKnezevic:
### Problem description
tt-torch wheel was not useable without some hacky overrides like setting PYTHONPATH and LD_LIBRARY_PATH. It also did not pull in all requirements needed. 

### What's changed
Modified wheel build to place needed libs in the correct place so path doesn't need to be manually set. Added requirements. Added decorators to verify_module and test_model to reset torch dynamo. This was previously done in conftest, but we want to be able to run test_basic.py without the whole repo. I had to increase atol for one of the tests, without conftest it fails since random seed does not get reset. 

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ class install_metal_libs(install_lib):
                 "tools",
             )
         )
-        dest_tools_dir = os.path.join(self.install_dir, "tt_metal", "tools")
+        dest_tools_dir = os.path.join(self.install_dir, "tt-metal", "tools")
         if os.path.exists(src_tools_dir):
             os.makedirs(os.path.dirname(dest_tools_dir), exist_ok=True)
             shutil.copytree(src_tools_dir, dest_tools_dir, dirs_exist_ok=True)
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -119,14 +141,17 @@ def run(self):
...
     cmake_args=cmake_args,
     cmdclass={
         "install_lib": install_metal_libs,
     },
     zip_safe=False,
     install_requires=[
         "torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl",
+        "stablehlo@https://github.com/openxla/stablehlo/releases/download/v1.0.0/stablehlo-1.0.0.1715728102%2B6051bcdf-cp310-cp310-linux_x86_64.whl",
         "numpy",
+        "onnx",
```
jameszianxuTT, line 153: Shouldn't we lock down the onnx version here as well? It is pinned in the [requirements.txt](https://github.com/tenstorrent/tt-torch/blob/437ed8e2be09cd465c0563e3619dd80bf316446c/requirements.txt#L27)

AleksKnezevic: yeah, good point.

ddilbazTT (approved):

LPanosTT (approved):

setup.py
```diff
@@ -111,22 +133,25 @@ def run(self):
...
     long_description_content_type="text/markdown",
-    include_package_data=True,
     cmake_args=cmake_args,
     cmdclass={
         "install_lib": install_metal_libs,
     },
     zip_safe=False,
     install_requires=[
         "torch@https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.6.0%2Bcpu.cxx11.abi-cp310-cp310-linux_x86_64.whl",
+        "stablehlo@https://github.com/openxla/stablehlo/releases/download/v1.0.0/stablehlo-1.0.0.1715728102%2B6051bcdf-cp310-cp310-linux_x86_64.whl",
```
brataTT, line 151: Do torch and mlir stablehlo version need to be aligned?

If so, the PR doesn't seem to be rebased on top of the LLVM uplift.
Any chance of a stablehlo version mismatch here?

ddilbazTT: That's a good point, but I don't think there will be version clashes here. We've been using the latest stablehlo official release: https://github.com/openxla/stablehlo/releases/tag/v1.0.0 We just moved it from `requirements.txt` to here.

AleksKnezevic: Yeah, I've replicated some reqs from requirements.txt here so they automatically get pulled in when the user pip installs the wheel. This allows you to just install the wheel and run.

jameszianxuTT (approved):

PR #841, opened 2025-05-28
Move e2e-compile full-deepseek-ai/DeepSeek-R1-Distill-Qwen-32B-eval to own group

kmabeeTT:
### Ticket
None

### Problem description
 - This model is often (not always) dying (reason unknown), first seen on May 17 b062ba9 nightly run
 - Suspected out-of-memory, but locally I see segfault on tt_mlir.compile_ttir_to_bytestream() call from backend.py
 - Needs debug. op-by-op for same model also dies lately.

### What's changed
- Move it to it's own group so it doesn't take down entire group and kill reporting

### Checklist
- [x] Run on branch isolated, seems like only test killed


jameszianxuTT (approved):
Thanks, good idea to isolate.

PR #848, opened 2025-05-28
Adding individual block test 

AleksKnezevic:
Added a test to run an individual transformer block from stable diffusion 3.5 medium

### DIFF
--- a/tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
+++ b/tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
@@ -3,10 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 import torch
 import pytest
-from tests.utils import ModelTester, skip_full_eval_test
+from tests.utils import ModelTester
 from diffusers import (
-    StableDiffusion3Pipeline,
-    FlowMatchEulerDiscreteScheduler,
     SD3Transformer2DModel,
 )
 
@@ -55,7 +53,7 @@ model_info_list = [
     model_info_list,
     ids=[model_info[0] for model_info in model_info_list],
 )
-def test_stable_diffusion_transformer(record_property, model_info, mode):
+def test_stable_diffusion_transformer_block(record_property, model_info, mode):
     model_group = "red"
     model_name, model_path = model_info
 
--- a/tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
+++ b/tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
@@ -1,0 +1,77 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+from tests.utils import ModelTester, skip_full_eval_test
+from diffusers import (
+    StableDiffusion3Pipeline,
+    FlowMatchEulerDiscreteScheduler,
+    SD3Transformer2DModel,
+)
+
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Do we plan to run this in nightly CI at some point (if yes may need to tweak model name to not conflict with existing test)?

AleksKnezevic:
@kmabeeTT, wasn't planning on it, but changed it just in case.

kmabeeTT (approved):

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "model_info",
+    model_info_list,
+    ids=[model_info[0] for model_info in model_info_list],
+)
+def test_stable_diffusion_transformer(record_property, model_info, mode):
```
ddilbazTT, line 58: Maybe we should change this name to `test_stable_diffusion_transformer_block`

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
```diff
@@ -0,0 +1,77 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+from tests.utils import ModelTester, skip_full_eval_test
+from diffusers import (
+    StableDiffusion3Pipeline,
```
ddilbazTT, line 8: nit: we are not using the pipeline or scheduler so we could have removed it for cleanliness

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
```diff
@@ -0,0 +1,77 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import pytest
+from tests.utils import ModelTester, skip_full_eval_test
```
ddilbazTT, line 6: nit: we are not using `skip_full_eval_test` I am assuming this runs end-to-end

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
```diff
@@ -0,0 +1,77 @@
...
+            "hidden_states": hidden_states,
+            "encoder_hidden_states": encoder_hidden_states,
+            "temb": temb,
+            "joint_attention_kwargs": joint_attention_kwargs,
+        }
+        return arguments
+
+
+model_info_list = [
+    ("SD3.5-medium-transformer-block", "stabilityai/stable-diffusion-3.5-medium"),
```
ddilbazTT, line 45: nit: we could add large too

ddilbazTT (approved):
Test looks good, but why are we adding this? Out of curiosity, since it's really similar to the test we're already running

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
AleksKnezevic: I didn't look up the input shapes for large, I assume they'll be different.

tests/models/stable_diffusion/test_stable_diffusion_transformer_block.py
AleksKnezevic: fixed.

PR #851, opened 2025-05-29
Support for model execution through EmitC path

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/2870

### Problem description
Beside `tt-mlir` runtime, we also have `TTNN->EmitC` which generates C++ file which calls TTNN ops directly through TTNN C++ API. We want to enable that path for testing of models.

### What's changed
Added infra which enables E2E compilation and execution of model through EmitC path.

Summary of few design decisions that should be considered:
- Added a new target in `CompileDepth`: `EXECUTE_CPP`
- Currently `EXECUTE_CPP` is not an independent execution target; it still requires execution in the runtime, and then compilation and execution of shared object where the outputs of the two are compared if they match exactly. In the future we probably want those two to be independent, and compare each with  a `PyTorch` golden with same accuracy parameters, but as TTNNEmitC is still under development this can give as better feedback at the moment about the things that are potentially different in both paths.
- The way to run the model test in `EXECUTE_CPP` mode is with envvar `TT_TORCH_EXECUTION_DEPTH=EXECUTE_CPP` which should override `eval-full` pytest target to run in this mode instead of `EXECUTE`. Other pytest targets should behave the same as before. Example to test: `TT_TORCH_EXECUTION_DEPTH=EXECUTE_CPP pytest -ssv tests/models/resnet50`

TODO:
- Add docs once we agree on the API.
- Most models are currently supported, blockers are linked in the ticket issue (`resnet`/`resnet50` should work); once those issues are resolved I expect most if not all models that currently run with `tt-mlir` runtime to execute in EmitC path as well. When those issues are resolved and I veirfy locally that those models run successfully, we should consider approach of adding it to CI.

### Checklist


tt_torch/dynamo/executor.py
```diff
@@ -363,20 +366,37 @@ def __call__(self, *inputs):
...
                 outputs = tt_mlir.run(
                     device,
                     binary,
                     program_idx,
-                    (preprocessed_weights + preprocessed_activations),
+                    inputs,
+                )
+
+            if self.compiler_config.compile_depth == CompileDepth.EXECUTE_CPP:
+                so_path = self.mcg.so_paths[device_idx]
```
AleksKnezevic, line 386: The executor class is returned to the user when torch.compile is called. When it's run, outputs are expected. You should get the outputs here and return them, like 'EXECUTE' depth does. These are wrapped in test functions which will run the pytorch model on host and on device and verify the accuracy of the outputs.

AleksKnezevic (changes requested):
Thanks for the PR! I made a couple of comments regarding returning outputs and where to se overrides, otherwise it's good. 

I don't see any tests where EXECUTE_CPP is enabled. You should add some in `test_basic.py`

tests/utils.py
```diff
@@ -572,9 +575,10 @@ def remap_compile_depth(self, compile_depth, min_pcc):
             CompileDepth.COMPILE_OP_BY_OP: "INCORRECT_RESULT",
             CompileDepth.EXECUTE_OP_BY_OP: "INCORRECT_RESULT",
             CompileDepth.EXECUTE: "INCORRECT_RESULT",  # ambiguous between incorrect / passed
+            CompileDepth.EXECUTE_CPP: "INCORRECT_RESULT",  # ambiguous between incorrect / passed
         }
 
-        if compile_depth is not CompileDepth.EXECUTE:
+        if compile_depth not in (CompileDepth.EXECUTE, CompileDepth.EXECUTE_CPP):
```
jameszianxuTT, line 581: I think this might confuse our reporting - Would a model compiling to `EXECUTE_CPP` be considered a distinct model than one compiling to just EXECUTE? 

I don't think would be considered a distinct model, so I think we should:
1. Keep this branch as        `if compile_depth is not CompileDepth.EXECUTE:`
2. Keep the mapping of CompileDepth.EXECUTE_CPP to "INCORRECT_RESULT"

jameszianxuTT (changes requested):

tests/utils.py
AleksKnezevic: It should not be considered a distinct model.

Issue #733, opened 2025-05-08
Cannot get the device from a tensor without an allocated buffer - hit in phi-2 priority model and others


kmabeeTT:
phi-2 model (CI [link](https://github.com/tenstorrent/tt-torch/actions/runs/14721026773/job/41317831019#step:10:397)) and others hit this error when getting output tensor from device (suspiciously early, only after 30 sec of running) - needs debug.

```
>       outputs = tt_mlir.run(device, binary, program_idx, preprocessed_inputs)
E       RuntimeError: TT_THROW @ /__w/tt-torch/tt-torch/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/tensor/tensor.hpp:325: tt::exception
E       info:
E       Cannot get the device from a tensor without an allocated buffer
E       backtrace:
E        --- /__w/tt-torch/tt-torch/env/venv/lib/libTTMLIRRuntime.so(+0x16eda8) [0x7fec1496eda8]
E        --- /__w/tt-torch/tt-torch/env/venv/lib/_ttnn.so(+0xaf64d8) [0x7fec129ad4d8]
E        --- tt::tt_metal::Tensor tt::tt_metal::tensor_impl::to_host_helper<int>(tt::tt_metal::Tensor const&, bool, tt::stl::StrongType<unsigned char, ttnn::QueueIdTag>)
E        --- tt::tt_metal::Tensor tt::tt_metal::tensor_impl::to_host<int>(tt::tt_metal::Tensor const&, bool, tt::stl::StrongType<unsigned char, ttnn::QueueIdTag>)
E        --- /__w/tt-torch/tt-torch/env/venv/lib/_ttnn.so(_ZN2tt8tt_metal11tensor_impl8dispatchIZNS1_15to_host_wrapperIJRNS0_6TensorERbRNS_3stl10StrongTypeIhN4ttnn10QueueIdTagEEEEEEDaDpOT_EUlTyDpOT0_E_JS5_S6_SC_EEEDaNS0_8DataTypeEOT_SI_+0xb6) [0x7fec13d9ffc6]
E        --- /__w/tt-torch/tt-torch/env/venv/lib/_ttnn.so(+0x1ee8caa) [0x7fec13d9fcaa]
E        --- /__w/tt-torch/tt-torch/env/venv/lib/libtt_metal.so(+0x19894a) [0x7fec1455794a]
[...]

PR #854, opened 2025-05-29
Fix tt_mlir.run() bug w/ dealloc before read output_rt when tensor is returned multiple times in graph

kmabeeTT:
### Ticket
#733 

### Problem description
- Some models (phi-2, codegen, vilt) have graphs that return same tensor multiple times
- Logic in `ttmlir.run()` wasn't safe here, would iterate over outputs and read+dealloc, leading to assert firing when already deallocated tensor is attempted to be to read back to host.

### What's changed
- Simple fix, don't deallocate rt_output tensors until they are all read back to host first
- Promote Codegen and Phi-2 to full-model-execute, no longer hitting bug

### Checklist
- [x] Error is gone in 3 affected models (phi-2, codegen, vilt)


kmabeeTT:
My first thought was I better update `ttmlir.run_async()` too, and try test in data parallel mode, but looks like that function does not deallocate tensors so it doesn't have this same issue. @hshahTT is that expected that it doesn't deallocate tensors?

kmabeeTT:
Forced push after rebase. DRAM OOM issues mentioned in failing reasons here might be resolved, am checking and will update, though core fix in this PR remains unaffected.

vmilosevic:

tt_torch/csrc/bindings.cpp
```diff
@@ -292,12 +292,18 @@ std::vector<at::Tensor> run(tt::runtime::Device device,
...
   std::vector<at::Tensor> outputs;
   outputs.reserve(rt_outputs.size());
 
   for (size_t i = 0; i < rt_outputs.size(); ++i) {
     auto &rt_output = rt_outputs.at(i);
     outputs.emplace_back(create_torch_tensor(rt_output));
+  }
+
+  for (size_t i = 0; i < rt_outputs.size(); ++i) {
+    auto &rt_output = rt_outputs.at(i);
```
sshonTT, line 306: Would it be possible to merge this loop with the one starting at line 300, right after the emplace_back call?

kmabeeTT: Yes but then it brings back the bug :) - they were merged and my change is to unmerge them so that all the reads happen first, and then the deallocates rather than read, dealloc, read, dealloc.

Issue #864, opened 2025-05-30
Fix Mistral/ Pixtral models after torch uplift


ddilbazTT:
Mistral/ Pitxral tests are failing after torch+transformers uplift:
```
pytest -s tests/models/mistral/test_mistral.py::test_mistral[full-ministral3b-eval]
E                   RuntimeError: "bitwise_or_cpu" not implemented for 'BFloat16'
```

I will add sub-issues to separate each test error.

PR #865, opened 2025-05-30
Torchaudio

ddilbazTT:
### Ticket
None

### Problem description
We need to uplift PyTorch 2.6 to 2.7

### What's changed
PyTorch uplift requires Transformers uplift: https://github.com/pytorch/pytorch/issues/141196 
Generative models return DynamicCache object, which is now flattened and appended to outputs
Added a torchaudio model (`tests/models/seamless_m4t/test_seamless_m4t.py`)
Fixed `tests/models/speecht5_tts/test_speecht5_tts.py` which now runs
Removed `set_model_eval` which has been outdated for a while now
Quarantined mistral tests since they fail with changes in this PR, opened an issue: https://github.com/tenstorrent/tt-torch/issues/864 
Changed Model MLIR file creation and artifact upload:
- at the start, we delete any file that might have been created from another run
- instead of writing to a new file, we append MLIR since models usually have more than 1 subgraph after `torch.compile`
- MLIR artifact is merged even if the test whose artifact is used failed. For example, if one of the nightly execution tests would fail, this was disabling merging of nightly execution tests even if others created MLIR files. Now, that won't be the case.

Worked on this PR with @jameszianxuTT

### DIFF
--- a/.github/Dockerfile.ci
+++ b/.github/Dockerfile.ci
@@ -25,16 +25,6 @@ WORKDIR $BUILD_DIR/$PROJECT_NAME
 RUN cd third_party && \
     cmake -B toolchain -DBUILD_TOOLCHAIN=ON
 
-# Build torchvision wheel and copy it to /dist
-RUN git clone https://github.com/pytorch/vision.git && \
-    cd vision && \
-    git checkout v0.21.0 && \
-    TORCHVISION_USE_VIDEO_CODEC=0 TORCHVISION_USE_FFMPEG=0 CC=clang \
-    CXX=clang++ _GLIBCXX_USE_CXX11_ABI=1 USE_CUDA=OFF python3.10 setup.py bdist_wheel && \
-    cp dist/*.whl $TTTORCH_DIST_DIR && \
-    cd .. && \
-    rm -rf vision
-
 # Final stage
 FROM ghcr.io/tenstorrent/tt-torch/tt-torch-base-ubuntu-22-04:${FROM_TAG} AS ci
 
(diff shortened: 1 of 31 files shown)


ddilbazTT:
Asked everybody's review since rebasing after this change will require a fresh environment. 
Running op-by-op tests: https://github.com/tenstorrent/tt-torch/actions/runs/15349175587

AleksKnezevic (approved):
Couple of nits regarding not removing/comment out tests in yml, but adding a pytest.skip with a link to an issue, otherwise it looks good to me.

vmilosevic:

AleksKnezevic (approved):
Awesome, thanks!

.github/Dockerfile.ci
```diff
@@ -25,16 +25,6 @@ WORKDIR $BUILD_DIR/$PROJECT_NAME
 RUN cd third_party && \
     cmake -B toolchain -DBUILD_TOOLCHAIN=ON
 
-# Build torchvision wheel and copy it to /dist
```
jazpurTT, line 28: Change looks good. So we are moving the torchvision install to the env activation, just curious why?

jazpurTT (approved):

.github/Dockerfile.ci
jameszianxuTT: I think it's because don't need to source build torchvision anymore since we don't need to force cxx11 compat since the upgraded torchvision now comes with that compatibility (since the upgraded torch comes with that compatibility) 

So now we can just pip install it

hshahTT (approved):

jameszianxuTT (approved):

ddilbazTT:
> Nit, but maybe you want to update the `getting_started.md` doc to remove the torchvision install note and update the torch wheel link from torch 2.5.0 (I guess this hasn't been updated in a while) to 2.7.0

I couldn't find this in `getting_started.md` or `build.md`, is it somewhere else?

Issue #870, opened 2025-05-30
run_generate=False


AleksKnezevic:
Most of our generative models currently run with `run_generate=True`. This runs the generative loop on device, compiling and executing multiple times. Additionally, this returns a stream of tokens which is difficult to validate. 

Change all the tests to run with `run_generate=False`. Add one additional test that runs `run_generate=True` but doesn't validate token output.

PR #901, opened 2025-06-05
run_generate=False

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/870

### Problem description
Change generative model tests to run with `run_generate=False` so that PCC reflects accuracy of logits and not tokens. Add an additional test with `run_generate=True`. 

### What's changed
All instances of `run_generate=False` switched to `run_generate=True` causing the model to only run a forward pass and not the generative loop. `is_token_output=True` was removed from the model testers along with `assert_eval_token_mismatch=False` in test_model calls. In most cases `assert_atol=False` and `use_cache=False` were added.

This effected codegen, falcon3, flan_t5, got_neo, mamba, phi, qwen2_casual_lm, seamless_m4t and t5 models. A new test test_codegen_generate.py was created containing the codegen model with `run_generate=True` and added to the nightly workflow.

Some noteworthy changes:
- Seamless_m4t required a monkey-patch to avoid a `torch.rand()` call in the models forward pass. Hidden layers were significantly reduced to allow the model to fit in memory. PCC is less than 0.5 and so `assert_pcc` was disabled.
- qwen_casual_lm had required PCC lowered to 0.98 `required_pcc=0.98` and mamba lowered to 0.95 `required_pcc=0.95`

### Checklist

### DIFF
--- a/tests/models/codegen/test_codegen_generate.py
+++ b/tests/models/codegen/test_codegen_generate.py
@@ -1,0 +1,62 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+# Reference: https://huggingface.co/docs/transformers/model_doc/codegen#usage-example
+
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import pytest
+from tests.utils import ModelTester
+import torch
+from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
+
+
+class ThisTester(ModelTester):
+    def _load_model(self):
+        checkpoint = "Salesforce/codegen-350M-mono"
+        model = AutoModelForCausalLM.from_pretrained(
+            checkpoint, torch_dtype=torch.bfloat16
+        )
+        self.tokenizer = AutoTokenizer.from_pretrained(checkpoint)
+        return model
+
+    def _load_inputs(self):
+        text = "def hello_world():"
+        inputs = self.tokenizer(text, return_tensors="pt")
+        return inputs
+
+
+@pytest.mark.parametrize(
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
+def test_codegen_generate(record_property, mode, op_by_op):
+    model_name = "codegen"
+    cc = CompilerConfig()
+    if op_by_op:
+        cc.compile_depth = CompileDepth.EXECUTE_OP_BY_OP
+        if op_by_op == OpByOpBackend.STABLEHLO:
+            cc.op_by_op_backend = OpByOpBackend.STABLEHLO
+
+    tester = ThisTester(
+        model_name,
+        mode,
+        compiler_config=cc,
+        record_property_handle=record_property,
+        is_token_output=True,
+        run_generate=True,  # run model.generate(**inputs)
+        assert_atol=False,
+    )
+
+    results = tester.test_model(
+        assert_eval_token_mismatch=False
+    )  # don't validate token output
+
+    if mode == "eval":
+        print(tester.tokenizer.decode(results[0]))
+
+    tester.finalize()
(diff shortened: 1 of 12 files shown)


gengelageTT:
Tests are passing in nightly here: https://github.com/tenstorrent/tt-torch/actions/runs/15475311565

vmilosevic:
<table><tr><th><th>Tests</th><th>Passed ✅</th><th>Skipped ⚠️</th><th>Failed</th></tr><tr><td>TT-Torch Tests</td><td>764 ran</td><td>749 passed</td><td>15 skipped</td><td>0 failed</td></tr></table>

tests/models/seamless_m4t/test_seamless_m4t.py
```diff
@@ -11,15 +11,36 @@
...
+from transformers.integrations.deepspeed import is_deepspeed_zero3_enabled
+from transformers.integrations.fsdp import is_fsdp_managed_module
+from transformers.modeling_outputs import BaseModelOutput
 import torchaudio
+import types
+
+original_torch_rand = torch.rand
+
+
+def fake_rand(*args, **kwargs):
```
jameszianxuTT, line 24: There's a 'proper' way to do [monkeypatching](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-functions) with pytests. I don't think it's really necessary in this case, but might be good to know.

tests/models/codegen/test_codegen_generate.py
```diff
@@ -0,0 +1,62 @@
...
+    "mode",
+    ["eval"],
+)
+@pytest.mark.parametrize(
+    "op_by_op",
+    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
+    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
+)
+def test_codegen_generate(record_property, mode, op_by_op):
+    model_name = "codegen"
```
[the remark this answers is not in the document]
jameszianxuTT: I think if you give it model_name `codegen_generate` that would help avoid clobbering it in reporting.


---
It occurs to me that the goal of this change:
>PCC reflects accuracy of logits and not tokens

Might have been better served by running a generative loop and finding a setting in Transformers to output raw logits at each step and not token ids, so instead of comparing outputs like

1. (Batch x TokensToGenerate x int64)            (original shape - output token IDs) 
2. (Batch x VocabSize x fp32)                           (this change - logits for one step)

We might've dumped and compared logits from the generation loop:

3. (Batch x TokensToGenerate x VocabSize x fp32)

vmilosevic:
<table><tr><th><th>Tests</th><th>Passed ✅</th><th>Skipped ⚠️</th><th>Failed</th></tr><tr><td>TT-Torch Tests</td><td>764 ran</td><td>749 passed</td><td>15 skipped</td><td>0 failed</td></tr></table>

tests/models/seamless_m4t/test_seamless_m4t.py
gengelageTT: The forward pass of SeamlessM4TConformerEncoder contains this definition for dropout_probability and skip_the_layer:


            dropout_probability = torch.rand([])

            skip_the_layer = (
                True if self.training and (dropout_probability < self.config.speech_encoder_layerdrop) else False
            )
            
There is no way to avoid this call of `torch.rand()` when `run_generate=False` from what I can tell, but since `self.training=False` it never gets used and so setting it to be 0 or 1 doesn't impact the model as long as `self.training=False`. Since `torch.rand()` isn't supported in MLIR, this change allows the model to compile.

This model does take up much more memory when `run_generate=False` and I had to significantly reduce the number of hidden layers. We could avoid both these issues if it's possible to do what James suggested in a comment above:

> It occurs to me that the goal of this change:
> "PCC reflects accuracy of logits and not tokens"
> Might have been better served by running a generative loop and finding a setting in Transformers to output raw logits at each step and not token ids

gengelageTT: Compiling only the `model.text_decoder`, similar to the stable diffusion transformer test, fixed the memory error, the PCC error and the `torch.rand()` issue. I've pushed these changes.

AleksKnezevic (approved):

vmilosevic:

Issue #926, opened 2025-06-11
Data Anomaly Tracker


jameszianxuTT:
I will be doing some bookkeeping on nightly result data product. There are a number of anomalies in reporting from the following sources:
 
- Errors in pytest script source
- Errors in yamls 
- Errors in CI pipeline / data collection infra
- Errors in SQL queries
- Errors in database design

This issue will list these anomalies.

jameszianxuTT:
1. XGLM shows up in our full model execution pytests but is actually recorded as a compiling model. This is due to a [source-level](https://github.com/tenstorrent/tt-torch/blob/3bf683b74c3e1e85e28ad33124e6dc0f9aa590d1/tests/models/xglm/test_xglm.py#L53) override in the pytest.

PR #932, opened 2025-06-12
Misc small test/model/yml changes (ghostnetv2, xglm, llama7b, falcon3-1b, n300 tests)

kmabeeTT:
### Ticket
None

### What's changed
- Remove test_xglm.py ancient hardcoded CompileDepth.TTNN_IR and put back to op-by-op nightly since an op fails at execute
 - Fix ghostnetv2_100.in1k in e2e-compile list was running op-by-op due to typo, passes (phew)
- Enable PCC checking in tiiuae/Falcon3-1B-Base, has been passing for a while
- Move tests/models/llama/test_llama_7b.py from op-by-op nightly to weekly since test_llama_7b_pipeline_parallel exists in nightly full-model-execute now and passes
- Remove cc.dump_info = True from bunch of newly added n300 tests, causes huge IR dumps, was for bringup/debug

### Checklist
- [x] Run tests locally

### DIFF
--- a/.github/workflows/run-e2e-compile-tests.yml
+++ b/.github/workflows/run-e2e-compile-tests.yml
@@ -53,7 +53,7 @@ jobs:
                   tests/models/llama/test_llama_7b.py::test_llama_7b[full-eval]
                   tests/models/falcon/test_falcon.py::test_falcon[full-eval]
                   tests/models/torchvision/test_torchvision_image_classification.py::test_torchvision_image_classification[single_device-full-eval-googlenet]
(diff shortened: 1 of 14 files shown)


jameszianxuTT (approved):

jameszianxuTT:
Addresses first two anomalies for ghostnet-v2 and xglm in #926

gengelageTT:
There are also 10+ n300 tests with `dump_debug=True` instead of `dump_info=True` which is also no longer needed and could be removed.

PR #936, opened 2025-06-13
Uplift to latest tt-forge-models to get get_file(url) rd-only cache fix (take two)

kmabeeTT:
### Ticket
None

### What's changed
- Try again, accidentally pointed to wrong/invalid commit.

### Checklist
- [x] Tested locally


jameszianxuTT:
I think the timing of this change broke the build for uplifts, which was rebased on main to a now nonexistent commit in tt_forge_models.

```
-    set(TT_FORGE_MODELS_VERSION "8d018cf5e8fac3aa3850029d34d1c75ab1c260af")
+    set(TT_FORGE_MODELS_VERSION "8222c58102587c4fd9fbf24ae521e8208203100e")
```

Perhaps a force push took place in tt_forge_models to overwrite the head commit hash, while a pointer to that head commit hash was still live in tt-torch (and possibly other repos). If that was the case, I recommend not force pushing, or else we will end up with these invalid commit references in other repos.

kmabeeTT:
> Ah I see! That makes sense. That must have been pretty unlucky timing then that uplifts triggered right between the time your branch was deleted and your followup PR fixed it 😵‍💫

Yeah, crap I would have totally relaunched uplifts if I had of realized or thought about it, darn.

PR #855, opened 2025-05-29
Test parametrization using model metadata

ssaliceTT:
### Ticket
Link to Github Issue: [Add TTNN IR compile depth specification as pytest parameterization instead of as env variable](https://github.com/tenstorrent/tt-torch/issues/755)

***NOTE***: This PR is just a proof of concept to demonstrate the feasibility of the proposed changes.

### Problem description
The original goal was to change the method of controlling compile depth in pytest as it's currently done via an environment variable `TT_TORCH_COMPILE_DEPTH`, which made CI reflection tasks, such as identifying test failures, difficult. This external dependency complicated efforts to create an automatic failure bisector because the pytest test name did not fully characterize the test's behavior.

### What's changed
Changes made to solve the issue at hand, that also form an overhaul of our testing philosophy for future purposes, are addressed below:
1. **Added `ModelMetadata` Class**:
   - Encapsulates model-specific metadata such as `model_name`, `expected_compile_depth`, `expected_op_by_op_backend`, `model_group`, and `assert_pcc`.
   - Defined directly in source test files for each model, making it easier to see the current working state of the model without digging through CI logs.

2. **Improved Parametrization for tests**:
   - Test functions now use `ModelMetadata` for parametrization, reducing redundancy and improving readability.
   - Metadata overrides `execute_mode` parametrization when needed, ensuring tests only run up to the point known to work.
   - The override ensures all tests pass successfully by limiting runs to configurations known to work.

3. **Enhanced Modularity**:
   - Test files now directly plug metadata into the tester, reducing branching logic within test files.
   - Example: In `test_EfficientNet.py`, the `model_group` parameter is passed directly from `ModelMetadata` to `tester` instead of using conditional branches within the test file to set this attribute like before.

4. **Scalability**:
   - The `ModelMetadata` class can be extended, allowing additional attributes to be added as needed for future testing requirements.

5. **CLI Filters for Pytest**:
   - Added CLI options (`--compile-depth` and `--op-by-op-backend`) in `conftest.py` to filter tests dynamically.
   - Simplifies CI workflows by eliminating manual updates and removing the need for hardcoding tests into CI.
   - More filters can be added as metadata for models increase.
   - to demo this run ```bash pytest tests/models/EfficientNet/test_EfficientNet.py --collect-only --compile-depth=STABLEHLO --op-by-op-backend=TORCH```

### Next Steps
If this proof of concept is approved, I can go ahead and make changes to all the test files to comply with these changes. Our CI will also have to be updated as a part of this.

Issue #755, opened 2025-05-12
Add TTNN IR compile depth specification as pytest parameterization instead of as env variable


jameszianxuTT:
### Is your feature request related to a problem? Please describe

Use of an environment variable to control the behavior of a pytest makes CI reflection tasks harder (eg. figuring out which tests failed is complicated, because things external to the test affect it). 

@nsumrakTT Is attempting to create an automatic failure bisector but this is made complicated by the fact that the [pytest full test name does not fully characterise the test.](https://github.com/tenstorrent/tt-torch/pull/744#discussion_r2081840415) 

https://github.com/tenstorrent/tt-torch/blob/fef23ce5c15e7b810e57b5fc558a278a92c428b7/.github/workflows/run-e2e-compile-tests.yml#L131

 

### Describe the solution you'd like

Add a [parameterization](https://docs.pytest.org/en/stable/example/parametrize.html) like op_by_op_shlo/op_by_op_torch/eval for the `ttnn_ir` compile depth, and remove the TT_TORCH_COMPILE_DEPTH env variable from the pytest.

These `mark_parameterize` sections should probably be edited, potentially changed to represent  "compile_depth"s. 

```
@pytest.mark.parametrize(
    "op_by_op",
    [OpByOpBackend.STABLEHLO, OpByOpBackend.TORCH, None],
    ids=["op_by_op_stablehlo", "op_by_op_torch", "full"],
)
```

The goal is that we can specify both the `cc.compile_depth` and `cc.op_by_op_backend` using a single parameterization with values for "full" (maybe rename to "execute" would be more appropriate), op by op torch, op by op stablehlo, and compile to ttnn IR.

PR #944, opened 2025-06-16
Compile e2e tests refactored

ssaliceTT:
### Ticket
Link to Github Issue: [Add TTNN IR compile depth specification as pytest parameterization instead of as env variable](https://github.com/tenstorrent/tt-torch/issues/755)

### Problem description
The original goal was to change the method of controlling compile depth in pytest as it's currently done via an environment variable `TT_TORCH_COMPILE_DEPTH`, which made CI reflection tasks, such as identifying test failures, difficult. This external dependency complicated efforts to create an automatic failure bisector because the pytest test name did not fully characterize the test's behavior.

### What's changed
The changes proposed in https://github.com/tenstorrent/tt-torch/pull/855 has been applied to the compile e2e test files. A short summary of the changes are as follows:
- Added `ModelMetadata` Class
- Improved Parametrization for tests with overrides from ModelMetadata
- Better modularity as test files now directly plug metadata into the tester, reducing branching logic within test files.
- Improved scalability since `ModelMetadata` class can be extended, allowing additional attributes to be added as needed for future testing requirements.
- New CLI Filters for Pytest

### Next Steps
If approved, I can go ahead and make changes to all the test files to comply with these changes and our CI will also have to be updated as a part of this. In this pull request I haven't made changes to the CI to comply with the changes in the test files.


jameszianxuTT:
1. Please rebase and fix the conflicts so CI runs on this change
> In this pull request I haven't made changes to the CI to comply with the changes in the test files.
2. Please also add these changes in CI.
I think we can't switch over to doing full test autodetection yet using the new CLI options you've added in conftest.py since some more infra work needs to be done to support that, but this change is still backwards compatible with the way we currently specify tests to run in each mode via testgroups + a list of tests with their full parameterizations detailed. So, the changes needed to the YML files should be simple, i.e. just removing the TTNN_IR override in the e2e compile yaml and checking that it still works (and does run in compile).

tests/models/torchvision/test_torchvision_image_classification.py
```diff
@@ -5,22 +5,30 @@
...
 
 
-class ThisTester(ModelTester):
-    # pass model_info instead of model_name
-    def __init__(self, model_info, mode, *args, **kwargs):
-        # model name in model_info[0]
-        self.model_info = model_info
-        super().__init__(model_info[0], mode, *args, **kwargs)
+# TorchVision (TV) model metadata
+# Special Subclass of ModelMetada made for this test to include weight_names attr
```
jameszianxuTT, line 19: Not sure from a sw design perspective what would be better here - should we make the precedent to subclass ModelMetadata when we want to add a field that only pertains to a few tests, or stuff it all into a monolithic class? 

Some considerations against subclassing:
- We may want to serialize this class in the future to push it into the database, which would be less complicated if we don't subclass
- More decisions need to be made on whether a field is "specialized" enough, like "weights_name", to be subclasssed. Though, we could take the standard that ModelMetadata should only contain fields that pertain to all/most model tests, and to create subclasses if extra fields are needed.

vmilosevic:
<table><tr><th><th>Tests</th><th>Passed ❌️</th><th>Skipped</th><th>Failed</th></tr><tr><td>TT-Torch Tests</td><td>0 ran</td><td>0 passed</td><td>0 skipped</td><td>0 failed</td></tr></table>

vmilosevic:
<table><tr><th><th>Tests</th><th>Passed ☑️</th><th>Skipped ⚠️</th><th>Failed ❌️</th></tr><tr><td>TT-Torch Tests</td><td>766 ran</td><td>750 passed</td><td>15 skipped</td><td>1 failed</td></tr></table>

<table><tr><th>Test</th><th>Result</th></tr><tr><td colspan="2"><strong>TT-Torch Tests</strong></td></tr><tr><td colspan="2"><em>pytest</em></td></tr><tr><td>test_benchmark.test_crashsafe_multitest_assert</td><td>❌ failure</td></tr></table>

tests/models/EfficientNet/test_EfficientNet.py
```diff
@@ -33,60 +33,59 @@ def _load_inputs(self):
         return tfms(img).unsqueeze(0)
 
 
+# Metadata for EfficientNet models
+EFFICIENTNET_VARIANTS = [
+    ModelMetadata(model_name="efficientnet-b0", model_group="red"),
```
AleksKnezevic, line 38: @kmabeeTT does this work with the infra we're setting up with tt-forge-models?

kmabeeTT: First thought is I kind of wish this PR was more targeted to solving just a single problem at once (like not relying on env-vars) rather than encompasing larger refactors.  

James makes good point, I agree - about using `ModelInfo` from tt-forge-models here, somehow. Yes there is some overlap here with my pending PR in tt-forge-models repo that has more buy-in from other FE leads (we had meeting on it this morning, folks seemed mostly aligned), so we should be careful here. It will have variant names (b0-b7 here when the model is ported there), and also the model_group tag, queryable with `get_model_info()` API (see code from my PR here: https://github.com/tenstorrent/tt-forge-models/blob/kmabee/model_variants_wip/albert/masked_lm/pytorch/loader.py#L58)

Also with those changes, variants of a model (like efficientnet-b0 through b7 here) are queryable via `available_variants = ModelLoader.query_available_variants()` so they do not even neccesarily need to be listed out explicitly (though they still could be), example usage:

image

Trying to think about what makes sense here timing wise... the changes are not yet merged in tt-forge-models and it will take some time (weeks) for them all to appear.  

1. One option is to hold off on merging this PR and re-evalate once tt-forge-models goes in (slight preference), 
2. another option is to just split out the fix from this PR for the original problem we want to solve (compile depth env-var usage) if that makes sense.
3. Another option is to merge this, but then update it all again after tt-forge-models is updated with all models, variants (less preferred)

vmilosevic:
<table><tr><th><th>Tests</th><th>Passed ❌️</th><th>Skipped</th><th>Failed</th></tr><tr><td>TT-Torch Tests</td><td>0 ran</td><td>0 passed</td><td>0 skipped</td><td>0 failed</td></tr></table>

vmilosevic:

vmilosevic:

tests/models/EfficientNet/test_EfficientNet.py
kmabeeTT: Thanks James - agreed about not wanting FE-specific stuff like PCC in tt-forge-models, and I see what you mean about joining FE-common and FE-specific attributes.

I'm finding it hard to fully appreciate all the changes here, and kind of preferred the source level branches (could be put in a function for better encapsulation) because it allowed us to assume default behavior for majority of cases for free, and override fields only when needed, and not need to introduce `ModelMetadata` and duplicately enumerate all variants to run when many use defaults.

For current changes, would it work here with several fields inside `ModelMetadata` being dependent on parallelization mode or architecture?  

- Take for instance Falcon7B limits `compile_depth` to avoid execute because it doesn't fit on device during execute, but it actually would in `pipeline_parallel` mode, but now we've hardcoded it in `ModelInfo`.  
- And some models pass PCC for wormhole, but fail for blackhole, how to control that if the settings are part of `ModelInfo`?

kmabeeTT: How about something like just specifying `ModelMetaData` map by `variant_name` for models that diverge from defaults rather than needing to duplicately enumerate all models in a list of `ModelMetaData`?   First, pytest variants are auto-discovered from available variants like my tt-torch PR wants to soon start to do:

```
available_variants = ModelLoader.query_available_variants()

@pytest.mark.parametrize(
    "variant_info",
    available_variants.items(),
    ids=list(available_variants.keys()),
)
```

Then, something like:

```
// Note: model_info, like before is a ModelMetaData instance
model_info = get_fe_model_info(variant_name)
tester = ThisTester(model_info.model_name, model_info.assert_pcc, ... etc)
```

Where `get_fe_model_info(variant_name)` is wrapper that contains the list of `ModelMetaData` (but now a map), or we leave it as a global variable as-is, and also gives us single entry point to call `get_model_info()` from tt-forge-models to supply the FE-common fields (model_group, model_name, etc) to `ModelMetaData` directly, (or `ModelInfo` object itself if we want), and then it will also supply the FE-specific arguments if your map contains an override for this variant, otherwise return a default-constructed `ModelMetaData` object.  

In my head this way many tests barely need to change, and the stuff we do need to change (add) is just for divergences from default behavior, and programatically generated queries etc would still work.

Issue #943, opened 2025-06-16
op_by_op flow can take huge amount of memory for goldens (30GB+) causing OOM in CI (OFT, SD3.5)


kmabeeTT:
**Problem:**

We have been seeing some models like SD3.5 Large transformer ([link](https://github.com/tenstorrent/tt-torch/actions/runs/15648259967/job/44089556398)) and OFT([link](https://github.com/tenstorrent/tt-torch/actions/runs/15659352859/job/44115019140)) get killed in op-by-op flow for some time, preventing nightly to pass, occasionally passes on rerun.   Spent some time adding initial memory profiling debug prints and see that op-by-op goldens is mostly responsible (30+GB), specifically this line where golden is added to dict and in the case of OFT model, never freed (?):

```
node_to_tensor[node] = golden
```

Example, in OFT, debug prints show 172 MB golden tensor (this type of thing happens hundreds of times) being added to increase the size of the dict to grow to 33GB over the course of the model:

```
Processing global_op_idx: 2085 (2085/2350): aten.add.Tensor | time: 0.0000 s |
[Before golden 2085] Memory Usage: 34935.96 MB
Golden     global_op_idx: 2085 (2085/2350): aten.add.Tensor | time: 0.0254 s |
[Before run 2085] Memory Usage: 35108.70 MB
Size of golden in MB: 172.8193359375
Total size of node_to_tensor dictionary in MB: 33975.702998161316

Processing global_op_idx: 2086 (2086/2350): aten.ge.Scalar | time: 0.0000 s |
[Before golden 2086] Memory Usage: 34763.26 MB
Golden     global_op_idx: 2086 (2086/2350): aten.ge.Scalar | time: 0.0106 s |
[Before run 2086] Memory Usage: 34806.31 MB
Size of golden in MB: 43.204833984375
Total size of node_to_tensor dictionary in MB: 33802.883662223816

<snip>

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================= 1 passed, 39 warnings in 36.31s ========================
        Command being timed: "pytest -svv tests/models/oft/test_oft.py::test_oft[op_by_op_torch-eval]"
        User time (seconds): 140.19
        System time (seconds): 81.42
        Percent of CPU this job got: 579%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:38.23
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
[...]

PR #948, opened 2025-06-17
Disable consteval_parameters to reduce host RAM usage in SD3.5 Large Transformer test 

kmabeeTT:
### Ticket
#943 

### Problem description
- SD3.5 Large Transformer test still uses lots of memory even after intermediates fix and CI machines only have 32GB or 48GB

### What's changed
- Remove setting of consteval_parameters with same comment as other tests have, it's enough to solve problem
- Reduces memory from (65GB->32GB) when combined with recent merged intermediates fix

### Checklist

### DIFF
--- a/tests/models/stable_diffusion/test_stable_diffusion_transformer.py
+++ b/tests/models/stable_diffusion/test_stable_diffusion_transformer.py
@@ -98,7 +98,7 @@ def test_stable_diffusion_transformer(record_property, model_info, mode, op_by_o
 
     cc = CompilerConfig()
     cc.enable_consteval = True
-    cc.consteval_parameters = True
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Bunch of other tests with same comment/workaround:

```
falcon/test_falcon3.py:    # consteval_parameters is disabled because it results in a memory related crash
flux/test_flux.py:    # consteval_parameters is disabled because it results in a memory related crash
phi/test_phi_3p5_moe.py:    # consteval_parameters is disabled because it results in a memory related crash
stable_diffusion/test_stable_diffusion_3_5.py:    # consteval_parameters is disabled because it results in a memory related crash
stable_diffusion/test_stable_diffusion_transformer.py:    # consteval_parameters is disabled because it results in a memory related crash
```

Issue #614, opened 2025-04-14
Generative model


AleksKnezevic:
Bring up generative model of llama3b with past cache. Let's discuss how to do this.

PR #951, opened 2025-06-17
Add path to disable backend consteval pass from compiler config

jameszianxuTT:
### Ticket
None

### Problem description
The default for the consteval option in compileTTIRToTTNN changed a month ago to default to true. This caused our consteval environment variable to be unable to suppress the backend consteval pass. We also didn't have a programmatic path to set the TTIRToTTNN consteval pass option without an env variable.

Required as a stopgap for #614 

### What's changed
Plumb the compiler_config.enable_consteval option through the backend to allow it to be disabled properly, remove the now invalid use of the env variable.

### Checklist

### DIFF
--- a/tt_torch/csrc/bindings.cpp
+++ b/tt_torch/csrc/bindings.cpp
@@ -162,9 +162,11 @@ std::string compile_stable_hlo_to_ttir(std::string_view code) {
 std::tuple<py::bytes, std::string>
 compile_ttir_to_bytestream(std::string_view code,
                            std::string_view sys_desc_path,
-                           size_t len_activations, size_t len_graph_constants) {
-  auto [binary_ptr, ttnn] = tt::torch::compileTTIRToTTNN(
-      code, sys_desc_path, len_activations, len_graph_constants);
+                           size_t len_activations, size_t len_graph_constants,
+                           bool enable_consteval = true) {
+  auto [binary_ptr, ttnn] =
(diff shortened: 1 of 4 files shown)


jameszianxuTT:
I would like to turn this off for now because it has complex interactions with the llama static cache, which:
- is traced as a graph constant and constevaled at a tt-torch level and at this pass level
  - because static cache is marked as a buffer at transformers source level
     - because we just stuff buffers into our graph constants when processing the fx graph at the start of compilation. We don't actually represent buffers "properly" 


For reference: In this block, we set the `constant_inputs` to the graph:
https://github.com/tenstorrent/tt-torch/blob/91873b5720d44729b0b1152a4877fb7d07bc058e/tt_torch/dynamo/passes.py#L485-L499

jameszianxuTT:
Ref to backend consteval default change: https://github.com/tenstorrent/tt-mlir/commit/534e625f969b415db2f3723a76bbfb15865beb48

jameszianxuTT:
Closing because we want to minimize model-specific compiler config overrides, and this path to disable consteval for generative models only will mask consteval-related issues for generative models.

vmilosevic:

jameszianxuTT:
Reopening because we do want to disable consteval for llama generative tests, it doesn't fully work and we have no path to sharing constant weights between the prefill and decode graphs from tt-torch.

Issue #960, opened 2025-06-18
resnet50 mlir file missing from model mlir dumps


ddilbazTT:

ddilbazTT:
@sgholamiTT I fixed this issue :) You should be able to find Resnet50 in model-mlir-execute-push https://github.com/tenstorrent/tt-torch/actions/runs/15815100384

sgholamiTT:
> [@sgholamiTT](https://github.com/sgholamiTT) I fixed this issue :) You should be able to find Resnet50 in model-mlir-execute-push https://github.com/tenstorrent/tt-torch/actions/runs/15815100384

Yes, found it. This resolution is very helpful for me as I'll be looking into other models too.
Thanks.

PR #970, opened 2025-06-19
Fix mlir dumps #960

ddilbazTT:
### Ticket
#960 

### Problem description
MLIR dumps were not being uploaded properly due to a combination of problems:
- `model_name` was not initialized in `compiler_config`
- environment variables were read in `compiler_config` initialization, which included cleaning up of old mlir files for the model
- `sanitize_filename` would return empty string during `compiler_config` initialization because `model_name` is not set
- so during cleanup of old mlir files, all files under `model_mlir` would be deleted because they all start with empty string

### What's changed
Added `set_model_name` method that also cleans up old mlir dumps if the associated environment variable is set

### Checklist


jameszianxuTT (approved):

jameszianxuTT:
Had a small chat offline, I think if this is a CI only tool it might be better to write out the logs entirely and have some script to dissect them into individual MLIR files + account for graph breaks. This would make the flow more annoying to use locally though.

Otherwise, splicing the mlir dump file logic inline into our ModelTester et al. requires a lot of complex graph-break handling logic to handle file lifetime and cleanup. But I don't want to block this change, maybe we can discuss more further.

vmilosevic:

AleksKnezevic (approved):

Issue #779, opened 2025-05-15
Extend pipeline parallel pass to topologically sort the graph.


AleksKnezevic:
The transformers library infer_auto_device_map sorts modules on separate device based on where they're defined in the model, not topological order. Thus requiring manual overrides to some modules to allow the graph to execute fully on each device. We should add a pass that can topologically sort the modules and set devices, or at the very least tell the user which module placements need to be overriden.

AleksKnezevic:
Let's do a design review before you start on this.

PR #969, opened 2025-06-19
Register backend as `tt` #963

ddilbazTT:
### Ticket
#963 

### Problem description
Register backend as `tt` so users can simply do:
```
import tt_torch
...
tt_model = torch.compile(..., backend="tt", ...)
```

### Checklist
- [X] Registered backend
- [X] Imported backend in tt_torch
- [X] Updated demos/ tests which use `torch.compile(..., backend=backend, ...)`

PR #971, opened 2025-06-19
Topological Sort of device_map in Pipeline Parallel 

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/779

### Problem description
The transformers library infer_auto_device_map sorts modules on separate devices based on where they're defined in the model, not topological order. Thus requiring manual overrides to some modules to allow the graph to execute fully on each device. The pipeline parallel pass should automatically handle this.

### What's changed
The approach to ensure the device_map assigned devices in topological order was to continue to call infer_auto_device_map and then check for any instances when the input to a node (determined by the gm) comes from a later device. If such a case is encountered, then the node providing the input should be moved to the lowest device index that satisfies all its consumers. The inputs to the moved node are recursively checked to ensure moving the node doesn't introduce new issues.

While this approach works for the issue seen in **tests/models/llama/test_llama_7b_pipeline_parallel.py**
it could introduce the possibility of the first device(s) becoming overloaded and running out of memory if the initial device_map is very disorganized. 
This can be seen in **tests/torch/test_basic_multichip.py::test_pipeline_parallel_topological_sort** as all the modules are moved to the first device despite there being alternate device_maps that could work and would be more balanced. 
A possible way to reduce the chance of this happening is mentioned in `sort_device_map()`. I chose not to implement this as it would add complexity that I think is unnecessary unless we encounter this issue. Also, it doesn't guarantee the devices will be balanced.

This is what changed specifically:
- **dynamo/passes.py**
  - `node_to_device()` takes in `key` (default `False`) and `verbose` (default `True`). This allows it to return the key from device_map along with the device if `key=True` and disables the printed warning if `verbose=False`
  - NEW `sort_device_map()` finds nodes that are out of topological order in the device_map and calls `move_device_map_key()` to fix the issue
  - NEW `move_device_map_key()` changes the device_map to assign a new device to the module that is causing an issue. It also moves any inputs to that modules nodes that need to be moved as well
  - `split_onto_devices()` calls `sort_device_map()` before splitting the graph onto the devices
- **tests/torch/test_basic_multichip.py**
  - NEW `test_pipeline_parallel_topological_sort()` tests that incorrect device maps will be fixed (also checks the recursive functionality of `move_device_map_key()`)
- **tests/models/llama/test_llama_7b_pipeline_parallel.py**
  - removed manual change of device_map

### Checklist

### DIFF
--- a/tests/torch/test_basic_multichip.py
+++ b/tests/torch/test_basic_multichip.py
@@ -42,3 +42,39 @@ def test_pipeline_parallel():
     golden = host_model(x, y)
     verify_against_golden((golden,), (calculated,), True, True, required_atol=0.1)
     DeviceManager.release_parent_device(parent_device, True)
+
+
+def test_pipeline_parallel_topological_sort():
+    class Basic(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.l1 = nn.Linear(32, 32)
+            self.l2 = nn.Linear(64, 32)
+            self.l3 = nn.Linear(32, 32)
+
+        def forward(self, x, y):
+            x = self.l1(x)
+            y = self.l2(y)
+            z = self.l3(y)
+            return x + y + z
+
+    options = BackendOptions()
+    cc = CompilerConfig()
+    options.compiler_config = cc
+    cc.enable_consteval = True
+    cc.consteval_parameters = True
+    cc.device_map = {"l1": 0, "l2": 1, "l3": 0}
+    parent_device = DeviceManager.create_parent_mesh_device([1, 2])
+    device1 = DeviceManager.create_sub_mesh_device(parent_device, (0, 0))
+    device2 = DeviceManager.create_sub_mesh_device(parent_device, (0, 1))
+    options.devices = [device1, device2]
+
+    host_model = Basic()
+
+    model = torch.compile(host_model, backend=backend, options=options)
+    x = torch.rand(32, 32)
+    y = torch.rand(32, 64)
+    calculated = model(x, y)
+    golden = host_model(x, y)
+    verify_against_golden((golden,), (calculated,), True, True, required_atol=0.1)
+    DeviceManager.release_parent_device(parent_device, True)
--- a/tests/torch/test_basic_multichip.py
+++ b/tests/torch/test_basic_multichip.py
@@ -82,7 +82,7 @@ def test_pipeline_parallel_topological_sort():
 
     host_model = Basic()
 
-    model = torch.compile(host_model, backend=backend, options=options)
+    model = torch.compile(host_model, backend="tt", options=options)
     x = torch.rand(32, 32)
     y = torch.rand(32, 64)
     calculated = model(x, y)
--- a/tests/torch/test_basic_multichip.py
+++ b/tests/torch/test_basic_multichip.py
@@ -44,17 +44,16 @@ def test_pipeline_parallel():
     DeviceManager.release_parent_device(parent_device, True)
 
 
-class L3(nn.Module):
-    def __init__(self):
-        super().__init__()
-        self.linear = nn.Linear(32, 32)
-
-    def forward(self, a, b):
-        # Example: add the two inputs, then apply linear
-        return self.linear(a + b)
-
-
 def test_pipeline_parallel_topological_sort():
+    class L3(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.linear = nn.Linear(32, 32)
+
+        def forward(self, a, b):
+            # Example: add the two inputs, then apply linear
+            return self.linear(a + b)
+
     class Basic(nn.Module):
         def __init__(self):
             super().__init__()
--- a/tests/torch/test_basic_multichip.py
+++ b/tests/torch/test_basic_multichip.py
@@ -44,26 +44,38 @@ def test_pipeline_parallel():
     DeviceManager.release_parent_device(parent_device, True)
 
 
+class L3(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.linear = nn.Linear(32, 32)
+
+    def forward(self, a, b):
+        # Example: add the two inputs, then apply linear
+        return self.linear(a + b)
+
+
 def test_pipeline_parallel_topological_sort():
     class Basic(nn.Module):
         def __init__(self):
             super().__init__()
             self.l1 = nn.Linear(32, 32)
             self.l2 = nn.Linear(64, 32)
-            self.l3 = nn.Linear(32, 32)
+            self.l3 = L3()
+            self.l4 = nn.Linear(32, 32)
 
         def forward(self, x, y):
             x = self.l1(x)
             y = self.l2(y)
-            z = self.l3(y)
-            return x + y + z
+            z = self.l3(x, y)
+            a = self.l4(z)
+            return a
 
     options = BackendOptions()
     cc = CompilerConfig()
     options.compiler_config = cc
     cc.enable_consteval = True
     cc.consteval_parameters = True
-    cc.device_map = {"l1": 0, "l2": 1, "l3": 0}
+    cc.device_map = {"l1": 0, "l2": 1, "l3": 1, "l4": 0}
     parent_device = DeviceManager.create_parent_mesh_device([1, 2])
     device1 = DeviceManager.create_sub_mesh_device(parent_device, (0, 0))
     device2 = DeviceManager.create_sub_mesh_device(parent_device, (0, 1))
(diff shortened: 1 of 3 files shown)


tt_torch/dynamo/passes.py
```diff
@@ -247,9 +250,80 @@ def _generate_golden_intermediate_cache(gm, inputs, compiler_config):
...
+            ):
+                device_map = move_device_map_key(gm, input_key, node_device, device_map)
+
+    return device_map
+
+
 # The following function splits the graph onto the devices specified in the device_map
 # We create empty subgraphs for each device and then add the nodes to the appropriate subgraph
 def split_onto_devices(gm, compiler_config):
+
```
AleksKnezevic, line 322: Perhaps we can move this to below line 351, that is if there is no device map, we don't need to sort it.

tests/torch/test_basic_multichip.py
```diff
@@ -41,3 +41,50 @@ def forward(self, x, y):
...
+            z = self.l3(x, y)
+            a = self.l4(z)
+            return a
+
+    options = BackendOptions()
+    cc = CompilerConfig()
+    options.compiler_config = cc
+    cc.enable_consteval = True
+    cc.consteval_parameters = True
+    cc.device_map = {"l1": 0, "l2": 1, "l3": 1, "l4": 0}
```
AleksKnezevic, line 76: so I understand the logic, this will place everything on device 0, right?

AleksKnezevic (approved):
This looks great, left a couple of nits, but otherwise it's good to go. 

I added a couple more people to review as well. Having more eyes on code is always better 😄

ddilbazTT:
@AleksKnezevic Right now we do not test `tests/torch/test_basic_multichip.py` on PR, but given we are going to be making a lot of multichip support changes, should we also add that in the on PR run? For example on this PR, we cannot confirm the changes would not break multichip support.

tests/torch/test_basic_multichip.py
```diff
@@ -41,3 +41,50 @@ def forward(self, x, y):
...
+    parent_device = DeviceManager.create_parent_mesh_device([1, 2])
+    device1 = DeviceManager.create_sub_mesh_device(parent_device, (0, 0))
+    device2 = DeviceManager.create_sub_mesh_device(parent_device, (0, 1))
+    options.devices = [device1, device2]
+
+    host_model = Basic()
+
+    model = torch.compile(host_model, backend=backend, options=options)
```
ddilbazTT, line 84: Could you do 

```
import tt_torch
...
model = torch.compile(host_model, backend="tt", options=options)
``` 
based on this PR? https://github.com/tenstorrent/tt-torch/pull/969 Thanks!

tt_torch/dynamo/passes.py
```diff
@@ -143,10 +145,11 @@ def node_to_device(node, device_map):
     for i in range(1, len(parsed_vals) + 1):
         layer = ".".join(parsed_vals[:i])
         if layer in device_map:
-            return device_map[layer]
+            return (device_map[layer], layer) if key else device_map[layer]
 
-    print(f"Warning: No device found for node {node}")
-    return None
+    if verbose:
+        print(f"Warning: No device found for node {node}")
```
ddilbazTT, line 151: nit: I don't think we use `if verbose` anywhere except for unused code. I think we use (albeit, not really well) debug flags in the compiler config. You can take a look into `dump_debug` `dump_info` `tt_torch/tools/utils.py`

tt_torch/dynamo/passes.py
```diff
@@ -247,9 +250,80 @@ def _generate_golden_intermediate_cache(gm, inputs, compiler_config):
...
+
+    # Find all nodes that will be moved and check their inputs
+    nodes_with_key = []
+    for graph_node in gm.graph.nodes:
+        _, key = node_to_device(graph_node, device_map, key=True, verbose=False)
+        if key == node_key:
+            nodes_with_key.append(graph_node)
+
+    for curr_node in nodes_with_key:
+        # Recursively move all inputs of curr_node to target_device if they are on a later device
```
ddilbazTT, line 275: nit: I might be following incorrectly, but I think the comment here and what's happening don't match. I assume that 

> Recursively move all inputs of curr_node to target_device if they are on a later device

means `target_device > input_device`

tt_torch/dynamo/passes.py
```diff
@@ -247,9 +250,80 @@ def _generate_golden_intermediate_cache(gm, inputs, compiler_config):
...
+    device_map = compiler_config.device_map.copy()
+    for node in gm.graph.nodes:
+        node_device = node_to_device(node, device_map, verbose=False)
+        for input_node in node.all_input_nodes:
+            input_device, input_key = node_to_device(
+                input_node, device_map, key=True, verbose=False
+            )
+            if (
+                node_device is not None
+                and input_device is not None
```
ddilbazTT, line 311: you can simply do `if node_device and input_device and input_device > node_device` assuming `0`, empty string etc is also not expected to check this if statement.

ddilbazTT (approved):

LPanosTT (approved):

jameszianxuTT (approved):

tt_torch/dynamo/passes.py
```diff
@@ -247,9 +250,80 @@ def _generate_golden_intermediate_cache(gm, inputs, compiler_config):
...
+def move_device_map_key(gm, node_key, target_device, device_map, call_stack=None):
+    # Move node_key in device_map to target_device and recursively move any of its input nodes
+    # that are on a device with a higher index than target_device.
+    # Returns the updated device_map.
+
+    if call_stack is None:
+        call_stack = set()
+
+    # Prevent infinite recursion when a node_key has internal dependencies
+    if node_key in call_stack:
```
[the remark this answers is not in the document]
gengelageTT: The node_key is the module that is assigned a device in the device_map. I'm checking the inputs of each node in the module to make sure they're on the target_device or earlier. However, some of the nodes likely depend on nodes within the same module (not yet on the target device), and since we already plan to move that module and are currently in the process of doing so we can skip that node - which is what this line checks. Once the module has been moved we remove it from the call stack in case it happens to need to move again.

Looking at it again I think it might be simpler to check that `input_key != node_key` before calling `move_device_map_key` in line 281.

tt_torch/dynamo/passes.py
gengelageTT: If `target_device < input_device` we need to move the input to `target_device` before we can move the current node. 'l'll reword this comment so "later device" is more clear

tt_torch/dynamo/passes.py
gengelageTT: This function was initially created with the warning always printing regardless of the debug settings. Since I call this function repeatedly, I just wanted to disable print when I call it to avoided unnecessary repetition. Even if we want the print statements to only happen if debug flags are on, I would still need the `verbose` input to avoid the repeats. I can rename it if that would reduce confusion. Do you have any other ideas for how to approach this?

tests/torch/test_basic_multichip.py
gengelageTT: yes. I could add some logic to find the last node at the start of `sort_device_map()` and move it to the last device (if it isn't already on it) if we want to ensure there will always be at least 1 node on both devices.

tt_torch/dynamo/passes.py
```diff
@@ -247,9 +250,80 @@ def _generate_golden_intermediate_cache(gm, inputs, compiler_config):
...
+    # Current implementation moved inputs to the device of the node that consumes them if
+    # the input is on a later device.
+    # If this causes the first device to become full, an alternative approach could be to
+    # consider moving the consuming node to the device of the input.
+    # - Calculate the cost to move the input vs consumer and choose the cheaper option.
+
+    device_map = compiler_config.device_map.copy()
+    for node in gm.graph.nodes:
+        node_device = node_to_device(node, device_map, verbose=False)
+        for input_node in node.all_input_nodes:
```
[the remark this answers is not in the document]
gengelageTT: It returns all input nodes of a `torch.fx.Node`

> Return all Nodes that are inputs to this Node. This is equivalent to iterating over args and kwargs and only collecting the values that are Nodes.

https://docs.pytorch.org/docs/stable/fx.html#torch.fx.Node

AleksKnezevic:
> @AleksKnezevic Right now we do not test `tests/torch/test_basic_multichip.py` on PR, but given we are going to be making a lot of multichip support changes, should we also add that in the on PR run? For example on this PR, we cannot confirm the changes would not break multichip support.

Good idea, @gengelageTT, can you please add that to on PR.

gengelageTT:
> @ddilbazTT I think this is already in the onPR tests: https://github.com/tenstorrent/tt-torch/actions/runs/15767030851/job/44445634908?pr=971#step:9:311

@AleksKnezevic ^

PR #981, opened 2025-06-23
Move print statement in `dump_module`

ddilbazTT:
### Ticket
#960 

### Problem description
Print statement would have been reached even if the user doesn't want to dump mlir modules, clobbering output

### What's changed
Moved print statement and replaced with asserts

### Checklist


tt_torch/dynamo/backend.py
```diff
@@ -102,13 +102,15 @@ def dump_module(module, name, compiler_config):
...
 
-    if not compiler_config.model_name:
-        print("Cannot dump module, no model name provided", file=sys.stderr)
-        return
-
     if compiler_config.save_mlir_override and name.lower() in (
         n.lower() for n in compiler_config.save_mlir_override
     ):
+        if not compiler_config.model_name:
+            print("Cannot dump module, no model name provided for save_mlir_override")
```
jameszianxuTT, line 109: Since we're here anyways - why not assert on both of these? Are they not both fatal conditions for dumping MLIR override logs?

Issue #922, opened 2025-06-11
Add pipeline parallel tests for falcon3 7B and 10B


AleksKnezevic:
Add N300 pipeline parallel tests for falcon 3 models that don't fit on one device. Please note, there's a limitation to current pipeline parallel support (see https://github.com/tenstorrent/tt-torch/issues/779). It may be worth addressing that first if it's a blocker.

PR #986, opened 2025-06-24
Add Falcon3 7B and 10B Pipeline Parallel Tests

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/922 

### Problem description
Add N300 pipeline parallel tests for falcon 3 models that don't fit on one device.

### What's changed
Added pipeline parallel tests for falcon3 7B and 10B. 

**Issue**: The 10B model introduced a new error where `infer_auto_device_map()` assigns modules to device `"disk"` when it estimates devices 0 and 1 will be full. This causes `split_onto_devices()` to fail since it expects only integer device assignments. Additionally, manual changes to `device_map` proved that this model can fit on 2 devices and so the prediction that device 0 and 1 will be full is not always correct.

[EDIT]: This has changed to be a much simpler solution, see comment below

**Solution**: Added logic to `split_onto_devices()` that:
1. Moves all `"disk"` modules to the last device in `device_map`
2. Warns user that this change may cause OOM error and suggests increasing the `max_memory` per device if needed

This approach is a simple solution that allows the device map to be balanced by manually increasing the max_memory. The added test shows this by using higher `max_memory` values (10GiB vs 8GiB) for the 10B model to create a more balanced device map which allows the test to pass.

**Alternate approach**: Balance `device_map` in the pass instead of manually changing `max_memory`. If this approach is preferred I can make an issue to implement it and could work on that next.

[EDIT]: A self contained path to creating a junitxml report for tests that bypass the ModelTester was also added, which extracts the common logic from ModelTester report generation and exposes it as a new static function ModelTester.GenerateCustomTestReport 

Summary:
- **tt-torch/backend/passes.py**
  - move modules assigned to `"disk"` to the last device
  - warn user that increasing max_memory might be necessary
- **tests/falcon/test_falcon3_7b_10b_pipeline_parallel.py**
  - replicates format of **tests/llama/test_llama_7b_pipeline_parallel.py**
  - Falcon3 10B `max_memory` increased to 10GiB per device
- **.github/workflows/run-full-model-execution-tests.yml**
  - added tests to on-pr high-mem quarantine since they fail on lower memory n300s (same as the llama test)
 

### Checklist

### DIFF
--- a/tests/models/falcon/test_falcon3_7b_10b_pipeline_parallel.py
+++ b/tests/models/falcon/test_falcon3_7b_10b_pipeline_parallel.py
@@ -58,13 +58,10 @@ def test_falcon_pipeline_parallel(record_property, model_name, mode, op_by_op):
     dont_split = (
         model._no_split_modules if hasattr(model, "_no_split_modules") else None
     )
-    max_memory = (
-        {0: "10GiB", 1: "10GiB"}
-        if model_name == "tiiuae/Falcon3-10B-Base"
-        else {0: "8GiB", 1: "8GiB"}
-    )
+    # The devices have 12GB of memory each, but we set it to 11GB to ensure
+    # there's enough room for activation tensors and other overhead.
     device_map = infer_auto_device_map(
-        model, max_memory=max_memory, no_split_module_classes=dont_split
+        model, max_memory={0: "11GiB", 1: "11GiB"}, no_split_module_classes=dont_split
     )
 
     options = BackendOptions()
--- a/tests/models/falcon/test_falcon3_7b_10b_pipeline_parallel.py
+++ b/tests/models/falcon/test_falcon3_7b_10b_pipeline_parallel.py
@@ -5,9 +5,9 @@ import torch
 import pytest
 
 from transformers import AutoTokenizer, AutoModelForCausalLM
-from tests.utils import ModelTester, skip_full_eval_test
+from tests.utils import ModelTester
 from tt_torch.tools.device_manager import DeviceManager
-from tt_torch.tools.utils import CompilerConfig, CompileDepth, OpByOpBackend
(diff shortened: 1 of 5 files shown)


tests/models/falcon/test_falcon3_7b_10b_pipeline_parallel.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Create submeshes that target different devices
+    device1 = DeviceManager.create_sub_mesh_device(parent_device, (0, 0))
+    device2 = DeviceManager.create_sub_mesh_device(parent_device, (0, 1))
+
+    dont_split = (
+        model._no_split_modules if hasattr(model, "_no_split_modules") else None
+    )
+    max_memory = (
+        {0: "10GiB", 1: "10GiB"}
```
AleksKnezevic, line 62: Why is it different per model? I believe this should be 12GB per chip. I made a mistake in the other model I brought up.

gengelageTT:
Switched to have `max_memory={0: "11GiB", 1: "11GiB"}`. Devices have 12GiB each but need some room saved for activation tensors and overhead. Chose 11GiB since Llama has OOM for 12GiB and Falcon3 10B sees one module on "disk" for 10GiB.

vmilosevic:

jameszianxuTT:
@gengelageTT brings up a good point - because we bypass the ModelTester here, we also won't report on these model statuses.

Test reporting is tightly coupled to ModelTester, so we either:
1. Change model tester so it can handle these tests
2. Expose a path to test reporting uncoupled to model tester

@AleksKnezevic  I think the latter is preferable/simpler, so I will expose some small APIs to perform test reporting without going through the ModelTester and attach it to this PR

vmilosevic:

jameszianxuTT (approved):

AleksKnezevic (approved):

PR #1006, opened 2025-06-27
Patch buffer cache management logic to handle graph breaks 

jameszianxuTT:
### Ticket
None

### Problem description
Buffer cache management doesn't properly account for graph breaks; where a single model execution can result in multiple calls to the executor for each subgraph, where each subgraph has a different number of constants to cache.

This patch does over-eager cache invalidation in graph break scenarios (i.e. between every graph break, if the set of graph constants change) in order to prevent overcommitment of device DRAM.

### What's changed
- Add set check of set(already cached torch tensor graph constants) vs. set(mcg.graph_constants) and perform cache invalidation based on this
- Add check before cache invalidation call that the cache we are invalidating is not empty. use of list(dict_keys) accidentally can throw a [None] to the deallocator which would result in a runtime error   

### Checklist

### DIFF
--- a/tt_torch/dynamo/executor.py
+++ b/tt_torch/dynamo/executor.py
@@ -288,18 +288,26 @@ class Executor:
 
         device_constant_cache = self.constant_cache.get(device, {})
 
-        # Check if cache is uninitialized or has invalid entries
+        # Check if cache is uninitialized or the set of cached constants doesn't match the current graph's constants
         cache_uninitialized = not device_constant_cache
         cache_invalidated = False
         if device_constant_cache:
-            for torch_constant in self.mcg.constant_inputs[device_idx]:
-                if torch_constant not in device_constant_cache:
-                    cache_invalidated = True
-                    break
+            mcg_constants = set(self.mcg.constant_inputs[device_idx])
+            cache_constants = set(device_constant_cache.keys())
+            cache_invalidated = mcg_constants != cache_constants
 
         # Reset cache if needed by repopulating it from mcg constant inputs
         if cache_uninitialized or cache_invalidated:
-            self._cleanup_runtime_tensor_list(list(device_constant_cache.values()))
+            # Only cleanup tensors that are not None. Complex graph break scenarios can cause a single
+            # graph execution to decompose into multiple __call__s with subgraphs with different constants
+            # or none at all.
+
+            runtime_tensors_to_cleanup = [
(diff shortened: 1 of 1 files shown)


kmabeeTT (approved):
Thanks for quick fix... let's see how it does on that CI run - I don't follow this all as well as I'd like, but it feels good.

tt_torch/dynamo/executor.py
```diff
@@ -288,18 +288,26 @@ def _try_initialize_or_invalidate_constant_cache(self, device, device_idx):
 
         device_constant_cache = self.constant_cache.get(device, {})
 
-        # Check if cache is uninitialized or has invalid entries
+        # Check if cache is uninitialized or the set of cached constants doesn't match the current graph's constants
         cache_uninitialized = not device_constant_cache
         cache_invalidated = False
         if device_constant_cache:
-            for torch_constant in self.mcg.constant_inputs[device_idx]:
```
jameszianxuTT, line 295: This logical check was not fully correct. It only checked that there were no actual constant inputs that were uncached; i.e. that new constant inputs didn't show up. 

The change checks the other direction as well - that there are no precached values that do not show up in the actual graph constants; i.e. that we didn't end up with fewer graph constants than we expect, and if we do, then also invalidate the cache.

tt_torch/dynamo/executor.py
```diff
@@ -288,18 +288,26 @@ def _try_initialize_or_invalidate_constant_cache(self, device, device_idx):
...
-                if torch_constant not in device_constant_cache:
-                    cache_invalidated = True
-                    break
+            mcg_constants = set(self.mcg.constant_inputs[device_idx])
+            cache_constants = set(device_constant_cache.keys())
+            cache_invalidated = mcg_constants != cache_constants
 
         # Reset cache if needed by repopulating it from mcg constant inputs
         if cache_uninitialized or cache_invalidated:
-            self._cleanup_runtime_tensor_list(list(device_constant_cache.values()))
```
jameszianxuTT, line 302: This was unsafe, because device_constant_cache.values() could be `None` but casting to list would turn that into `[None]` which was an invalid input to `_cleanup_runtime_tensor_list`

jameszianxuTT:
Cache invalidation related to graph breaks is tbd to understand and work on. Technically we don't have a current usecase for this, but it's a problematic edgecase.

vmilosevic:

Issue #994, opened 2025-06-25
Wrap tt_mlir.device with a class to capture device-associated entities (e.g. runtime tensor caches)


jameszianxuTT:
A runtime tensor cache should be associated with a tt-mlir device, such that device closure implies deallocation of cache tensors, but there is nothing explicitly linking them at the moment. This would require a change to everywhere we use "Device" in tt-torch, but would make the caching logic more readable.

As part of this change, also change BackendOptions.buffer_cache / constant_cache into their own objects rather than nested dictionaries, alongside clean APIs like InitializeCache(), AddToCache() etc. to reduce logic inside the Executor.

jameszianxuTT:
Closing due to deprecation of tt-torch. Implementing similar logic in tt-xla via BufferInstance persistence.

PR #1011, opened 2025-06-27
Move implicit executor device acquisition to start of executor __call__

jameszianxuTT:
### Ticket
None

### Problem description
Automatic device acquisition via _get_device() happens after cache initialization, so if the device is not specified, the first executor __call__ acquires a device during get_inputs(). This causes the initial constant / buffer cache to be attached to a None device. 

At the first cache insertion time, the device _is_ initialized by automatic device acquisition, so the cache has two incoherent perspectives on the device to attach the cache to - the list of tensors to cache is attached to a None device while the actual allocated runtime tensors are attached to a real device. The caching operation "misses" due to device mismatch.

Then, at destruction time, a runtime error is encountered due to attempted deallocation of uncached (None) runtime tensors.

### What's changed
- Move automatic device acquisition to the start of the Executor __call__ so the device exists before cache initialization time.
- (minor) Add output filtering to llama demo so it only prints output tokens to stdout.
### Checklist

### DIFF
--- a/demos/llama3.2_generate_demo.py
+++ b/demos/llama3.2_generate_demo.py
@@ -10,7 +10,6 @@ from transformers import (
     AutoModelForCausalLM,
     StaticCache,
 )
-import time
 from tests.utils import clear_dynamo_cache
 
 _global_max_cache_len = 64 + 64
@@ -67,6 +66,7 @@ def main():
     model, tokenizer = load_model()
     input_args = load_inputs(model, tokenizer)
     generated_ids = input_args["input_ids"]
+    print(tokenizer.decode(generated_ids[0].tolist()), end="", flush=True)
 
     clear_dynamo_cache()
     cc = CompilerConfig()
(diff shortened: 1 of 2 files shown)


tt_torch/dynamo/executor.py
```diff
@@ -424,7 +424,7 @@ def insert_runtime_tensors_into_caches(torch_tensors, runtime_tensors):
 
         # execute conversion from torch tensors to runtime tensors
         runtime_activations_and_weights = tt_mlir.preprocess_inputs(
-            self._get_device(device_idx=device_idx),
```
jameszianxuTT, line 427: This was previously the point at which the device was first acquired, if left unspecified when invoking our backend. It is a bit weird to be in call -> get_inputs -> tt_mlir.preprocess_inputs -> self._get_device

Device acquisition is a side effect of this call: self._get_device(device_idx=device_idx) which I have moved to happen earlier in the __call__ (only if the device is not initialized and passed into the backend, which is the case for our ModelTester flow)

demos/llama3.2_generate_demo.py
```diff
@@ -10,7 +10,6 @@
     AutoModelForCausalLM,
     StaticCache,
 )
```
jameszianxuTT, line 12: I will squeeze in this llama generate demo output cleanup in this PR as well, minor change and already validated in tt-forge demos.

jameszianxuTT:
This relates to #994 in that it makes device lifetime management a little more sensible.

vmilosevic:

jameszianxuTT:
Also - minor change [llama test output](https://github.com/tenstorrent/tt-torch/actions/runs/15934536110/job/44951775783?pr=1011#step:19:20)

Now produces a single line of tokens:
`<|begin_of_text|>This is a sample text from 2010-2011. The text is from the 2010-2011 edition of the book. The text is from the 2010-201`

AleksKnezevic (approved):

PR #1012, opened 2025-06-27
Fix - replace matrix.build entry with correct runner name

jazpurTT:
### Ticket
Follow-up fix for #993 

### Problem description
Missed one of the matrix.build 'eval_6' test entries and it was skipping the tests all together.

### What's changed
Replaces 'runs-on: n300' with 'wh-runner: n300' and updated the test names for nightlies to avoid conflict

### Checklist
- [x] Passed 'eval_6' without skipping tests


jameszianxuTT:
Thanks @gengelageTT for noticing this so our nightlies didn't get destroyed

Issue #1024, opened 2025-07-02
Data Parallell Detr Test OOM and leaves machine in a bad state


jameszianxuTT:
Repro with `pytest -svv tests/models/detr/test_detr.py::test_detr[data_parallel-full-eval]`

This error is consistently reproed [in July 1 nightly](https://github.com/tenstorrent/tt-torch/actions/runs/16015939196/job/45213989747#step:10:56)

jameszianxuTT:
Autobisection points to https://github.com/tenstorrent/tt-mlir/commit/83be68784772c8b61e8b51d043e742a4de147f51

As a POC of autobisection, here's what those logs look like:
```
./autobisect.py --good 72f18042f3c2c663ab6680cd537275884be57dc3 --bad ef91a93c05e3db2f8c77e0d12f7f63360e8da585 --test-command "pytest -svv tests/models/detr/test_detr.py::test_detr[data_parallel-full-eval]"
\2025-07-02 14:23:12 | INFO     | Bisection tool starting
2025-07-02 14:23:12 | INFO     | tt-torch directory: /localdev/jameszianxu/tt-torch
2025-07-02 14:23:12 | INFO     | Good commit: 72f18042f3c2c663ab6680cd537275884be57dc3
2025-07-02 14:23:12 | INFO     | Bad commit: ef91a93c05e3db2f8c77e0d12f7f63360e8da585
2025-07-02 14:23:12 | INFO     | Test command: pytest -svv tests/models/detr/test_detr.py::test_detr[data_parallel-full-eval]
2025-07-02 14:23:12 | INFO     | Working in directory: /localdev/jameszianxu/tt-torch
2025-07-02 14:23:12 | INFO     | === Starting bisection process ===
2025-07-02 14:23:12 | INFO     | Starting bisect of tt-torch from 72f18042 to ef91a93c
2025-07-02 14:23:12 | INFO     | Found 5 commits in tt-torch between 72f18042 and ef91a93c
2025-07-02 14:23:12 | INFO     | Checking tt-torch commit e1937fbc (3/5)
2025-07-02 14:23:12 | INFO     | Building tt-torch
2025-07-02 14:23:12 | INFO     | Build command: source env/activate && rm -rf third_party/tt-mlir && cmake -G Ninja -B build && cmake --build build && cmake --install build


2025-07-02 14:38:26 | SUCCESS  | Build completed successfully in 0:15:13.972713
[...]

PR #1025, opened 2025-07-02
Disable detr single device + data parallel tests due to nightly dram oom failure

jameszianxuTT:
### Ticket
Follow up in #1024 

Data parallel detr test on n300 is failing consistently in nightly due to DRAM OOM leading to segfault and test group not being reported.

The test in question is tests/models/detr/test_detr.py::test_detr[data_parallel-full-eval] and causing failures in the full eval nightly data parallel 1 group.

Edit. looks like the single device test also hits OOM and segfaults, so both test groups are not reporting.
tests/models/detr/test_detr.py::test_detr[single_device-full-eval]

### DIFF
--- a/.github/workflows/run-full-model-execution-tests-nightly.yml
+++ b/.github/workflows/run-full-model-execution-tests-nightly.yml
@@ -399,7 +399,7 @@ jobs:
 
     - name: Upload Test Report Models
       uses: actions/upload-artifact@v4
-      if: ${{ success() || failure() && env.RUN_TEST != 'skip' }}
+      if: ${{ (success() || failure()) && env.RUN_TEST != 'skip' }}
       with:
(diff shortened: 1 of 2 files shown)


jameszianxuTT:
Local testing shows this may also leave the board in a bad state, leading to these debug prints when running a different model after the detr test failure:
```
Read unexpected run_mailbox value: 0x40 (expected 0x80 or 0x0)
2025-07-02 13:48:49.304 | critical |          Always | Read unexpected run_mailbox value from core (x=25,y=17) (assert.hpp:107)
```

However, this doesn't seem to prevent subsequent tests from passing.

jameszianxuTT:
Validation - limited nightly run on July 1 Nightly failing tests. With DETR disabled, these tests should now pass and report results.

https://github.com/tenstorrent/tt-torch/actions/runs/16030815577

jazpurTT (approved):

PR #1034, opened 2025-07-07
Pass expansion_factor_token=1 to avoid breaking mlpmixer after update 0.2.0 => 0.3.0

jameszianxuTT:
### Ticket
None 

### Problem description
OnPR and nightly CI is failing due to MLPMixer update from July 6th breaking tests on main.

### What's changed
Change expansion_factor_token to 1, to avoid using default 0.5 which causes a typecast to float in the mlpmixer package and breaks their example / default library usage. 

### Checklist

### DIFF
--- a/tests/models/mlpmixer/test_mlpmixer.py
+++ b/tests/models/mlpmixer/test_mlpmixer.py
@@ -20,6 +20,7 @@ class ThisTester(ModelTester):
             dim=512,
             depth=12,
             num_classes=1000,
+            expansion_factor_token=1,  # see mlpmixer package issue #17 https://github.com/lucidrains/mlp-mixer-pytorch/issues/17
         )
         model = model.to(torch.bfloat16)
         return model


AleksKnezevic (approved):

jameszianxuTT:
I am not sure what this parameter is for, but it defaults to 0.5 which breaks the model constructor due to some type coercion to float. 

I tried running the example in their repo readme and it also fails with the same signature. A look at their changenotes also [provides more evidence](https://github.com/lucidrains/mlp-mixer-pytorch/issues/17) that this is just a mistake.

jameszianxuTT:
Looks like the maintainer pushed a fix incredibly fast and updated python package version to 0.3.1 just now, which means we don't need this patch on our side.

Our [CI on a different workflow](https://github.com/tenstorrent/tt-torch/actions/runs/16105516852/job/45489776063#step:10:31) picked up this new package and the test is now passing.

Issue #1010, opened 2025-06-27
Bring up pipeline parallel static cache model


AleksKnezevic:
Please add static cache to the pipeline parallel llama 7b model similar to how it was added to 3b. 

@jameszianxuTT can help with debug.

jameszianxuTT:
Why did this get tagged as a community issue? That doesn't seem right

AleksKnezevic:
@skhatriTT, any idea why this got tagged as community issue?

skhatriTT:
Sorry for late response. We have a github action workflow to check if the issue is created by org member or community and according to logs it didn't identify you as org member(weird!). Will work on this to get it resolved. Thanks for flagging this!

PR #1042, opened 2025-07-10
Bring up pipeline parallel static cache model 

gengelageTT:
### Ticket
#1010 

### Problem description
Add generative loop with static caching to llama_7b model which runs using pipeline parallel. Combines **/demos/llama3.2_generate_demo.py** with **/tests/models/llama/test_llama_7b_pipeline_parallel.py**. This combination introduced mismatches in input sizes between `example_inputs` and `gm.graph_inputs` before lowering as well as incorrect labeling of input types (`USER` vs `INTER_DEVICE`) after lowering.

### What's changed
The test **test_llama_7b_generative_pipeline_parallel.py** was created and added to **run-full-model-execution-tests.yml** with the other pipeline parallel tests.

The remaining changes are in **tt-torch/dynamo/passes.py**:

The `prev_device_idx` is set to the lowest device the previous node was needed on instead of the highest.
- a placeholder (input) was being assigned to the second device when it was need on the first. This created an unnecessary `INTER_DEVICE` input and a mismatch with the `example_inputs` ordering. This was happening because the input was assigned to the `prev_device_idx` which was the second device since the previous input was needed on both devices.

The `user_input_index` (which sets the producer_index) only increments once per node.
-  when it incremented too many times the `producer_index` became incorrect and so certain `gm.graph_inputs` were pointing to the wrong input in `example_inputs` causing shape mismatches

The values in `node_to_new_nodes` aren't rewritten if the `device_idx` is different
- when they were accidentally re-written there were nodes unnecessarily added as `INTER_DEVICE` inputs which caused an issue since `tt_mlir.run()` got more inputs than expected.

Added an assert to check that the input shape matches what's expected to help catch any similar issues in the future. 

### Checklist

### To-Do
When testing pipeline parallel on **llama3.2_generate_demo.py** the device map was assigning the last module to the first device which should cause all modules to move to the first device (bad). However, `sort_device_map()` was not moving the modules over because it doesn't handle `input_device=None` correctly which is something that needs to be fixed. The test causing the issue and a more descriptive explanation is on [this branch](https://github.com/tenstorrent/tt-torch/tree/gengelage/issue-1010-llama3b-bug).
- [ ] Fix sort_device_map to handle input_device = None
- [ ] Consider adjusting sort_device_map to ensure that some modules remain on later devices.

[EDIT]: issues added #1043 #1044


vmilosevic:

tt_torch/dynamo/passes.py
```diff
@@ -421,7 +425,9 @@ def update_device_index(node):
                     )
                     mcg.graph_inputs[device_idx].append(mci)
                     consumer_input_indices[device_idx] += 1
-                    user_input_index += 1
+                    if device_idx == devices[-1]:
+                        # user_input_index should only increment once per node
+                        user_input_index += 1
```
AleksKnezevic, line 430: good catch

tt_torch/dynamo/passes.py
```diff
@@ -403,14 +403,18 @@ def update_device_index(node):
                 assert prev_device_idx is not None
                 devices = [prev_device_idx]
             for device_idx in devices:
-                prev_device_idx = device_idx
+                if device_idx == devices[0]:
+                    # prev_device_idx should be set to the earliest device
+                    prev_device_idx = device_idx
```
AleksKnezevic, line 408: Why is this the case?

gengelageTT: Some of the placeholders/inputs are needed on both devices, which would have made `prev_device_idx=1` instead of 0. However, since we go through the nodes in the order that they appear topologically, if we encounter a node that's needed on both devices we should probably assume that generally speaking we're still working on adding nodes to the first device and so `prev_device_idx` should be 0. 

This caused an error because one of the placeholders had `None` returned from `node_to_device()` for each of its users and so it got assigned to device 1 instead of device 0 where it's needed. The placeholder before it was needed on both devices which is what set `prev_device_idx` to 1 and not 0. Because it was on the wrong device an `INTER_DEVICE` input was made but we don't support `INTER_DEVICE` inputs from device 1 to device 0.

This isn't perfect since there's a very small chance we assign an input to an earlier device when it isn't needed until later, however that wouldn't cause any issues with running the model it would just add an extra `INTER_DEVICE` input between device 0 and device 1. I'm looking into if there's a way to avoid returning `None` from `node_to_device()`  since this also makes sorting the device map difficult. If it's possible it could eliminate this issue entirely.

AleksKnezevic (approved):

jameszianxuTT (approved):

PR #1065, opened 2025-07-16
Added option to enable optimizer and build with tt-mlir op model

jnie-TT:
### Problem description
Currently there's no way for the user to enable optimizer and build tt-mlir with op model.

### What's changed
Added a compiler config flag to enable optimizer, and cmake config flag to build with op model, both defaulting to false.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -34,6 +34,9 @@ else()
     set(TT_RUNTIME_ENABLE_PERF_TRACE OFF CACHE BOOL "Enable runtime Tracy perf tracing")
     message(STATUS "TT_RUNTIME_ENABLE_PERF_TRACE is set to: ${TT_RUNTIME_ENABLE_PERF_TRACE}")
 
+    set(TTMLIR_ENABLE_OPMODEL OFF CACHE BOOL "Enable op model for optimized compile")
+    message(STATUS "TTMLIR_ENABLE_OPMODEL is set to: ${TTMLIR_ENABLE_OPMODEL}")
+
     include(ExternalProject)
     set(TTMLIR_PREFIX ${TTTORCH_SOURCE_DIR}/third_party/tt-mlir)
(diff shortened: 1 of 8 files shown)


vmilosevic:

third_party/CMakeLists.txt
```diff
@@ -58,6 +61,7 @@ else()
           -DTTMLIR_ENABLE_STABLEHLO=ON
           -DTTMLIR_ENABLE_RUNTIME=ON
           -DTTMLIR_ENABLE_BINDINGS_PYTHON=OFF
+          -DTTMLIR_ENABLE_OPMODEL=${TTMLIR_ENABLE_OPMODEL}
```
jameszianxuTT, line 64: @LPanosTT Just noticed this, another directive to transmit to tt-mlir through tt-xla will be needed

tt_torch/csrc/tt-mlir-interface.cpp
```diff
@@ -231,6 +231,7 @@ compileTTIRToTTNN(std::string_view code, std::string_view system_desc_path,
 
   options.enableFusing = true;
   options.enableConstEval = enable_consteval;
+  options.optimizerPassEnabled = enable_optimizer;
```
jameszianxuTT, line 234: Will this cause a failure later on if TTMLIR_ENABLE_OPMODEL is false?

If so - I think we have a binding `is_runtime_debug_enabled` to check if the build supports debug-required calls to runtime - maybe something like that could be implemented here to check if the tt-mlir build supports optimizer before enabling it

jameszianxuTT (approved):
Aside from nit about checking that optimizer is compiled into the build before enabling it, which may be irrelevant, looks good

tt_torch/csrc/tt-mlir-interface.cpp
jnie-TT: Good point! I added a check for op model definition in `compile_ttir_to_bytestream` when optimizer is enabled, also binded API

PR #1080, opened 2025-07-21
Add libgl to 'Getting Started from Source' documentation 

ddilbazTT:
### Ticket
None

### Problem description
`libgl` is necessary for YOLOv5 and it is included in Dockerfile. It is missing from `Getting Started from Source` documentation. Saber was having problems with this. 

### What's changed
Add libgl to `Getting Started from Source`

### Checklist

### DIFF
--- a/docs/src/getting_started_build_from_source.md
+++ b/docs/src/getting_started_build_from_source.md
@@ -125,7 +125,8 @@ sudo apt-get install -y \
     patchelf \
     libyaml-cpp-dev \
     libboost-all-dev \
-    lcov
+    lcov \
+    libgl
 ```
 
 Install OpenMPI:


AleksKnezevic (approved):

kmabeeTT:
Saw this late, but from talking to Erikka, I think in general we do not need to list model specific libs in documentation (there are others missing like ffmpeg, for example), since it's not core compiler deps and that is up to customers/users to know what to install (or just use docker which has everything).

einnesTT:
You don't have to list it if it doesn't break something to do a regular install. We have dependency lists right now only because you have to have them to complete installation. (Also the list may have changed, either more things or less, I didn't build tt-torch this week, I'm still using the one I set up from last week.) If it's model specific, currently the user finds out when the model doesn't run and they're prompted to add whatever library is needed. Ideally, they are told per model or get requirements.txt etc. but having it like that is okay.

einnesTT:
As requested though, I'll check it and add it if build from source wants it.

Issue #1082, opened 2025-07-21
Move model-specific dependencies into model-specific requirements.txt


kmabeeTT:
Currently we have a bunch of model specific pip installed dependencies in top level requirements.txt and also hardcoded inside 6x model test files using manage_dependencies fixture (which is nice, that it installs at start of test and uninstalls at end of test):

```
tests/models/hand_landmark/test_hand_landmark.py:dependencies = ["mediapipe"]
tests/models/openpose/test_openpose.py:dependencies = ["controlnet_aux==0.0.9"]
tests/models/timm/test_timm_image_classification.py:dependencies = ["timm==1.0.9"]
tests/models/timm/test_timm_image_classification_n300.py:dependencies = ["timm==1.0.9"]
tests/models/vovnet/test_vovnet_onnx.py:dependencies = ["timm==1.0.9"]
tests/models/yolov5/test_yolov5.py:dependencies = ["ultralytics==8.2.92", "ultralytics-thop==2.0.6"]
```

Goal:

1. Move all deps into requirements.txt per model (in tt-torch maybe to start, eventually tt-forge-models) and update fixture to install at beginning of test, uninstall at end of test.

I'll put up first/initial PR to at least organize (with comments) top level requirements.txt to better differentiate between core deps and model-specific deps, without migrating them, and convert the 6x models from using dependencies inside the test files to requirements.txt next to the test files (easy change).

PR #1083, opened 2025-07-21
Add comments+organize top level requirements.txt and create 3x model specific requirements.txt

kmabeeTT:
### Ticket
#1082

### Problem description
- We want to organize python requirements to be model-specific

### What's changed
- This is just foundation, part 1 change to add low effort comments to top level requirements.txt to group dependencies into core dependencies, developer dependencies and model-specific dependencies (which will later be moved) so folks looking at it will be less confused.  No changes to dependencies here.
- Somewhat related, move hardcoded model-specific dependencies out of few tests and into their own requirements.txt.
- Remove manage_dependencies fixture and replace with manage_model_requirements fixture (automatically used, no need to opt-in), does the same as before (install deps at start of test, uninstall at end of test) but using requirements.txt instead of hardcoded in test.
- All in all, no functional changes here to keep risk low before pending release. Futher rounds of cleanup needed.

### Checklist
- [x] Run modified tests locally


vmilosevic:

tests/conftest.py
```diff
@@ -24,16 +24,27 @@ def run_around_tests():
...
+
+        yield
+
+        with open(req_file, "r") as f:
+            pkgs = [
+                line.strip().split("==")[0]
+                for line in f
+                if line.strip() and not line.startswith("#")
+            ]
+        subprocess.check_call([sys.executable, "-m", "pip", "uninstall", "-y"] + pkgs)
```
jameszianxuTT, line 44: This is also going to cause some chaos with python's install process when running multiple tests in the same venv (as we do).

pip uninstall won't uninstall dependencies of packages, so each execution of this script is going to dangle some dependencies that aren't explicit in the top level requirements.txt, which could lead to test-ordering based package installation conflicts

Plus, an additional complication with partial porting to this model.

What if we have two tests in the same job that use torchvision where:
1. First test uses a test-scope requirements.txt that specifies something like `torchvision` which is also specified in duplicate in the top level requirements.txt. Then, when that test cleans itself up, it'll uninstall `torchvision` from the venv
2. Some later test doesn't have a test-scope requirements.txt, and uses torchvision it assumes was installed from the top level requirements.txt, but can't find it because test in (1) uninstalled it as part of its cleanup routine

There are some even more messed up cases I imagine here with earlier tests uninstalling dependencies of future tests, if we do a slow port of tests over to this model

jameszianxuTT (approved):
My concerns are mostly related to a future test migration strategy where we slowly port over tests leading to duplicate deps in test-scoped requirements.txt and global requirements.txt

I think the safest way to execute this migration would be to one-shot the entire thing. It's a good change and necessary, I only forsee issues with test migration in the future

PR #1091, opened 2025-07-22
Few various doc updates and debug print for XLA backend

kmabeeTT:
### Ticket
None

### What's changed
 - Organize top level requirements.txt into group with comments for clarity. Refactor later.
 - Add info print to alert user they are using experimental XLA backend.
 - Add important notice about future direction with XLA backend in README.md

### Checklist
- [x] Tested locally

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,22 +1,37 @@
+# Core Dependencies
 --extra-index-url https://download.pytorch.org/whl/cpu
 torch==2.7.0
 torch-xla==2.7.0
-torchvision
-torchaudio
-black
-bi-lstm-crf
-mdutils
 ninja
-pre-commit
 pybind11
+protobuf
+setuptools==59.6.0 #  pinned torchvision wheel build in env/activate requires lower setuptools pinned version. tt-torch #589
+wheel
+onnx==1.17.0 # 1.18.0 fails with "Unsupported model IR version: 11, max supported IR version: 10"
+onnxruntime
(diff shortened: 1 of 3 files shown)


jameszianxuTT (approved):

requirements.txt
```diff
@@ -1,22 +1,37 @@
+# Core Dependencies
```
ddilbazTT, line 1: nit: could we possibly list alphabetically as well? This is only for optical reasons and not necessary in the slightest. I also don't know if it would create conflicting dependencies given the order

tt_torch/dynamo/experimental/xla_backend.py
```diff
@@ -707,6 +707,8 @@ def __del__(self):
 
 @register_backend(name="tt-experimental")
 def xla_backend(gm, example_inputs, options: BackendOptions = None):
+    print("Note: Using experimental XLA backend.")
```
ddilbazTT, line 710: Why don't we wrap this around dump_info or something? Do we want this printed by default?

ddilbazTT (approved):

tt_torch/dynamo/experimental/xla_backend.py
kmabeeTT: Yeah we want this printed by default, and considered even putting it in all caps. At least for now, to be obvious when new backend is used when there is sometimes uncertainty.

PR #1094, opened 2025-07-23
Transformation for random number generation

mmanzoorTT:
### Ticket
Link to Github Issue

### Problem description
TTNN API does not support random number generation.

### What's changed
Add a transformation to convert torch.ops.aten.rand.default op to
randomly generated constant tensor.

### Checklist

### DIFF
--- a/tests/experimental/test_torch_xla_transformations.py
+++ b/tests/experimental/test_torch_xla_transformations.py
@@ -1,0 +1,37 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from tt_torch.tools.verify import verify_against_golden
+import torch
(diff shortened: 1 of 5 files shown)


tests/experimental/test_torch_xla_transformations.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+from tt_torch.tools.verify import verify_against_golden
+import torch
+import torch_xla
+
+import pytest
+
+
+def test_bernoulli_transformation():
```
brataTT, line 11: Do we also want a test for the "tt" backend since that implementation is also included? Or do we only want to add stuff to experimental now that we're moving away from torch-mlir?

LPanosTT:
Will this value end up being constant folded rather than generating a new random tensor each iteration?

PR #1103, opened 2025-07-25
Flip ModelTester default backend=tt-experimental and related test updates

kmabeeTT:
### Ticket
None

### Problem description
 - We want to switch to using tt-experimental (tt-xla) backend by default across most tests sooner rather than later.  Bunch of testing, pass rate is high.  Will continue to push on remining failures in Nightly if any.
 - We also don't have CI capacity to maintain 2 nightlies (normal and experimental) especially on BH side, so if pass rate is high prefer to get this in to ease CI burden.

Note: The tenstorrent branch of torch-xla that we'd like to consume isn't available on PyPi yet, but did some preliminary testing on branch July24 between it (hack to load from s3 bucket) and standard torch-xla=2.7.0 and see exact same results, so when that wheel is ready we can switch to it later.

### What's changed
- Update ModelTester to use backend="tt-experimental" by default, override with backend="tt"
- Add TT_TORCH_FORCE_LEGACY_BACKEND=1 to switch back to legacy tt backend in tests/groups if needed in future
- Some test updates to drop PCC by 0.1 (albert, qwen2), enable consteval (needed in several bigger models) 
- Override arg type map with asif change for failing mobilenet ssd and variants
- ModelTester workaround to automatically use legacy tt backend in data_parallel tests to avoid assert (#1105)
- Use legacy tt backend in 4+5+6=15 n300 batch parallel tests hitting L1 OOM errors (#1105)
- Use legacy tt backend in few tests where they failed like inception_v4.tf_in1k (#1104)
- Use legacy tt backend in phi-2 model to avoid host OOM on CI (#1108)
- Remove experimental CI yml files and references, no longer needed with single nightly. 
- Remove test_e2e_compile yml usage no longer supported with experimental backend, less maintenance.

### Checklist
- [x] A bunch of branch testing on nightly between wormhole (mostly) and blackhole (see comments in PR)


jameszianxuTT (approved):

jameszianxuTT:
Nit: this should be removed:
https://github.com/tenstorrent/tt-torch/blob/46575b72482d148678b4856db4fff113e19dad82/.github/workflows/produce_data.yml#L12

kmabeeTT:
Planning to merge this shortly. 

Here is some more links on CI runs used for testing, which is now clean for wormhole execute tests. What's not tested is full blackhole nightly and op-by-op wormhole nightly, will follow up if anything fails on next real Nightly after this is merged, but not expecting much since blackhole onPR is passing here, and some model op-by-op tests were run with experimental backend too in nightly.

July 24 Nightly-Experimental WH-Only w/ new torch-xla wheel from s3-bucket
`kmabee/experimental_backend_testing`
https://github.com/tenstorrent/tt-torch/actions/runs/16507120910

July 24 Nightly-Experimental WH-Only w/ default torch-xla=2.7.0 version (same results as previous link - great!)
`kmabee/experimental_backend_testing_v270`
https://github.com/tenstorrent/tt-torch/actions/runs/16510213188

Two links in prevous comment I'll mention again here for completeness:

July 24 OnPR full set of tests for this PR/branch - just had eval_6 group failing (covers WH, BH)
`kmabee/experimental_backend_flip`
https://github.com/tenstorrent/tt-torch/actions/runs/16512130507

July 24 Nightly - Subset of previously-failing test groups from push/nightly links above - almost clean (eval_6, eval_9 fails)
`kmabee/experimental_backend_flip.testing`
https://github.com/tenstorrent/tt-torch/actions/runs/16514518311

July 25 Nightly subset of previously failing tests in previous link (eval_6, eval_9) now passing:
`kmabee/experimental_backend_flip.testing`
https://github.com/tenstorrent/tt-torch/actions/runs/16528513828

tests/models/timm/test_timm_image_classification.py
```diff
@@ -127,6 +127,12 @@ def test_timm_image_classification(
         else 0.99
     )
 
+    # FIXME - Issue https://github.com/tenstorrent/tt-torch/issues/1104
```
jameszianxuTT, line 130: I think this is the test that Lewis fixed, [failing in previous invok](https://github.com/tenstorrent/tt-torch/actions/runs/16487606588/job/46616273758#step:10:179) due to a pooling op.

Should be fixed once tt-xla uplift gets merged with [his change](https://github.com/tenstorrent/tt-xla/pull/899) this error should go away. 

Not sure if we care but you could also just xfail if model_name == "inception_v4.tf_in1k"

LPanosTT (approved):

tests/models/timm/test_timm_image_classification.py
kmabeeTT: Yep exactly that's right - it will take some time to get that fix (some tt-xla fail and debug needed it turns out) to prefer to keep it for now to not lose model count if we don't have to.

PR #1130, opened 2025-08-01
Make pytest --forked set only on some groups via new pytest-args variable in yml

kmabeeTT:
### Ticket
None

### Problem description
- The pytest --forked flag was slowing down when running lots of pytests in single process like `tests/experimental/test_torch_xla_basic.py`(locally see 10x slowdown) and it was really only needed for risky/crashy tests anyways

### What's changed
- provide mechanism for it to be easily added to groups via pytest-args and add it to a couple places in push/nightly ymls
- Shrink number of test groups by combining two groups in nightly

### Checklist
- [x] Testing on branch https://github.com/tenstorrent/tt-torch/actions/runs/16666049006

### DIFF
--- a/.github/workflows/run-full-model-execution-tests-nightly.yml
+++ b/.github/workflows/run-full-model-execution-tests-nightly.yml
@@ -44,7 +44,7 @@ jobs:
         build: [
           {
             # Approximately 70 minutes.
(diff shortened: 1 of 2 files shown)


kmabeeTT:
Going to force merge this since it fixes slow/timed out push CI broken for past few days.

PR #1132, opened 2025-08-01
Initial changes for TT-Forge-Models automatic model discovery and runner test

kmabeeTT:
### Ticket
None

### Problem description
- We want to have a single model runner that can detect and import models from tt-forge-models and a way to control FE-specific overrides (PCC) and mark which tests can pass, etc
- We also want this to replace our collection of hand-written models under tests/models/ eventually (soon)

### What's changed
- Here are initial batch of changes for this runner with some nice features
- It will automatically detect and run models/variants, uses pytest-split to split into groups
- We can specify overrides including arch specific overrides in test_config.py, and tag tests by status
- Put new files under tests/runner/ for this, did not touch any existing model tests yet (soon)
- Collects 176 tests/variants, can pass ~129 today locally, CI needs a bit more work, so not scheduling on CI yet
- The utils.py and tt_torch/dynamo/backend.py changes Aleks made, I don't recall what prompted them, will look into that or try reverting them
- There are still a bunch of TODO items, but want to merge these initial changes first.

### Checklist
- [x] High pass rate (-m expected_passing can pass for wormhole/blackhole 129 tests) locally, almost same on CI


.github/workflows/run-tt-forge-models-tests.yml
```diff
@@ -0,0 +1,166 @@
...
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /etc/udev/rules.d:/etc/udev/rules.d
+        - /lib/modules:/lib/modules
+        - /opt/tt_metal_infra/provisioning/provisioning_env:/opt/tt_metal_infra/provisioning/provisioning_env
+        - /mnt/dockercache:/mnt/dockercache
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Check for runner # Check if we need to skip testing when a wh-runner is not set for wormhole run or a bh-runner on blackhole run
```
jameszianxuTT, line 59: As a TODO for later - I think this infrastructure to variably run tests on blackhole/wormhole runners no longer makes sense in this implementation. This include the whole of this Check for runner step, and all the `if: ${{ env.RUN_TEST != 'skip' }}` references in file.

Pytest split generates test groups purely on runtime and we don't yet have infra to solve the optimization problem of generating test groups with more complex constraints (eg. run test A on wormhole, run test B on blackhole etc.), nor a sense of what constraints we might want to express in the future. 

One future solution might be test-scoped arch skips, e.g a test would read its config to see if it's specified to be skipped on a specific arch for whatever reason, and then check the actual arch it is running on via tt_mlir get arch binding. If it detects its current arch should be skipped based on a config setting, then we can skip from within the test. This seems to me like a more elegant solution than marking/creating entire test groups in the YAML with supported arch.

kmabeeTT:
There is a verification step if the pytest is run with `--validate-test-config` that I plan to add to onPush/onPR CI to make sure the `test_config.py` keys (model path + variant name specific) used to specify tagging + PCC etc, is valid (which may be caught if tt-forge-models updates and changes path of models or name of variants for example), but I wanted to catch James before he leaves today, so it's not quite included here yet in CI.  

The output looks like this:

```
ERROR: Unknown test entries in test_config.py (not found in collected tests):
  - vgg/pytorch-vgg19-full-eval
    Did you mean: vgg/pytorch-vgg19_bn-full-eval?
  - yolo_v6/pytorch-full-eval
    Did you mean: yolov6/pytorch-full-eval?
  - yolo_x/pytorch-full-eval
    Did you mean: yolox/pytorch-full-eval?

WARNING: The following tests were collected but are missing from test_config.py:
  - beit/pytorch-base-full-eval
  - beit/pytorch-large-full-eval
  - d_fine/pytorch-large-full-eval
```

There are a buch of other TODO items stil WIP (before I can delete tests and replace them with this in nightly) that I removed from this PR to keep it targeted, will up a follow up PR for them probably.

.github/workflows/run-tt-forge-models-tests.yml
```diff
@@ -0,0 +1,166 @@
+name: Run TT Forge Models Tests
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
jameszianxuTT: TODO It would be _nice_ if the docker build and tt-torch build jobs were conditionally inlined into this test yaml so it would be possible to use via workflow_dispatch , but that's out of scope for this change.

tests/runner/conftest.py
```diff
@@ -0,0 +1,100 @@
...
+            print(f"  - {test_name}")
+            suggestion = difflib.get_close_matches(test_name, _collected_nodeids, n=1)
+            if suggestion:
+                print(f"    Did you mean: {suggestion[0]}?")
+        raise pytest.UsageError(
+            "test_config.py contains entries not found in collected tests."
+        )
+
+    if unlisted:
+        print("\nWARNING: The following tests are missing from test_config.py:")
```
jameszianxuTT, line 98: Github actions can actually surface warnings in test summary annotations, I've never tried it before but here's the mechanism:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message

Also a nit but would be nice to use this mechanism eventually so we see these warnings up front in run summary rather than having to dig through logs to find it.

.github/workflows/run-tt-forge-models-tests.yml
kmabeeTT: Good point - Yep, this was just copy/pasted solution from existing ymls as solution to get wh+bh working where I currently don't have any cases where tests are WH or BH specific (though they will come soon).  Completely agree, I have an idea for how to turn test-scoped arch skip to be used in CI yml file via something like:

-m "expected_passing and $ARCH" 

... to filter out tests. It could assume a test is good for all archs and set marker with $ARCH (I'm making up that vartiable here) in the name, and removed if test is arch specific... something like that anyways..

kmabeeTT:
I don't recall what prompted Aleks two file changes, but I see same pass rate when I drop them in WH + BH locally, so I'm going to drop and force push.

jazpurTT (approved):
Looks good to me, thanks!
This is something we can work on a follow up, but when looking into I thought about how to improve our wh and bh runner infra for testing na dI thought we could maybe add a wh/bh field in test_config, pick that up and pytest.mark the tests, and in CI the jobscan  be filtered by runner and run them. We can discuss in more details if you think if something we can do.

PR #1139, opened 2025-08-05
Change all device references in CI to /dev/tenstorrent

jameszianxuTT:
### Ticket
None

### Problem Description
Loudbox runners are now tagged as `wormhole_b0` meaning tests can arbitrarily be picked up on multidevice runners. Existing device-in-container mounting will lead to an illegal configuration with only a subset of pcie devices mounted in container and all subsequent tests to fail.

### What's changed
Remove references to specific single-device mounting to YAML containers and mount everything to /dev/tenstorrent as suggested in our docs. This does not impact tests that require a set amount of devices because they manually acquire devices during test initialization rather than greedily acquiring all devices available on the system.

### Checklist

### DIFF
--- a/.github/workflows/fail-inspector.yml
+++ b/.github/workflows/fail-inspector.yml
@@ -48,7 +48,7 @@ jobs:
     name: "Test ${{matrix.build.runs-on}}-${{ matrix.build.c }}-${{ matrix.build.commit }}"
     container:
       image: "ghcr.io/tenstorrent/tt-torch/tt-torch-ci-ubuntu-22-04:${{ needs.inspect.outputs.docker-tag }}"
-      options: --device /dev/tenstorrent/0
+      options: --device /dev/tenstorrent
       volumes:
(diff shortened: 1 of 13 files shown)


jameszianxuTT:
Looks like we're still seeing intermittent zipfile corruption on `torch-p150-4` with no known root cause (but this machine is also observed to have significant perf issues most likely caused by CPU throttling from VM setup). Rerunning CI checks that failed blackhole tests.

jameszianxuTT:
This PR is not fully tested until we can actually target some of these to llmbox, it seems that `llmbox-2` is now offlne. What's proven by this onPR checks passing is that passing `/dev/tenstorrent` instead of a specific device `/dev/tenstorrent/0` doesn't introduce new failures when running on known n150/n300/p150 hardware. 

It doesn't prove tests will run on llmbox which is nearly untested to this point

BotDarren:
LGTM

PR #1141, opened 2025-08-05
Create additional case to locate pjrt_plugin_tt.so for bare venv

jameszianxuTT:
### Ticket
None

### Problem description
Experimental backend tests will fail when running from a bare venv (eg. in tt-forge demos/benchmarks) because they are unable to locate the pjrt_plugin_tt.so 

### What's changed
Added logic to find it relative to venv root in <VENV_ROOT>/lib/pjrt_plugin_tt.so

### Checklist
- [x] Local testing. Changes are innocuous as they add a branch that is never hit in our CI since it's handled in a previous case


jameszianxuTT:
imo changes are innocuous as they add a branch that is never hit in our CI since it's handled in a previous case:
```
        source_path = os.path.join(
            os.path.dirname(__file__), "../env/venv/lib/pjrt_plugin_tt.so"
        )
        if os.path.exists(source_path):
            return source_path
```

Issue #1158, opened 2025-08-08
Memory leak somewhere between pytests in tt-torch (20GB in phi2)


kmabeeTT:
**Context:**

Recently when adding new tests from tt-forge-models to tt-torch I hit OOM on CI in previously passing and untoched phi2 models (when adding other models) which sure feels like a memory leak of some kind. I put some comments to this effect [here](https://github.com/tenstorrent/tt-torch/pull/1137#issuecomment-3162426551) in the PR and as quick workaround I merged the new tests along with `gc.collect()` in tt-forge-models runner `test_models.py` at end which worked around problem.

No surprise, but this memory leak isn't constrained to just models run via auto model discovery `test_models.py` but our hand-written tests in CI too, example below.

The problem is visible if we port simple `memory_usage_tracker()` fixture from tt-forge-fe which samples process memory usage before test, during test run (separate thread), after test, and after garbage collection, we can see that there is huge amount of memory not freed in phi2 model.  Below output is with experiment to comment out the `gc.collect()` it adds at end, for illustrative purposes:

**Example:**

```

% /usr/bin/time -v pytest -svv tests/models/phi/test_phi_1_1p5_2.py::test_phi[full-microsoft/phi-2-eval] tests/models/mamba/test_mamba.py::test_mamba[full-mamba-790m-hf-eval] --durations=0 --log-memory-usage

Memory usage before test: 895.25 MB

Test memory usage:
    By test: 31411.68 MB
    Minimum: 895.25 MB
    Maximum: 32306.93 MB
    Average: 15543.27 MB

Memory usage before garbage collection: 22207.31 MB
Memory usage after garbage collection (skipped): 22207.31 MB <=== Wow, not good.

<first test done, second test starting now>

Memory usage before test: 22207.31 MB <=== Inherits from previous test

Test memory usage:
    By test: 4491.36 MB
    Minimum: 12395.38 MB
    Maximum: 26698.67 MB
    Average: 13046.32 MB

Memory usage before garbage collection: 12729.41 MB
Memory usage after garbage collection (skipped): 12729.41 MB


================== 2 passed, 3 warnings in 405.55s (0:06:45) ===================
        Command being timed: "pytest -svv tests/models/phi/test_phi_1_1p5_2.py::test_phi[full-microsoft/phi-2-eval] tests/models/mamba/test_mamba.py::test_mamba[full-mamba-790m-hf-eval] --durations=0 --log-memory-usage"
        User time (seconds): 873.83
[...]

PR #1159, opened 2025-08-08
Add memory_usage_tracker fixture + cleanup for OOM and add back more tests

kmabeeTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/1158

### Problem description
- Some tests hits OOM on CI unexpectedly, in part due to memory leak between tests, but we didn't have good way to know

### What's changed
- Port memory_usage_tracker() from tt-forge-fe which samples before, during, after test and does memory cleanup
- Needed to add pybind on malloc_trim (similar to FFE) for it to be usable
- Optionally log memory usage to CSV file. Deploy in push+nightly and upload artifact
- Memory cleanup allows us to add more tests that would previously get killed due to leak on prior tests

### Checklist
- [x] Tested logic locally, more tests on CI


kmabeeTT:
Linked ticket shows the benefit of this change, copy paste here (when running phi-2 model, drops by 20GB):

```
Memory usage before test: 896.23 MB
Memory usage before garbage collection: 22201.40 MB
Memory usage after garbage collection: 4672.91 MB <=== Huge Drop
Memory usage after malloc_trim: 1321.39 MB <=== Even more

Memory usage before test: 1321.39 MB <=== Much better than 21GB
Memory usage before garbage collection: 7770.61 MB
Memory usage after garbage collection: 4722.73 MB
Memory usage after malloc_trim: 1563.35 MB
```

tests/conftest.py
```diff
@@ -190,3 +196,113 @@ def pytest_configure(config):
...
+    Note:
+        - This fixture is automatically used for all tests due to the `autouse=True` parameter.
+        - The interval for memory readings can be adjusted by changing the sleep duration in the `track_memory` function.
+        - Min, max, and avg memory usage are calculated based on the recorded memory readings from system memory.
+    """
+    process = psutil.Process()
+
+    # Initialize memory tracking variables
+    start_mem = process.memory_info().rss / (1024 * 1024)  # MB
+    print(f"Memory usage before test: {start_mem:.2f} MB")
```
kmabeeTT, line 229: Unsure whether we want to keep these prints. They were useful to illustrate/debug for me, but they are noisy. In tt-forge-fe they are logger.info, but we don't use loguru, so I made them prints. Example output for phi2:

```
                 Always |  WARNING | User provided a tensor of data type: Int64 which is not supported by runtime/ttnn. Casting to: Int32, this may impact throughput and the integrity of the data.                                                                                 [60/4816]
                 Always |  WARNING | User provided a tensor of data type: Int64 which is not supported by runtime/ttnn. Casting to: Int32, this may impact throughput and the integrity of the data.
Memory usage before test: 1321.39 MB
Neither required_atol, or relative_atol is provided. Setting required_atol=0.01.
[MODEL NAME] pytorch_mamba_mamba-790m-hf_nlp_causal_lm_huggingface
Warning: Model is generative but `run_generate=False`. This will run `model(**inputs)` instead of `model.generate(**inputs)`
Note: Using experimental XLA backend.
Results for output 0:
  PCC: 0.9553, threshold: 0.95 __
  ATOL: 48.5000, threshold: 0.01 __ (assert_atol == False)

Generated text:  [', are you?"?"\n']
PASSEDTest memory usage:
    By test: 6662.57 MB
    Minimum: 1321.39 MB
    Maximum: 7983.96 MB
    Average: 7430.71 MB
Memory usage before garbage collection: 7770.61 MB
Memory usage after garbage collection: 4722.73 MB
Memory usage after malloc_trim: 1563.35 MB
```

jameszianxuTT (approved):

tests/conftest.py
jameszianxuTT: Yeah I agree these prints should be suppressed or gated behind a custom verbosity flag if we really wanted to see them, it would help minimize the amount of chaff we throw to stdout. The usecase of this feature is to debug memory issues in CI, but we can inspect the memory artifact for this.

tests/conftest.py
```diff
@@ -190,3 +196,113 @@
...
+    min_mem = min(min_mem, end_mem)
+    max_mem = max(max_mem, end_mem)
+    total_mem += end_mem
+    count += 1
+    avg_mem = total_mem / count
+
+    by_test = max_mem - start_mem
+
+    # Log memory usage statistics
+    print(f"Test memory usage:")
```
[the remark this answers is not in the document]
kmabeeTT: maybe, copied from tt-forge-fe, updated this line to include test_name, now warning gone :)

kmabeeTT:
Now that this is in nightly and dumping statistics, here is screenshot of output CSV artifact ([memory-usage-wormhole-grp_5-47752470285-expected_passing](https://github.com/tenstorrent/tt-torch/actions/runs/16857198749/artifacts/3728027559) downloaded from a [nightly](https://github.com/tenstorrent/tt-torch/actions/runs/16857198749) run:

image

PR #1181, opened 2025-08-20
Add support for TT_TORCH_IR_LOG_LEVEL for experimental backend.

mmanzoorTT:
### Ticket
None

### Problem description
'TT_TORCH_IR_LOG_LEVEL' is not supported for experimental backend.

### What's changed
Setting TT_TORCH_IR_LOG_LEVEL env variable will dump following graphs:
1. Torch graph module
2. Torch compiled graph module
3. VHLO module
4. SHLO module
5. TTIR module
6. TTNN module

### Checklist

### DIFF
--- a/tt_torch/dynamo/experimental/xla_backend.py
+++ b/tt_torch/dynamo/experimental/xla_backend.py
@@ -584,6 +584,12 @@ def bypass_assert_tensor_metadata(gm):
     return gm
 
 
+def dump_module(module, name, compiler_config):
+    if compiler_config.dump_info:
+        print(f"{name} module", flush=True)
+        print(module, flush=True)
+
+
 def xla_pass_pipeline(gm, example_inputs, compiler_config):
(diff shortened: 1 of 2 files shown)


jameszianxuTT:
In the PR comment:
> Setting TT_TORCH_IR_LOG_LEVEL env variable will dump following graphs

I think this should be clarified to only reenable the INFO level dump for old backend, not the DEBUG level one that would give IR before and after every compiler pass. That DEBUG level is useful to see the actual result of TTIR->TTIR transformations like the kv cache fusion, as well as the various sharding passes for multichip.

Also, since the lowering past FX level is now out of our hands, I think the "DEBUG" level IR dump needs to be implemented in tt-xla. I have such a change in [tt-xla for my own testing](https://github.com/tenstorrent/tt-xla/commit/d861782d4ede3aee01f475122923e6b85c3bea89).

PR #1180, opened 2025-08-20
Uplift third_party/tt-mlir to fdb981c6f047aa71d3044e95dbc81b3f6761117b 2025-08-20

vmilosevic:
This PR uplifts the third_party/tt-mlir to the fdb981c6f047aa71d3044e95dbc81b3f6761117b

PR #1185, opened 2025-08-21
Revert "Uplift third_party/tt-mlir to a3ea3703fb0db66c76a92173d4ba1af94ca27f67 2025-08-21"

jameszianxuTT:
Reverts tenstorrent/tt-torch#1180 which was accidentally automerged due to a bug in check-all-green which will permit merging under invalid circumstances. 

A Github outage on Aug 21 4UTC caused a specific job on `ubuntu-latest` to fail, skip all dependent tests, mark the changeset as mergable from the perspective of check-all-green.

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -133,6 +133,7 @@ jobs:
       - pre-commit
       - spdx
       - docker-build
+      - check-files
       - build
       - test
       - full-model-test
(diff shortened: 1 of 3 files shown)


jameszianxuTT:
Actually, on-push for the [uplift job](https://github.com/tenstorrent/tt-torch/actions/runs/17132920335) passed, suggesting my fix to cut the build runners back to non-shared did fix the profiling test failures, so we can actually merge. This is probably the first time I've seen something fail in onPR but pass in onPush.

Issue #1182, opened 2025-08-20
Disable parameter folding 


AleksKnezevic:
Please run nightly full model tests with consteval_parameters defaulted to false. If results match main, please merge in.

@jameszianxuTT FYI.

PR #1201, opened 2025-08-25
[wip] Remove references to consteval parameters

jameszianxuTT:
### Ticket
#1182 

### Problem description
When applying mark_sharding() on user tensors through the torch.compile flow, we are unable to correlate original user tensors with their potentially constant folded equivalents. 

Also, we aim to remove constant folding entirely from the frontend so it can be handled in MLIR.

### What's changed
Remove all references to consteval_parameters

### Checklist

### DIFF
--- a/demos/llama3.2_generate_demo.py
+++ b/demos/llama3.2_generate_demo.py
@@ -45,7 +45,6 @@ def main():
     clear_dynamo_cache()
     cc = CompilerConfig()
     cc.enable_consteval = False
-    cc.consteval_parameters = False
 
     options = BackendOptions()
     options.compiler_config = cc
--- a/demos/resnet/resnet50_data_parallel_async.py
+++ b/demos/resnet/resnet50_data_parallel_async.py
@@ -33,7 +33,6 @@ def download_image(url):
 def main():
     cc = CompilerConfig()
     cc.enable_consteval = True
-    cc.consteval_parameters = True
 
     num_devices = DeviceManager.get_num_available_devices()
     parent, devices = DeviceManager.acquire_available_devices()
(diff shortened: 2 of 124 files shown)


jameszianxuTT:
This is currently not mergable. As shown by @ssaliceTT's [nightly sweep](https://github.com/tenstorrent/tt-torch/actions/runs/17213658076), many of our nightly tests will be broken by removal of this option. 

Failures are all PCC drop from what I've seen, from small ~0.01 PCC drop up to 0.3 PCC drop in [[this example](https://github.com/tenstorrent/tt-torch/actions/runs/17213658076/job/48832686228#step:14:2397)].

The action of this change is to eliminate constant folding on parameters, which will prevent their hoisting from the graph at torch.compile time, and introduce potentially big TMs into runtime that weren't there before, like the ever-present [nn.Linear transpose](https://discuss.pytorch.org/t/why-does-the-linear-module-seems-to-do-unnecessary-transposing/6277) that brought initial attention to this problem.

jameszianxuTT:
One alternative solution for now, is to use the `skip_folding_node_fn` argument to `const_fold.split_const_subgraphs`, which lets us specify a set of nodes to CF during backend compilation.  ([Pytorch ref](https://github.com/pytorch/pytorch/blob/cf94cadbeee31a4d1d46a57f11bce7c9fd1cebc0/torch/fx/experimental/const_fold.py#L157))

I believe we can map a torch fx node -> tensor weight -> shard map introduced in #1189. Then return `false` if the user tensor weight has a shard_spec attached to it.

An even simpler solution would be just to turn off consteval_parameters for SPMD tensor parallel tests and demos.

PR #1245, opened 2025-11-14
Aknezevic/ultralytics

AleksKnezevic:
Removed 'ultralytics' from the development requirements.

Ticket
Link to Github Issue

Problem description
As per our lawyers we cannot use ultralytics until we have a licence from them. It will take too long to get licence, so in the mean time we need to remove it from our repos.

What's changed
Removed ultralytics


kmabeeTT (approved):
Yep, looks good.

